Issue #52, opened 2022-08-02
Spit `ui.paraprocess` logs into normal stdout via `tea.Printf`


nderjung:
This issue tracks the ability to simultaenously output logs via [`bubbletea`](https://github.com/charmbracelet/bubbletea)'s `Printf` method such that if an error occurs it is retrievable through the user's terminal.  For now, the user interface hides this by truncating the build log and this is not helpful if a larger error occurs.  To mitigate against this, simultaneously outputing this way (which would be hidden via `bubbletea`'s UI window mechanism) so we can have the progress bar print still at the top.

This issue tracks this implementation fix.

craciunoiuc:
This is no longer relevant, as simpler methods of printing exist, like `basic`/`json`/others which do not hide output

PR #83, opened 2022-08-04
Redirect `stderr` output from the `make` subprocess

craciunoiuc:
This PR adds the ability to `kraft build` to also print compilation warnings/errors/etc. to the console.

Edit: This is now printed only at the end 

Closes: #52


craciunoiuc:
I think this was fixed now, right?

craciunoiuc:
This was fixed in the meantime 🪄

PR #85, opened 2022-08-05
cmd/kraft: Add ability to check for an update

StefanJum:
Add an automatic check for available updates on every kraftkit run.
Add option to check for new prereleases.

Solves: #1 

Signed-off-by: Stefan Jumarea <stefanjumarea02@gmail.com>

If `kraft` is run without the `--check-prerelease` option, it automatically checks for a new release.
If the `--check-prerelease` option is set, it looks for a new prerelease. This is disabled by default.

### DIFF
--- a/cmd/kraft/kraft.go
+++ b/cmd/kraft/kraft.go
@@ -32,8 +32,17 @@
 package main
 
 import (
+	"fmt"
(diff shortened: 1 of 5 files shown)


nderjung:
Closing in favour of #276

Issue #33, opened 2022-07-22
Make Release pipeline dependent on Go-Checks pipeline


nderjung:

craciunoiuc:
No need for this anymore, we are not merging PRs that do not pass quality checks anyways

PR #86, opened 2022-08-10
.github/workflows: Add dependency on coding style for releases

craciunoiuc:
Please actually test this before merging.

The support for this is a bit sketchy, as there is no formal way to retrieve the status from previous jobs.
Also, if a job depends on multiple previous steps, it can't retrieve their status.

There are workarounds, but they aren't needed at the moment.

https://stackoverflow.com/questions/63148639/create-dependencies-between-jobs-in-github-actions

Closes: #33

### DIFF
--- a/.github/workflows/release-staging.yaml
+++ b/.github/workflows/release-staging.yaml
@@ -10,8 +10,12 @@ permissions:
   packages: write
 
 jobs:
+  gochecks:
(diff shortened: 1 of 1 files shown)


craciunoiuc:
No need for this anymore, we are not merging PRs that do not pass quality checks anyways

PR #113, opened 2022-10-24
Add unsource subcommand to `pkg`

craciunoiuc:
This worked out of the box because there was already existing functionality for it.

Currently gives no error if the source didn't exist in the first place (not sure if it should).

Closes: #18


craciunoiuc:
Migrating to new PR #182  as remote is stale

PR #117, opened 2022-10-28
Add nix flake

jschlumpp:
This allows installing kraftkit using an flakes-enabled nix installation.


craciunoiuc:
Closing in favour of using the existing NUR packages

We can revisit in the distant future

PR #124, opened 2022-10-31
Add `kraft.yaml` templating

### LINKED ISSUE #27: Kraftfile templating
This issue tracks the ability to have a "template" within the kraftfile, such that its contents is retrieved and applied as a base before the user's kraftfile is applied (in a additive fashion).

Templates offer a convenient way to using pre-configured Unikraft constructs as the basis for your application.

When initializing a new project to be built as Unikraft unikernel, 

### Example 1

The following template:

```yaml
spec: 0.6
name: my-template
version: 0.0.1
unikraft: 0.9
libraries:
  - newlib: v123
targets:
  - arch: x86_64
    plat: kvm
```

when used:

```yaml
spec: 0.6
name: my-app
version: 0.0.1
template: my-template:0.0.1
```

Results in:

```diff
  spec: 0.6
  name: my-app
  version: 0.0.1
- template: my-template@0.0.1
+ unikraft: 0.9
+ libraries:
+   - newlib: v123
+ targets:
+   - arch: x86_64
+     plat: kvm
```

### Example 2

```yaml
spec: 0.6
name: my-template
version: 0.0.1
unikraft: 0.9
libraries:
  - newlib: v123
targets:
  - arch: x86_64
    plat: kvm
```

Using the template:

```yaml
spec: 0.6
name: my-app
version: 0.0.1
template: my-template@0.0.1
unikraft:
  version: 0.9
  kconig:
    - MYLIB=y
```

Results in:

```diff
  spec: 0.6
  name: my-app
  version: 0.0.1
- template: my-template@0.0.1
  unikraft:
    version: 0.9
    kconig:
      - MYLIB=y
+ libraries:
+   - newlib: v123
+ targets:
+   - arch: x86_64
+     plat: kvm
```

## Implementation

1. Add a new optional directive to the `kraft-spec.json` so as to interpret a string or component by the name `template`.

2. Create a new a new unikraft package, `unikraft/template` with the following:

```go
type TemplateConfig struct {
	configuration.ComponentConfig

	Type component.ComponentType `yaml:",omitempty" json:"type,omitempty"`
}
```    

3. Add `Template` to `ApplicationConfig`:

```diff
type ApplicationConfig struct {
	component.ComponentConfig

	WorkingDir    string                `yaml:"-" json:"-"`
	Filename      string                `yaml:"-" json:"-"`
	OutDir        string                `yaml:",omitempty" json:"outdir,omitempty"`
+	Template      template.templateConfig   `yaml:",omitempty" json:"template,omitempty"`
	Unikraft      core.UnikraftConfig   `yaml:",omitempty" json:"unikraft,omitempty"`
	Libraries     lib.Libraries         `yaml:",omitempty" json:"libraries,omitempty"`
```

4. Introduce similar functionality to [`loadUnikraft`](https://github.com/unikraft/kraftkit/blob/staging/schema/loader.go#L251-L254) but for a single template.  Use [`WithType`](https://github.com/unikraft/kraftkit/blob/staging/unikraft/component/option.go#L72) and set it to `unikraft.ComponentTypeApp`.

5. So far, the schema is a static declaration of the project which has not used for dynamic population, we must wait until the template can be fully rendered, i.e. just-in-time.  Whilst it does have access to the package manager which does allow it to retrieve information, it has never been used so far to pull down components.  This is largely because we also want to pass to a `context.Context` object just-in-time as well as additional things [to the pulling mechanism (e.g. a progress bar function)](https://github.com/unikraft/kraftkit/blob/staging/cmd/kraft/build/build.go#L355-L358). The next step should be to simply add this new component to the [list of known components to the application](https://github.com/unikraft/kraftkit/blob/staging/unikraft/app/application.go#L439-L457) but this is faced with conflicting information "just-in-time" since the template will also need to provide to us new additional information, and therefore potentially new components. This means that during sequences [such as this](https://github.com/unikraft/kraftkit/blob/staging/cmd/kraft/build/build.go#L285) would not fully render by the intention of fulfilling the goals of a template.  To mitigate against this, we must throw an if a). a template is being used and b). the template component returns `false` with [`IsUnpackedInProject`](https://github.com/unikraft/kraftkit/blob/staging/unikraft/component/component.go#L193).  The use of a template should be conditional to it already existing and so we must then add pre-emptive code to retrieve it by modifying both `kraft build` and `kraft pkg pull` in order to retrieve it.

6. In `kraft build`, add new logic to find and pull [before the invocation of [`project.Components()`](https://github.com/unikraft/kraftkit/blob/staging/cmd/kraft/build/build.go#L285).  This is can be achieved by directly invoking the package manager if a template is set and pulling pre-emptively.  Similar logic should appear in [`kraft pkg pull`](https://github.com/unikraft/kraftkit/blob/staging/cmd/kraft/pkg/pull/pull.go#L249).  Ultimately, this code (along with other similar sequences should be abstracted into `internal/cmd/<similar sequence methods>`.

7. Back inside of `schema`, the method created earlier and inside the `Load` method, since we now have access to the template data, we can re-populate the whole returning project. Override values from the template by calling `NewApplicationFromOptions` with the same options on the template.  Start initially by loading the template, then override anything that is provided by the user.

8. Since we have now added potentially new configuration options which were not originally provided to us on an initial call of `NewApplicationFromOptions`, we must in fact change the way we are operating with regard to accessing these values.  Unexport all `ApplicationConfig` attributes and migrate these to methods, e.g. `Libraries` becomes `func (ac *AppplicationConfig) Libraries ([]*lib.LibraryConfig, error)`. The last error returns true if the template of the application is a). a set and non-empty and b). the template component returns `false` with `IsUnpackedInProject`.  Update all references to make this change. 

9. The above step guarantees access to the correct values at any moment of the use of the `ApplicationConfig` object, however, to initially populate these values, we must introduce `options.go` to the `unikraft/app` package.  New options should be used during the creation of the `ApplicationConfig`, e.g. `WithLibraries(libs ...*lib.LibraryConfig) (ApplicationConfigOption, error)` which is able to add to the now private `libraries` attribute.

craciunoiuc:
Introduce the possibility to use templates in Unikraft applications. More details in the linked issue. 

The PR follows the Issue steps, but some are reordered, as to be logical from a commit-steps point-of-view.

Will have to fix the checks.

Closes: #27

### DIFF
--- a/cmd/kraft/build/build.go
+++ b/cmd/kraft/build/build.go
@@ -391,6 +391,14 @@ func buildRun(opts *buildOptions, workdir string) error {
 	if err != nil {
 		return err
 	}
+
+	templateProject, err := schema.NewApplicationFromOptions(templateOps)
+	if err != nil {
+		return err
+	}
+
+	project = templateProject.MergeTemplate(project)
+
 	// Overwrite template with user options
 	components, err := project.Components()
 	if err != nil {
--- a/cmd/kraft/build/build.go
+++ b/cmd/kraft/build/build.go
@@ -289,7 +289,17 @@ func buildRun(opts *buildOptions, workdir string) error {
 	var processes []*paraprogress.Process
 	var searches []*processtree.ProcessTreeItem
 
-	for _, component := range project.Components() {
+	_, err = project.Components()
+	if err != nil {
+	}
+	}
+
+	// Overwrite template with user options
+	components, err := project.Components()
+	if err != nil {
+		return err
+	}
+	for _, component := range components {
 		component := component // loop closure
 
 		searches = append(searches, processtree.NewProcessTreeItem(
--- a/cmd/kraft/build/build.go
+++ b/cmd/kraft/build/build.go
@@ -42,6 +42,7 @@ import (
 	"kraftkit.sh/exec"
 	"kraftkit.sh/pack"
 	"kraftkit.sh/schema"
+	"kraftkit.sh/unikraft"
 
 	"kraftkit.sh/internal/cmdfactory"
 	"kraftkit.sh/internal/cmdutil"
@@ -54,7 +55,6 @@ import (
 	"kraftkit.sh/tui/paraprogress"
 	"kraftkit.sh/unikraft/app"
 	"kraftkit.sh/unikraft/component"
-	"kraftkit.sh/unikraft/target"
 
 	// Subcommands
 	"kraftkit.sh/cmd/kraft/build/clean"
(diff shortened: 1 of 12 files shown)


nderjung (approved):
Thanks so much @craciunoiuc!

PR #164, opened 2022-12-07
Expose CLI builder package as public package

nderjung:
This PR migrates `internal/cli` into the top-level and public package of `cmdfactory` such that it can be used externally by other Go modules and programs.


craciunoiuc (approved):
All good here. Thanks!

nderjung:
Thanks for having a look, I've now rebased it

nderjung:
Looks like there's a static code analysis error -- but this looks like a separate issue.

PR #165, opened 2022-12-10
Fix setting correct KConfig options in short arch switch

nderjung:
This PR fixes the translation switch used in taking the human-configurable and abbreviation for a architecture which is then used to correctly set the Unikraft KConfig option.


craciunoiuc (approved):
All good here. Thanks!

PR #169, opened 2022-12-14
Small corrections to README.md to reduce distractions while reading.

TheApeMachine:
- Adds `stage` as the most likely missing word in the sentence.
- Adds (opinionated) comma.
- Adds (opinionated) split between two similar words to optimize the flow of the text.

> Not sure if there is a more specific format to use, but noticed these things while getting started looking into this project.


nderjung:
Hi @TheApeMachine, thanks for the fix! :)

Please just edit your commit message to include a sign-off as [we require a DCO](https://unikraft.org/docs/contributing/submitting-changes/#developers-certificate-of-origin) before merging changes.  We need to add more information about contributing to KraftKit.

TheApeMachine:
Does it work like this (I updated the commit message) or does the pull request message require it too?

nderjung (approved):
Thanks! Just the commits, since this is the "source of truth".

PR #173, opened 2022-12-21
Simplify application creation API and fix target naming

nderjung:
This PR greatly simplifies the creation of a new `ApplicationConfig` based on the "project option" mode.  In this instantiation method, numerous "Kraftfiles" are read and parsed along with options as to which checks and mechanisms are performed or along with additional data to enrich the `ApplicationConfig`. The instantiation method now looks like:

```go
project, err := app.NewProjectFromOptions(
	app.WithProjectWorkdir("path/to/my/app"),
	app.WithProjectDefaultKraftfiles(),
)
```

Much of the complexity was introduced by employing the loading mechanisms from f2bb67c which imported code from the Docker Compose codebase.  The design choice here split up both the loading and the instantiating of a "project".  In this PR we essentially merge these facets since they were greatly inter-dependent and would have not been used independently. The result is a simpler API which is seen in all `cmd/*` changes.

Implicitly, this PR fixes an issue where the target name was set inconsistently. To mitigate this, we only set the kernel target name in the method `NewApplicationFromOptions`.  This is called implicitly by the `NewProjectFromOptions`.  Previously, an attempt to set the kernel name was made by `LoadTarget`.  This function was unable to set the name appropriately as we do not yet know the name by the time its called during the parsing a Kraftfile. Instead, standardise the `LoadTarget` method to be in-line with other top-level attributes of a Kraftfile, e.g. `LoadUnikraft` such that they have the same parameterisation.

After parsing all the Kraftfiles, `NewProjectFromOptions` performs a "clean" instantiation based on all the information it has retrieved from parsing all Kraftfiles.  As a result, we can now appropriately set the name in `NewApplicationFromOptions` method.

Finally, we use the kernel name in the `ApplicationConfig` object's `KConfigFile` method as opposed to the target name.  This provides us a more distinct mechanism for identifying the which dot config resulted in the final binary.

A number of additional fixes and simplifications are included.


craciunoiuc:
Things look good over here. Once you address the things discussed on Discord, it should be good to merge.

craciunoiuc (approved):
All good here. Thanks!

PR #192, opened 2023-01-20
Fix added bugs and inconsistencies

craciunoiuc:
First two commits remove a wrongly added binary and then ignore tool binaries from now on.

The last reverts a function from bubbletea that currently breaks when running in kraftkit.

### DIFF
--- a/tui/paraprogress/paraprogress.go
+++ b/tui/paraprogress/paraprogress.go
@@ -168,7 +168,11 @@ func (md *ParaProgress) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	if complete == len(md.processes) {
 		md.quitting = true
-		return md, tea.Sequence(tea.Batch(cmds...), tea.Quit)
+		batch := tea.Batch(cmds...)
+		if batch == nil {
+			return md, tea.Quit
+		}
+		return md, tea.Sequence(batch, tea.Quit)
 	}
 
 	return md, tea.Batch(cmds...)
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,10 @@ dist/
 # Ignore vscode folder
 .vscode/
 
+# Ignore tool binaries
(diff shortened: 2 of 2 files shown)


tui/paraprogress/paraprogress.go
```diff
@@ -168,7 +168,10 @@ func (md *ParaProgress) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	if complete == len(md.processes) {
 		md.quitting = true
-		return md, tea.Sequence(tea.Batch(cmds...), tea.Quit)
+		// This line will disable linting for the whole file, be careful
+		//nolint:staticcheck
+		//lint:ignore SA1019 using Sequence is currently breaking inside bubbletea
+		return md, tea.Sequentially(tea.Batch(cmds...), tea.Quit)
```
nderjung: This issue arises from upgrading bubbletea which exposes a new API. Since using this API throws a nil pointer exception and instead of ignoring the underlying issue and using the older now deprecated API by ignoring all issues in the file [a separate issue in itself](https://github.com/golangci/golangci-lint/issues/741), we should address the nil pointer directly. Taking a look at how the new API from bubbletea works shows that the nil will be returned by `tea.Batch` if the list of `cmds` is empty. So we can check for this return type and simply `tea.Quit` if this is the case:

```go

		batch := tea.Batch(cmds...)
		if batch == nil {
			return md, tea.Quit
		}

		return md, tea.Sequence(batch, tea.Quit)
```

Please adapt the above explanation text in the git commit message which fixes this issue.

nderjung (changes requested):

tui/paraprogress/paraprogress.go
nderjung: [Ideally the `nil` `cmd` should actually be checked by bubbletea](https://github.com/charmbracelet/bubbletea/issues/639).

nderjung (approved):
Thanks!

PR #194, opened 2023-01-21
Package management and component refactor

nderjung:
## Prerequisite checklist

## Description of changes

This PR provides a number of major improvements to KraftKit internals. The PR can roughly be split into two key areas of improvement:

 1. Migrate the transformation of components (i.e. libraries, architectures, platforms, etc.) into their respective packages and start to specialize each component without embedding a common structure but instead using a common interface; and,
 2. Provide overall better support for "package management" of components (that is, everything from packaging and distribution and vice-versa).

Other than stability, the improvements aid in the general ease-of-use in terms of Go API exposed by the `kraftkit.sh` module.  Using KraftKit APIs programmatically inside of KraftKit should be straightforward but also will help facilitate external extensions that base their work on the module. For example, building external plugins which feed into existing systems which aim in building unikernels systematically.

The two above areas are described below in greater detail.

### Migrating `transform`ation of components

This PR migrates the attributes of `ComponentConfig` into the individual components (concretely, `ComponentConfig` used to be embedded in `LibraryConfig` and other components and now it is no longer).  This abstraction was ultimately considered unnecessary and it complicated the use of each component.  This changeset is quite large to adjust for this migration and includes the following additional changes:

 - migrate the `Transform`ation function to each component package, allowing the the component to be more compartmentalized in terms of related work (that is, only the component should know how to "unmarshal" itself from an interface);
 - introduce a general-purpose "Unikraft Context" which mimics variables that are part of the Unikraft core build system. For now these are only useful for a complete build.
 - pass context to each transformer such that we can access the Unikraft context which is wrapped in a general context structure.  Later on we can add additional things like logging or looking up information in the package manager thanks to this new access.
 - unexport (privatize) the attributes of each component and expose their value through public (and immutable) methods.

This reduction and simplification ultimately enables us to keep the functionality of updating the attributes of the individual components centralized to the component (that is, the value of each attribute is highly controlled and should not necessarily be updated without proper function).

### Better package management abstraction and interfaces

Broadly, this PR creates clarity between subsystems which implement the`packmanager.PackageManager` and `pack.Package` interface.  Now there is clear separation between the two, where the `PackageManager` now is used to:

 1. Add and remove sources which return lists of packages;
 4. Allow for querying sources to find specific packages; and,
 5. Packaging (`Pack`ing) a component into an implementation-specific package and unpackaging (`Unpack`ing) a provided package into a component.

The last destinction (3.) was somewhat unclear in the previous implementation of KraftKit which this commit resolves.  Specifically, the packaging of a `unikraft.component.Component` type was handled in `cmd/kraft/pkg/pkg.go` when it should simply call the package manager with the provided project (which implements the `Component` interface). This allows the implementing interface to handle different component types and allows the APIs of KraftKit be used to package components which was not previously possible.

Since the `PackageManager` interface resolves `Package`s, the operation of `Pull`ing and `Push`ing have now been introduced as interface methods to `pack.Package`.  This is because the implementing package will have contextual information about where its origin is from and so it is possible to retrieve (`Pull`) from this origin as well as `Push` to it.

This change has ultimately simplified the use of using the package manager as well as building packages from components. This is idealized by the large simplification found in the file `cmd/kraft/pkg/pkg.go`.


craciunoiuc (approved):
All good here. Thanks!

PR #205, opened 2023-01-25
Pass KraftKit config as global CLI flags

nderjung:
### Prerequisite checklist

### Description of changes
This PR dynamically injects all configuration options for KraftKit into the cmdfactory's `Main` method which is used to setup the execution of a binary built with this system.  This allows us to dynamically at the command-line set global configuration options.  For example, to change the log type and level, one can simply invoke:

```shell
kraft --log-type basic --log-level debug pkg update
```

This will show a more verbose output without having to manually edit the configuration file (i.e. on Linux it's `~/.config/kraftkit/config.yaml`).

To make this possible, the cmdfactory had to be slightly re-arranged to expose a new method `AttributeFlags` which takes a given `obj` which is a struct and associate all attributes to the given `cobra.Command`. Previously, this was used exclusively when instantiating new commands, but now occurs also if choosing to run through the globally used `Main` method.  This association required changes to the default CLI options too, particularly removing the fact that each option returned a function.  This is no longer the case, we have all information associated at "option instantiation time".

Finally, to combat the fact that an object may already have values set when passed to `AttributeFlags`, the field's value is set right after the flag is created for said field.

### DIFF
--- a/cmd/kraft/kraft.go
+++ b/cmd/kraft/kraft.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"github.com/MakeNowJust/heredoc"
+	"github.com/rancher/wrangler/pkg/signals"
 	"github.com/spf13/cobra"
 
 	"kraftkit.sh/cmdfactory"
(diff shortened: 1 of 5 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #210, opened 2023-01-30
Miscelaneous fixes for supporting git cloning

craciunoiuc:
### Prerequisite checklist

### Description of changes

<s>This commit removes the dependency to the currently mostly-breaking `libgit` library. The `go-git` library does not use the `git` binary and, as such is independent of it.
Unfortunately, progress is now emulated, as there is no standard for progress, as it needs to be provided by the remote entity.
GitHub provides only the total number of blocks pulled, after this was done, and several other steps.
As such, the progress is shown relatively accurately up to ~30%, and then is just incremented up to 99%. We can't go to 100% as that would stop the download. 100% will be set by the process when it finishes.

We can get the current downloaded size by continuously checking the `.git` folder, but to get the total download size we are dependent on the external platform. It is different for GitHub/Gitlab/other gits.

Fortunately the use case is pretty limited and won't affect many people that they can't see the progress of downloads, as most of they are relatively small (1-2MBs).

In order to test this functionality, set the output to `fancy` and run using this command:

```
rm -rf .unikraft/unikraft && /home/cez/kraftkit/dist/kraft pkg pull -Z -g -w . https://github.com/unikraft/unikraft.git
```
</s>

This can now be fixed by linking libgit statically and compiling it together with kraftkit.

Close: #212  
Close: #196 
Close: #197 
Close: #207 
Close: #208 
Close: #209


cmd/kraft/pkg/pull/pull.go
```diff
@@ -226,22 +227,31 @@ func (opts *Pull) Run(cmd *cobra.Command, args []string) error {
...
+					query.Types = append(query.Types, t)
+				}
+
+				if len(n) > 0 {
+					query.Name = n
+				}
+
+				if len(v) > 0 {
+					query.Version = v
+				}
```
nderjung: No need to parse the URI, since this is an assumption made by the the `Catalog` method which will "understand" what to do with the input source. In the manifest implementation, it detects the source as being a valid HTTP URI, or a directory, etc.

Fine just to do this, also looks a bit neater:

```diff
                // Is this a list (space delimetered) of packages to pull?
        } else {
                for _, c := range strings.Split(query, " ") {
-                       query := packmanager.CatalogQuery{}
-                       t, n, v, err := unikraft.GuessTypeNameVersion(c)
-                       if err != nil {
-                               continue
-                       }
-
-                       if t != unikraft.ComponentTypeUnknown {
-                               query.Types = append(query.Types, t)
-                       }
-
-                       if len(n) > 0 {
-                               query.Name = n
+                       query := packmanager.CatalogQuery{
+                               NoCache: opts.NoCache,
                        }
-
-                       if len(v) > 0 {
-                               query.Version = v
+                       if t, n, v, err := unikraft.GuessTypeNameVersion(c); err == nil {
+                               if t != unikraft.ComponentTypeUnknown {
+                                       query.Types = append(query.Types, t)
+                               }
+
+                               if len(n) > 0 {
+                                       query.Name = n
+                               }
+
+                               if len(v) > 0 {
+                                       query.Version = v
+                               }
+                       } else {
+                               query.Source = c
                        }

                        queries = append(queries, query)
```

manifest/pack_pull_git.go
```diff
@@ -7,14 +7,60 @@ package manifest
...
-	"github.com/libgit2/git2go/v31"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
 
 	"kraftkit.sh/log"
 	"kraftkit.sh/pack"
 	"kraftkit.sh/unikraft"
 )
 
+type GitWriter struct {
```
nderjung: I'd keep this private and rename to `gitProgressCalculator`

manifest/pack_pull_git.go
```diff
@@ -7,14 +7,60 @@ package manifest
...
+	io.Writer
+
+	currentStep uint64
+	totalSteps  uint64
+
+	matcher *regexp.Regexp
+	popts   *pack.PullOptions
```
nderjung: Rather pass the progress function directly, i.e. `onProgress func(float64)`

manifest/pack_pull_git.go
```diff
@@ -7,14 +7,60 @@ package manifest
...
+		total, err := strconv.ParseUint(match[2], 10, 64)
+		if err != nil {
+			return len(data), err
+		}
+
+		gp.currentStep = current
+		gp.totalSteps = total
+	}
+
+	gp.popts.OnProgress(float64(gp.currentStep / gp.totalSteps))
```
nderjung: which would result in `gp.onProgress(float64(gp.currentStep / gp.totalSteps))`

manifest/pack_pull_git.go
```diff
@@ -33,23 +79,24 @@ func pullGit(ctx context.Context, manifest *Manifest, opts ...pack.PullOption) e
...
-					return 0
-				},
-			},
-		},
-		CheckoutOpts: &git.CheckoutOptions{
-			Strategy: git.CheckoutSafe,
-		},
-		Bare: false,
+	gitWriter := &GitWriter{
+		popts:       popts,
```
nderjung: `onProgress: popts.onProgress`

nderjung (changes requested):
Hey a few suggestions, though i believe you're still working on this PR

nderjung (approved):
Awesome, thanks :)

PR #227, opened 2023-02-10
Statically compile libgit2 into KraftKit

### LINKED ISSUE #31: Statically link binaries
This issue tracks statically linking all binaries available in KraftKit.

The major overhead here is the use of github.com/libgit2/git2go which depends on `libgit2-dev`.

The reason for using this library is that it offers progress callback (so cloning a repo can be monitored).  If native Go library which implements Git can offer the same, this should be used as an alternative.  Otherwise, this issue also explores statically linking against the `libgit2-dev` shared library https://github.com/libgit2/libgit2

### LINKED ISSUE #179: Remove libgit2 dependency
This issue tracks removing the [`git2go` dependency](github.com/libgit2/git2go). KraftKit currently makes use of `git2go` which use of an external `libgit2` shared object dependency which is used to calculate the progress of a clone (so when you run `kraft pkg pull` and it resolves to a git repo, the progress bar of the clone is shown).  Whilst useful, this dependency issue causes installation issues and is preventing some new users from using KraftKit.

The use of `git2go` is used only in one location in KraftKit:

https://github.com/unikraft/kraftkit/blob/staging/manifest/pack_pull_git.go#L37

The way to approach a rewrite is to use the almost ubiquitious use of [`go-git/v5`](github.com/go-git/go-git).  [Here's an example of a progress bar based on `v4`](https://gist.github.com/tyru/82cae8bad2b116f442d08eeef456e23e).

nderjung:
### Prerequisite checklist

### Description of changes

Statically compile libgit2 into kraftkit.  This has been a major source of frustration for newcomers to the project who must use a specific libgit2 shared object dependency in auxiliary to the supplied `kraft` binary.

To mitigate against this, statically compile libgit2 with KraftKit.

GitHub-Fixes: #179
GitHub-Closes: #31


.github/workflows/gochecks.yaml
```diff
@@ -22,8 +22,12 @@ jobs:
...
 
       # Temporary workaround until this pipeline uses the 'myself' buildenv
-      - name: Install libgit2-dev
-        run: apt-get update && apt-get install -y --no-install-recommends libgit2-dev
+      - name: Install libgit2-dev dependencies
+        run: |
+          apt-get update;
+          apt-get install -y --no-install-recommends \
+            cmake \
+            libgit2-dev;
```
craciunoiuc:
```suggestion
            cmake;
```

No longer needed.

.github/workflows/release-stable.yaml
```diff
@@ -32,9 +32,12 @@ jobs:
...
       
-      # Temporary workaround until this pipeline uses the 'myself' buildenv
-      - name: Install libgit2-dev
-        run: apt-get update && apt-get install -y --no-install-recommends libgit2-dev
+      - name: Install libgit2-dev dependencies
+        run: |
+          apt-get update;
+          apt-get install -y --no-install-recommends \
+            cmake \
+            libgit2-dev;
```
craciunoiuc:
```suggestion
            cmake;
```

No longer needed.

.github/workflows/release-staging.yaml
```diff
@@ -32,9 +32,12 @@ jobs:
...
       
-      # Temporary workaround until this pipeline uses the 'myself' buildenv
-      - name: Install libgit2-dev
-        run: apt-get update && apt-get install -y --no-install-recommends libgit2-dev
+      - name: Install libgit2-dev dependencies
+        run: |
+          apt-get update;
+          apt-get install -y --no-install-recommends \
+            cmake \
+            libgit2-dev;
```
craciunoiuc:
```suggestion
            cmake;
```

No longer needed.

Makefile
```diff
@@ -162,3 +141,15 @@ properclean: IMAGE       ?= $(REGISTRY)/$(ORG)/$(REPO)/$(ENVIRONMENT):$(IMAGE_TA
...
+git2go: $(VENDORDIR)/github.com/libgit2/git2go/v31/static-build/install/lib/pkgconfig/libgit2.pc
+
+$(VENDORDIR)/github.com/libgit2/git2go/v31/static-build/install/lib/pkgconfig/libgit2.pc: $(VENDORDIR)/github.com/libgit2/git2go/v31/vendor/libgit2
+	$(MAKE) -C $(VENDORDIR)/github.com/libgit2/git2go/v31 install-static
+
+$(VENDORDIR)/github.com/libgit2/git2go/v31/vendor/libgit2: $(VENDORDIR)/github.com/libgit2/git2go
+	$(GIT) -C $(VENDORDIR)/github.com/libgit2/git2go/v31 submodule update --init --recursive
+
+$(VENDORDIR)/github.com/libgit2/git2go:
+	$(GIT) clone --branch v31.7.9 --recurse-submodules https://github.com/libgit2/git2go.git $@/v31
```
craciunoiuc, line 156:
```suggestion
	$(GIT) clone --branch $(GIT2GOBRANCH) --recurse-submodules https://github.com/libgit2/git2go.git $@/v31
```

We should probably make it a variable so it is easier to change/we don't forget about it.

buildenvs/myself.Dockerfile
```diff
@@ -42,6 +42,7 @@ RUN set -xe; \
     apt-get install -y --no-install-recommends \
       build-essential \
       libgit2-dev \
+      cmake \
```
craciunoiuc:
```suggestion
      cmake \
```

No longer needed.

.gitignore
```diff
@@ -1,9 +1,6 @@
 # Ignore distributables folder
 dist/
 
-# Ignore vscode folder
-.vscode/
-
```
craciunoiuc: No need to remove the ignore to the vscode server. We can ignore it and add the changes at the same time.
For example I have some logs in the folder that are autogenerated and they pollute the `git status` command.

image

If we don't ignore it, there might also be a change to add these files accidentally.
I say we ignore it, but keep the `settings.json` change.

craciunoiuc (changes requested):
An initial set of comments from my side

Makefile
nderjung: Then we need a bit of Makefile magic to strip the `.7.9` out of the `v31.7.9` variable value, probably it is fine to leave as is for now

craciunoiuc (approved):
All good here. Thanks!

PR #242, opened 2023-02-17
unikraft/app: Pin ordering for libC's

craciunoiuc:
Because libraries are kept in a map, their ordering can't be guaranteed This comes in conflict with libC-type libraries, which need to sit first. This is not a problem related with them, and comes from other libraries using the `#include_next` syntax on them.

To fix this we pin either `musl` or `newlib` plus `pthread-embedded` to come first, if they exist.

### Prerequisite checklist

### DIFF
--- a/unikraft/app/application.go
+++ b/unikraft/app/application.go
@@ -276,7 +276,29 @@ func (app application) IsConfigured(tc target.Target) bool {
 func (app application) MakeArgs(tc target.Target) (*core.MakeArgs, error) {
 	var libraries []string
 
-	for _, library := range app.libraries {
+	// TODO: This is a temporary solution to fix an ordering issue with regard to
+	// syscall availability from a libc (which should be included first).  Long-term
+	// solution is to determine the library order by generating a DAG via KConfig
+	// parsing.
+	unformattedLibraries := lib.Libraries{}
+	for k, v := range app.libraries {
+		unformattedLibraries[k] = v
+	}
+
+	// All supported libCs right now
+	if unformattedLibraries["musl"] != nil {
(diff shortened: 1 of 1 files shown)


unikraft/app/application.go
```diff
@@ -276,7 +276,33 @@ func (app application) IsConfigured(tc target.Target) bool {
 func (app application) MakeArgs(tc target.Target) (*core.MakeArgs, error) {
 	var libraries []string
 
-	for _, library := range app.libraries {
+	//  _______ ____  _____   ____
+	// |__   __/ __ \|  __ \ / __ \
+	//    | | | |  | | |  | | |  | |
+	//    | | | |  | | |  | | |  | |
+	//    | | | |__| | |__| | |__| |
+	//    |_|  \____/|_____/ \____/
```
nderjung: Lol, you can remove this ;)

unikraft/app/application.go
```diff
@@ -276,7 +276,27 @@ func (app application) IsConfigured(tc target.Target) bool {
 func (app application) MakeArgs(tc target.Target) (*core.MakeArgs, error) {
 	var libraries []string
 
-	for _, library := range app.libraries {
+	// TODO dirty hack to get build to work regardless of order of libraries
+	// Remove this once fixed in the main repository
```
nderjung: Can you update the text:
```
// TODO: This is a temporary solution to fix an ordering issue with regard to
// syscall availability from a libc (which should be included first).  Long-term
// solution is to determine the library order by generating a DAG via KConfig
// parsing.
```

nderjung (approved):
Looks good to me, thanks!

Issue #167, opened 2022-12-10
Parallelize GitHub probing


nderjung:
This issue tracks parallelization of the probing function in the [`GitHubProvider`](https://github.com/unikraft/kraftkit/blob/staging/manifest/github.go) so as to enable faster updates from this source type.

PR #249, opened 2023-02-22
Adds parallelized GitHub probing

ls-1801:
### Prerequisite checklist

  - [X] Ran `make fmt` on your commit series before opening this PR;

### Description of changes
Parallelizes the probing of GitHub repositories. The initial fetch of all relevant repositories is still done sequentially, due to fetching each page individually. The overall speed-up is still significant (18s down to 4s).

#### Verify
The changes can be verified by deleting the kraftkit folder containing manifest (`~/.local/share/kraftkit`) and running the `kraft pkg update` command.

#### Open Questions:
- Should there be a limit on how many concurrent connections can be opened simultaneously?
- How do we deal with errors? Abort on the first error or collect all errors? Currently, the second option is implemented
- Where can I find relevant documentation?
- Is the order of manifests important? Currently, the order is based on when the GitHub requests finish.

### DIFF
--- a/manifest/github.go
+++ b/manifest/github.go
@@ -7,10 +7,12 @@ package manifest
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 
 	"github.com/gobwas/glob"
 	"github.com/google/go-github/v32/github"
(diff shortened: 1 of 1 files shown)


manifest/github.go
```diff
@@ -193,27 +195,61 @@ func (ghp GitHubProvider) manifestsFromWildcard() ([]*Manifest, error) {
...
+		var msg string
+		for _, err := range errs {
+			if msg == "" {
+				msg = err.Error()
+			} else {
+				msg = fmt.Sprintf("%s\n%s", msg, err.Error())
+			}
+		}
+		return nil, fmt.Errorf("%w", errors.New(msg))
 	}
```
nderjung: Since the project uses Go 1.20, we can use the [`errors.Join`](https://pkg.go.dev/errors#Join) method to wrap errors, i.e.:

```go
// combine errors into a single error
if len(errs) > 0 {
	return nil, errors.Join(errs...)
}

**Review by nderjung — CHANGES_REQUESTED** [review]
Hi, thanks so much for this! I tried it out locally and works like a charm. There's one suggestion in-line which helps reduce code complexity.

**Comment by nderjung** [comment]
To answer your questions:

> Should there be a limit on how many concurrent connections can be opened simultaneously?

This can be manipulated by the [`GOMAXPROCS` environmental variable](https://pkg.go.dev/runtime#GOMAXPROCS), so no need to provide additional instrumentation.

> How do we deal with errors? Abort on the first error or collect all errors? Currently, the second option is implemented

Collecting them as provided makes sense (aka "errgroup") and allows for partial caching of manifest packages whilst allowing for those that failed to be provided through the `errors.Join`.

> Where can I find relevant documentation?

We're working on that 😅 

> Is the order of manifests important? Currently, the order is based on when the GitHub requests finish.

The order is not important.

**Comment by nderjung** [comment]
Please can you also update your Git commit message to include:
```
GitHub-Fixes: #167
Signed-off-by: ...
```

ls-1801:
I changed the error aggregation to `errors.join`.

nderjung (approved):
Awesome! Thanks for this! 

Issue #213, opened 2023-02-07
Host setup subcommand


nderjung:
### Feature request summary

Using KraftKit requires host-specific instrumentation before usage which can be daunting for newcomers. To help ease the process of getting started with KraftKit, introduce a `kraft setup` subcommand which guides the user through a number of host-specific checks and installations which enable users to build and run unikernels. This step is distinct from the installation process of KraftKit itself which simply downloads the toolchain itself.

The `kraft setup` subcommand should recognise the user's host OS and distribution such that it can determine if running virtualized guests is indeed possible and to provide specific package installations through recognised package managers.

The guided setup should be verbose in its explanation, e.g.:

- If virtualization is not possible it should explain how to access such an environment and that building unikernels is still possible;
- It is possible in some cases to determine if the user is inside a Virtual Machine already such as VirtualBox and prompt an explanation as to how to enable nested-virtualization;
- It should prompt for root-privilege access when packages are about to be installed.

For unsupported OS/distribution combinations, a link to this issue should be provided such that users can follow the progress of additional distributions as this issue acts as an epic for this process.

The following OS/distributions should be implemented:

- [ ] General internal setup system (`kraftkit.sh/internal/setup`)
- [x] #214
- [x] #215
- [x] #216
- [x] #217
- [x] #218
- [x] #219
- [x] #220
- [x] #224

The guided setup should also prompt the user for which virtualization solution they wish to use and make necessary checks for this, including:

- [x] #221
- [x] #222
- [x] #223
- [ ] Create a OS group and chown the runtime directory

### Related architectures

None

### Related platforms

None

### Additional context

For example, the installation process for building and running unikernels via QEMU/KVM on Debian usually requires the following additional APT packages:

```shell
apt-get install -y --no-install-recommends \
  build-essential \
  libncurses-dev \
  libyaml-dev \
  flex \
  git \
  wget \
  socat \
  bison \
  unzip \
  uuid-runtime;
```

These packages should be listed under the `Linux/Debian` and `Linux/Ubuntu` tuple and the above command executed on behalf of the user.

The organization of the code introduced by `kraft setup` should be delivered via an internal package, i.e. `kraftkit.sh/internal/setup`, since this is KraftKit-specific methods for enabling its use.

craciunoiuc:
Closed in favour of the installer script which already does this

PR #257, opened 2023-03-04
cmd: Add kraft setup command

StefanJum:
The `kraft setup` command will install the required dependencies for building and running unikernels. It will also check the virtualization support of the environment, check if we are running kraft from inside a conainter or a virtual machine and give a warning about enabling nested virtualization

### Prerequisite checklist

### DIFF
--- a/cmd/kraft/kraft.go
+++ b/cmd/kraft/kraft.go
@@ -31,6 +31,7 @@ import (
 	"kraftkit.sh/cmd/kraft/rm"
 	"kraftkit.sh/cmd/kraft/run"
 	"kraftkit.sh/cmd/kraft/set"
+	"kraftkit.sh/cmd/kraft/setup"
 	"kraftkit.sh/cmd/kraft/stop"
 	"kraftkit.sh/cmd/kraft/unset"
 	"kraftkit.sh/cmd/kraft/version"
@@ -77,6 +78,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(ps.New())
 	cmd.AddCommand(rm.New())
 	cmd.AddCommand(run.New())
+	cmd.AddCommand(setup.New())
 	cmd.AddCommand(stop.New())
 
 	cmd.AddGroup(&cobra.Group{ID: "misc", Title: "MISCELLANEOUS COMMANDS"})
--- a/cmd/kraft/setup/setup.go
+++ b/cmd/kraft/setup/setup.go
@@ -1,0 +1,75 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2022, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package setup
+
+import (
+	"github.com/MakeNowJust/heredoc"
+	"github.com/spf13/cobra"
+
+	"kraftkit.sh/cmdfactory"
+	"kraftkit.sh/internal/setup"
+)
+
+type Setup struct {
+	WithOS   string `long:"with-os" usage:"Force the host OS"`
+	WithArch string `long:"with-arch" usage:"Force the architecture of the host"`
+	WithVMM  string `long:"with-vmm" usage:"Force the use of a specific VMM on the host"`
+	WithPM   string `long:"with-pm" usage:"Force the use of a specific package manager on the host"`
+}
+
+func New() *cobra.Command {
+	return cmdfactory.New(&Setup{}, cobra.Command{
+		Short:   "Setup the working environment for building and running unikernels",
+		Use:     "setup",
+		Aliases: []string{"sup"},
+		Args:    cmdfactory.MaxDirArgs(0),
+		Long: heredoc.Doc(`
+		Setup the working environment for building and running unikernels`),
+		Example: heredoc.Doc(`
+			# Setup the environment
+			$ kraft setup`),
+		Annotations: map[string]string{
+			"help:group": "build",
+		},
+	})
+}
+
+func (opts *Setup) Run(cmd *cobra.Command, args []string) error {
+	var err error
+	var sopts []setup.SetupOption
+
+	ctx := cmd.Context()
+
+	if opts.WithOS != "" {
+		sopts = append(sopts, setup.WithOS(opts.WithOS))
+	} else {
+		sopts = append(sopts, setup.WithDetectHostOS())
+	}
+
+	if opts.WithArch != "" {
+		sopts = append(sopts, setup.WithArch(opts.WithArch))
+	} else {
+		sopts = append(sopts, setup.WithDetectArch())
+	}
+
+	if opts.WithVMM != "" {
+		sopts = append(sopts, setup.WithVMM(opts.WithVMM))
+	} else {
+		sopts = append(sopts, setup.WithDetectVMM())
+	}
+
+	if opts.WithPM != "" {
+		sopts = append(sopts, setup.WithPM(opts.WithPM))
+	} else {
+		sopts = append(sopts, setup.WithDetectPM())
+	}
+
+	if err := setup.DoSetup(ctx, sopts...); err != nil {
+		return err
+	}
+
+	return err
+}
--- a/internal/setup/options.go
+++ b/internal/setup/options.go
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2022, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package setup
+
+import (
+	"context"
+
+	"github.com/shirou/gopsutil/v3/host"
+)
+
+type SetupOption func(*Setup) error
+
+type HostPackageManager interface {
+	Install(context.Context, ...string) error
+}
+
+// Check if the provided `os` variable is valid in format and supported.
+func WithOS(os string) SetupOption {
+	switch os {
+	case "ubuntu", "linux/ubuntu":
+		return NotSupported
+	case "windows":
+		return NotSupported
+	}
+
+	return NotSupported
+}
+
+// Check if the provided `arch` variable is valid in format and supported.
+func WithArch(arch string) SetupOption {
+	switch arch {
+	case "x86_64":
+		return NotSupported
+	case "arm64":
+		return NotSupported
+	}
+
+	return NotSupported
+}
+
+// Check if the provided `vmm` variable is valid in format and supported.
+func WithVMM(vmm string) SetupOption {
+	switch vmm {
+	case "kvm":
+		return NotSupported
+	case "xen":
+		return NotSupported
+	}
+
+	return NotSupported
+}
+
+// Check if the provided `pm` variable is valid in format and supported.
+func WithPM(pm string) SetupOption {
+	return NotSupported
+}
+
+func WithDetectHostOS() SetupOption {
+	os_name := detect_os()
+	return WithOS(os_name)
+}
+
+func WithDetectArch() SetupOption {
+	arch := detect_arch()
+	return WithArch(arch)
+}
+
+func WithDetectVMM() SetupOption {
+	vmm := detect_vmm()
+	return WithVMM(vmm)
+}
+
+func WithDetectPM() SetupOption {
+	return NotSupported
+}
+
+func detect_os() string {
+	os_info, _ := host.Info()
+	os_name := os_info.OS + "/" + os_info.Platform
+
+	return os_name
+}
+
+func detect_arch() string {
+	os_info, _ := host.Info()
+	arch := os_info.KernelArch
+
+	return arch
+}
+
+func detect_vmm() string {
+	os_virt, role, _ := host.Virtualization()
+	vmm := os_virt + "/" + role
+
+	return vmm
+}
+
+func NotSupported(*Setup) error {
+	return nil
+}
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2022, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package setup
+
+import (
+	"context"
+)
+
+type Setup struct {
+	WithOS   string `long:"with-os" usage:"Force the host OS"`
+	WithArch string `long:"with-arch" usage:"Force the architecture of the host"`
+	WithVMM  string `long:"with-vmm" usage:"Force the use of a specific VMM on the host"`
+	WithPM   string `long:"with-pm" usage:"Force the use of a specific package manager on the host"`
+}
+
+func DoSetup(ctx context.Context, sopts ...SetupOption) error {
+	s := Setup{}
+	for _, opts := range sopts {
+		if err := opts(&s); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}


craciunoiuc:
`conainter` -> `container` in the commit message 😄

cmd/kraft/kraft.go
```diff
@@ -77,6 +78,8 @@ func New() *cobra.Command {
 	cmd.AddCommand(ps.New())
 	cmd.AddCommand(rm.New())
 	cmd.AddCommand(run.New())
+	cmd.AddCommand(set.New())
```
craciunoiuc: Isn't there already a `set` above?

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,148 @@
...
+	"kraftkit.sh/cmdfactory"
+)
+
+type Setup struct{}
+
+func New() *cobra.Command {
+	return cmdfactory.New(&Setup{}, cobra.Command{
+		Short:   "Setup the working environment for building and running unikernels",
+		Use:     "setup",
+		Aliases: []string{"s"},
```
craciunoiuc: Double check that aliases don't clash with other commands.

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,148 @@
...
+		"bison",
+		"unzip",
+		"uuid-runtime",
+		"python3",
+		"python3-setuptools",
+		"python3-pip",
+		"qemu-kvm",
+		"qemu-system-x86",
+		"sgabios",
+	}
```
craciunoiuc: I think maybe we should restrict those by versions? Not sure.

At least a lower-bound we should set

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,148 @@
...
+
+	installCmd := exec.Command(cmd, args...)
+	installCmd.Stdout = os.Stdout
+	installCmd.Stderr = os.Stderr
+	err = installCmd.Run()
+
+	return err
+}
+
+func check_virt(os_name string, virt string, role string) error {
```
craciunoiuc: The default syntax for golang is camelcase, is this intentional? 😄

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,148 @@
...
+	// TODO add virtualization support checks for kvm/xen/VBox etc.
+	if strings.Compare(role, "guest") == 0 && strings.Compare(virt, "kvm") == 0 {
+		fmt.Println("=======================================================")
+		fmt.Println("Looks like you are running inside a KVM virtual machine.")
+		fmt.Println("To run unikernels, you need to enable nested virtualization.")
+		fmt.Println("You can check out the link below if you need help doing that:")
+		fmt.Println("https://www.howtogeek.com/devops/how-to-enable-nested-kvm-virtualization/")
+		fmt.Println("=======================================================")
```
craciunoiuc: I think you can use backticks to print using a single `Println`

StefanJum:
> `conainter` -> `container` in the commit message smile

Done, sorry :laughing:

craciunoiuc (changes requested):
Hey @StefanJum 

Added some quick notes from my side.

One thing I don't get here is what `setup` should be doing.

I was under the impression that it should set up the environment, like the config options and other os-specific options.

The installation I thought was handled by the install script:
https://get.kraftkit.sh

Also tagging @nderjung

StefanJum:
> Hey @StefanJum
> 
> Added some quick notes from my side.
> 
> One thing I don't get here is what `setup` should be doing.
> 
> I was under the impression that it should set up the environment, like the config options and other os-specific options.
> 
> The installation I thought was handled by the install script: https://get.kraftkit.sh
> 
> Also tagging @nderjung

Thanks, I'll check the comments.

From what I understood `kraft setup` should prepare the environment for building and running unikernels.
That would mean installing the dependencies and detecting the virtualization support (xen, kvm etc.) in order to use that when configuring/running the applications.
I think the best way to do that is by saving the found virtualization solution in the kraftkit `config` file, but the installation of all requirements would still be necessary as a first step.

Does the install script add all the requirements for building and running unikernels or just for building/running `kraft`?

@nderjung please correct me if I'm wrong.

cmd/kraft/setup/setup.go
StefanJum: nope, solved :)

craciunoiuc:
Actually you are right.

What I wanted to say is that maybe we should be kind-of merging these two together somehow.

From my perspective when you first run a tool you should have all dependencies already solved. I would just add an extra check in the install script to ask "Do you want to install building and running dependencies"

I'm fine either way. Let's see what Alex has to say

nderjung (changes requested):
Hi @StefanJum. Thanks for putting this together.

To be inline with how logic is approached in the `cmd/` directory, each subcommand program essentially calls functionality exposed by the "KraftKit framework". I think the `setup` cmd should be similar and therefore the logic for actually performing the checks, running any installation, etc. should occur as an exposed API from a different package. The goal here is to keep the file `cmd/setup/setup.go` as simple as possible.

[I suggest re-working the logic into a new package, `kraftkit.sh/internal/setup`. as is suggested by the corresponding issue for this feature](https://github.com/unikraft/kraftkit/issues/213).

Towards a re-usable API, i'd recommend starting simply and with something like the following:

`cmd/setup/setup.go`:
```go
import "kraftkit.sh/internal/setup"

type Setup struct {
  OS   string `long:"os" usage:"Force the host OS"`
  Arch string `long:"arch" usage:"Force the architecture of the host"`
  VMM  string `long:"vmm" usage:"Force the use of a specific VMM on the host"`
  PM   string `long:"pm" usage:"Force the use of a specific package manager on the host"`
}

// [...]

var sopts []setup.SetupOption

if opts.OS != "" {
  sopts = append(sopts, setup.WithOS(opts.WithOS))
}

// [...etc...]

if err := setup.Setup(ctx, ...sopts); err != nil {
  return err
}
```

`internal/setup/setup.go`:
```go
type Setup struct {
  arch *HostArch
  os   *HostOS
  pm   *HostPackageManager
  vmm  *HostVMM
}

func Setup(ctx context.Context, sopts ...SetupOption) error {
  s := Setup{}
  for _, opts := range sopts {
    if err := opt(&s); err != nil {
      return err
    }
  }

  var err error

  if s.arch == nil {
    s.arch, err = DetectHostArch()
    if err != nil {
      return err
    }
  }
    
  // [...]
}
```

`internal/setup/options.go`:
```go
// [...]

type SetupOption func(*Setup) error

func WithOS(os string) SetupOption {
 // This should check if the provided `os` variable is valid in format and
 // in support. E.g. providing `ubuntu` and `linux/ubuntu` should work,
 // but perhaps providing `windows`, `windows/choco` would not.
}
```

Additionally, each os/distribution should be a "provider" following an implementation to the following abstract interface:
```go
type HostPackageManager interface {
  Install(context.Context, ...string) error
}
```
The above interface needs a bit more thought, as @craciunoiuc has mentioned, because versions need to be listed. I can imagine the following could be a way to provide the listing:

`internal/setup/setup_ubuntu.go`:
```go
var DistUbuntu = [...]HostPackage{
  HostPackage{"package",  GT, "1.0.0"},
};
```
That said, I would be surprised if there were no existing Go packages that do not do something similar to what we're trying to achieve.

Additional comments:
 * Please can you use the updated copyright preamble, [here is an example of what it looks like](https://github.com/unikraft/kraftkit/blob/staging/pack/package.go#L1-L4);
 * Use `iostreams.G(ctx).Out` when redirecting information to the user via `fmt.Fprintf`, this is the global access point to `os.Stdout`, which keeps its configuration and possible redirection globally configurable;
 * Check `config.G(ctx).NoPrompt` to determine whether to prompt the user for questions, e.g. `Would you like to continue? [Yn]`;
 * Use `kraftkit.sh/exec` for executing commands and not the built-in `os/exec` package.

cmd/kraft/setup/setup.go
nderjung: Are these requirements actually necessary? I believe the list is now much smaller now. Plus this should be "switchable" the user may not necessarily want qemu :) so a VMM option, determination, should append this or prompt for it

nderjung:
To address the comments by @craciunoiuc, the `get.kraftkit.sh` program can and should recommend additional programs through the relevant package manager, e.g. "APT recommends". Whilst this is the recommended method for installation and any dependencies, a built-in command, `kraft setup`, can not only determine if and which binaries are necessary for a unikraft build, e.g. we can essentially run "`which`" internally, but should also help and aid in "setting up" the host environment. Along these lines, we can additionally check the host for virtualization features. 

In my mind, following an installation from `get.kraftkit.sh`, would be an immediate execution of `kraft setup`, which guides the user through host-setup. In a non-standard install, e.g. by pulling the binary directly from the releases page, the `kraft setup` command enables access to such installation.

I also imagine an extension to `kraft setup` in the future which would guide the user through setting up the global configuration with potential values. We can detect a "first installation" by whether the configuration file has been written. We could inject inside of `cmd/kraft/kraft.go` a check, for example:

```go
import "kraftkit.sh/internal/setup"

// [...]

func main() {
  // [...]
  
  if isFirstSetup, _ := setup.IsFirstUse(ctx); isFirstSetup {
    if err := setup.Setup(ctx); err != nil {
      fmt.Printf("could not perform first-use setup: %v", err)
      os.Exit(1)
    }
  }

  // [...]

  cmdfactory.Main(ctx)
}
```

craciunoiuc:
Yes, you are right. I forgot that many people actually pull the binary from the releases page.

StefanJum:
@nderjung I did a quick refactoring based on your comments, let me know how it looks.

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"kraftkit.sh/cmdfactory"
+	"kraftkit.sh/internal/setup"
+)
+
+type Setup struct {
+	WithOS   string `long:"with-os" usage:"Force the host OS"`
+	WithArch string `long:"with-arch" usage:"Force the architecture of the host"`
+	WithVMM  string `long:"with-vmm" usage:"Force the use of a specific VMM on the host"`
+	WithPM   string `long:"with-pm" usage:"Force the use of a specific package manager on the host"`
```
nderjung, line 20:
```diff
-	WithOS   string `long:"with-os" usage:"Force the host OS"`
-	WithArch string `long:"with-arch" usage:"Force the architecture of the host"`
-	WithVMM  string `long:"with-vmm" usage:"Force the use of a specific VMM on the host"`
-	WithPM   string `long:"with-pm" usage:"Force the use of a specific package manager on the host"`
+	OS   string `long:"os" usage:"Force the host OS"`
+	Arch string `long:"arch" usage:"Force the architecture of the host"`
+	VMM  string `long:"vmm" usage:"Force the use of a specific VMM on the host"`
+	PM   string `long:"pm" usage:"Force the use of a specific package manager on the host"`
```

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func New() *cobra.Command {
+	return cmdfactory.New(&Setup{}, cobra.Command{
+		Short:   "Setup the working environment for building and running unikernels",
+		Use:     "setup",
+		Aliases: []string{"sup"},
```
nderjung, line 27:
```diff
-		Aliases: []string{"sup"},
```

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+		Args:    cmdfactory.MaxDirArgs(0),
```
nderjung, line 28:
```diff
-		Args:    cmdfactory.MaxDirArgs(0),
+		Args:    cobra. NoArgs(),
```

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+		Long: heredoc.Doc(`
+		Setup the working environment for building and running unikernels`),
+		Example: heredoc.Doc(`
+			# Setup the environment
+			$ kraft setup`),
+		Annotations: map[string]string{
+			"help:group": "build",
```
nderjung, line 35:
```diff
-			"help:group": "build",
+			cmdfactory.AnnotationHelpGroup: "build",
```

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+	var err error
+	var sopts []setup.SetupOption
+
+	ctx := cmd.Context()
+
+	if opts.WithOS != "" {
+		sopts = append(sopts, setup.WithOS(opts.WithOS))
+	} else {
+		sopts = append(sopts, setup.WithDetectHostOS())
+	}
```
nderjung, line 50:
```diff
-		sopts = append(sopts, setup.WithOS(opts.WithOS))
-	} else {
-		sopts = append(sopts, setup.WithDetectHostOS())
-	}
+		sopts = append(sopts, setup.WithOS(opts.OS))
+	}
```

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if opts.WithArch != "" {
+		sopts = append(sopts, setup.WithArch(opts.WithArch))
+	} else {
+		sopts = append(sopts, setup.WithDetectArch())
+	}
```
nderjung, line 56:
```diff
-		sopts = append(sopts, setup.WithArch(opts.WithArch))
-	} else {
-		sopts = append(sopts, setup.WithDetectArch())
-	}
+		sopts = append(sopts, setup.WithArch(opts.Arch))
+	}
```

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if opts.WithVMM != "" {
+		sopts = append(sopts, setup.WithVMM(opts.WithVMM))
+	} else {
+		sopts = append(sopts, setup.WithDetectVMM())
+	}
```
nderjung, line 62:
```diff
-		sopts = append(sopts, setup.WithVMM(opts.WithVMM))
-	} else {
-		sopts = append(sopts, setup.WithDetectVMM())
-	}
+		sopts = append(sopts, setup.WithVMM(opts.VMM))
+	}
```

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if opts.WithPM != "" {
+		sopts = append(sopts, setup.WithPM(opts.WithPM))
+	} else {
+		sopts = append(sopts, setup.WithDetectPM())
+	}
```
nderjung, line 68:
```diff
-		sopts = append(sopts, setup.WithPM(opts.WithPM))
-	} else {
-		sopts = append(sopts, setup.WithDetectPM())
-	}
+		sopts = append(sopts, setup.WithPM(opts.PM))
+	}
```

cmd/kraft/setup/setup.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if err := setup.DoSetup(ctx, sopts...); err != nil {
+		return err
+	}
+
+	return err
```
nderjung, line 74:
```diff
-	if err := setup.DoSetup(ctx, sopts...); err != nil {
-		return err
-	}
-
-	return err
+	return setup.Setup(ctx, sopts...)
```

internal/setup/setup.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+)
+
+type Setup struct {
+	WithOS   string `long:"with-os" usage:"Force the host OS"`
+	WithArch string `long:"with-arch" usage:"Force the architecture of the host"`
+	WithVMM  string `long:"with-vmm" usage:"Force the use of a specific VMM on the host"`
+	WithPM   string `long:"with-pm" usage:"Force the use of a specific package manager on the host"`
```
nderjung, line 16:
```diff
-	WithOS   string `long:"with-os" usage:"Force the host OS"`
-	WithArch string `long:"with-arch" usage:"Force the architecture of the host"`
-	WithVMM  string `long:"with-vmm" usage:"Force the use of a specific VMM on the host"`
-	WithPM   string `long:"with-pm" usage:"Force the use of a specific package manager on the host"`
+ os   string
+ arch string
+ vmm  string
+ pm   string
```

internal/setup/setup.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func DoSetup(ctx context.Context, sopts ...SetupOption) error {
+	s := Setup{}
+	for _, opts := range sopts {
```
nderjung, line 21:
```diff
- func DoSetup(ctx context.Context, sopts ...SetupOption) error {
+ func Setup(ctx context.Context, sopts ...SetupOption) error {
    s := Setup{}
+
    for _, opts := range sopts {
```

internal/setup/options.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return arch
+}
+
+func detect_vmm() string {
+	os_virt, role, _ := host.Virtualization()
+	vmm := os_virt + "/" + role
+
+	return vmm
+}
```
nderjung, line 99: Functions in Go are never `snake_case` and are instead `camelBack`.

Additionally, the [`host.Virtualization()`](https://pkg.go.dev/github.com/shirou/gopsutil/v3@v3.23.2/host#Virtualization) method is a nice find. Perhaps it's best to serialize all possible outputs into a global list of supported values.

internal/setup/options.go
```diff
@@ -0,0 +1,103 @@
...
+
+func NotSupported(*Setup) error {
+	return nil
+}
```
nderjung, line 103:
```diff
- func NotSupported(*Setup) error {
- 	return nil
- }
+ const (
+    ErrNotSupportedVMM = errors.New("virtualization system not supported")
+    ErrNotSupportedPM  = errors.New("package manager not supported")
+    ErrNotSupportedOS  = errors.New("OS not supported")
+ )
```
(and place at the top of the file)

cmd/kraft/kraft.go
```diff
@@ -77,6 +78,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(ps.New())
 	cmd.AddCommand(rm.New())
 	cmd.AddCommand(run.New())
+	cmd.AddCommand(setup.New())
```
nderjung, line 81: Move this to the `misc` section

nderjung (changes requested):
Hi @StefanJum, thanks for the update, I can see this is still WIP, however, see comments inline.

craciunoiuc:
@nderjung @StefanJum I think this PR was/is still in its initial phases

I'll convert it to draft until this is picked back up

I know @StefanJum loves Golang so it is going to be a while

craciunoiuc:
Rip this PR 😭

PR #265, opened 2023-03-07
feat(run): append arch/plat to duplicate tgt names

### LINKED ISSUE #256: `kraft run` should display the `plat` and `arch` in the target selection menu when targets aren't named
### Describe the bug

In projects with multiple unnamed targets, like [our C helloworld](https://github.com/unikraft/app-helloworld/blob/RELEASE-0.12.0/kraft.yaml), `kraft run` defaults to showing the name of the project for each target: 

unnamed_targets

### Steps to reproduce

```
git clone https://github.com/unikraft/app-helloworld.git
```

```
cd app-helloworld
```

```
kraft build
```

```
kraft run
```

### Expected behavior

```console
$ kraft run
 > helloworld (plat:linuxu arch:x86_64)
 > some-named-target
 > helloworld (plat:kvm arch:arm64)
```

_No response_

_No response_

### Relevant log output

_No response_

antoineco:
### Prerequisite checklist

### Description of changes

Fixes #256.

When the returned list of targets contains multiple entries with the same name attribute, append the arch/plat to these entries to be able to tell them apart in the output of `kraft run`.

I originally tried to remove the automatic [name injection at load time](https://github.com/unikraft/kraftkit/blob/9b2664e3ec86d3929e7b3c8ed1ffba41d236be59/unikraft/target/transform.go#L24-L26), but there are a few code paths that require targets to have a name. Refactoring that logic sounds too intrusive for a UI change.

---

:warning: This doesn't entirely fix the `kraft run` command, because this code can't possibly do the right thing when duplicate names exist:

https://github.com/unikraft/kraftkit/blob/9b2664e3ec86d3929e7b3c8ed1ffba41d236be59/cmd/kraft/run/run.go#L215-L218

In a future PR, I think we should consider attaching an ID to each target.

---

Result:

krun1
krun2
krun3

### DIFF
--- a/unikraft/app/project.go
+++ b/unikraft/app/project.go
@@ -66,7 +66,7 @@ func NewProjectFromOptions(ctx context.Context, opts ...ProjectOption) (Applicat
 	}
 
 	if len(popts.kraftfiles) < 1 {
-		return nil, fmt.Errorf("no Kraft files specified")
+		return nil, fmt.Errorf("no Kraftfile specified")
 	}
 
 	var all []*application
--- a/cmd/kraft/run/run.go
+++ b/cmd/kraft/run/run.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sort"
 	"syscall"
 
 	"github.com/MakeNowJust/heredoc"
@@ -27,6 +28,7 @@ import (
 	machinedriver "kraftkit.sh/machine/driver"
 	machinedriveropts "kraftkit.sh/machine/driveropts"
 	"kraftkit.sh/unikraft/app"
+	"kraftkit.sh/unikraft/target"
 	"kraftkit.sh/utils"
 )
 
@@ -193,12 +195,12 @@ func (opts *Run) Run(cmd *cobra.Command, args []string) error {
 				return fmt.Errorf("unknown target: %s", opts.Target)
 			}
 
-		} else if len(target) == 0 && len(project.TargetNames()) > 1 {
+		} else if target == "" && len(project.Targets()) > 1 {
 			if config.G[config.KraftKit](ctx).NoPrompt {
 				return fmt.Errorf("with 'no prompt' enabled please select a target")
 			}
 
-			sp := selection.New("select target:", project.TargetNames())
+			sp := selection.New("select target:", selectionTargets(project.Targets()))
 			sp.Filter = nil
 
 			selectedTarget, err := sp.RunPrompt()
@@ -370,4 +372,17 @@ func (opts *Run) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	return nil
+}
+
+// selectionTargets returns the given target.Targets in a format suitable for
+// interactive prompts.
+func selectionTargets(tgts target.Targets) []string {
+	tgtStrings := make([]string, 0, len(tgts))
+	for _, tgt := range tgts {
+		tgtStrings = append(tgtStrings, fmt.Sprintf("%s (%s)", tgt.Name(), tgt.ArchPlatString()))
+	}
+
+	sort.Strings(tgtStrings)
+
+	return tgtStrings
 }
--- a/unikraft/app/merge.go
+++ b/unikraft/app/merge.go
@@ -20,6 +20,7 @@ package app
 import (
 	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
+
 	"kraftkit.sh/unikraft/core"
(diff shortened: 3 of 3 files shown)


unikraft/app/project.go
```diff
@@ -66,7 +66,7 @@ func NewProjectFromOptions(ctx context.Context, opts ...ProjectOption) (Applicat
 	}
 
 	if len(popts.kraftfiles) < 1 {
-		return nil, fmt.Errorf("no Kraft files specified")
+		return nil, fmt.Errorf("no Kraft file specified")
```
nderjung: `Kraft file` => `Kraftfile`

nderjung (changes requested):
Awesome, thanks! One minor suggestion, see in-line.

Btw, I have a fix for kraft run for QEMU. I was yak-shaving to get another partially related issue upstream first but i'll push the changes for the fix you mentioned today.

Edit: Perhaps we should _always_ suffix the name with the plat-arch combination, to keep the output consistent?

Edit 2: Could you add `GitHub-Fixes: #256` to the commit message body?

Thanks!

antoineco:
@nderjung I agree with your comments. I just addressed them in [03ad77de..4a4493e6](https://github.com/unikraft/kraftkit/compare/03ad77deb0c364f336290924d75f9e7fb57c1c86..4a4493e61995f138215e00a37336878709d26cb5).

Result:

image

unikraft/target/target.go
```diff
@@ -74,8 +74,14 @@ type TargetConfig struct {
...
 }
 
+var _ fmt.Stringer = (*TargetConfig)(nil)
+
 type Targets []TargetConfig
 
+func (tc TargetConfig) String() string {
+	return fmt.Sprintf("%s (%s)", tc.name, tc.ArchPlatString())
+}
+
```
nderjung: In this PR, the returned value from `String()` is actually the "human-readable" version of the target name, and not the "canonical name". The canonical name is actually `<name>_<plat>-<arch>`, as this is the binary name built by Unikraft, and this should be the returned value since it's appropriate across more of the system. Either the human-readable version is kept locally in the `cmd/kraft/run` namespace or the canonical name is the one that's used by implementing this `String` return.

To approach this, I'd also split it into its own commit. I'm wondering also whether it makes sense to propagate the `String`/`fmt.Stringer` interface to `unikraft.Nameable` ? Since then we could then provide `String` representations for all components. They would be implementation-specific of course, still, like this for targets. 

This would then alleviate the use of the wrapper method [`unikraft.TypeNameVersion`](https://github.com/unikraft/kraftkit/blob/staging/unikraft/type.go#L111), which would return an incorrect value if wrapping a `target.Target` input which is `unikraft.Nameable`.

nderjung (changes requested):
You've opened up a can of worms! (maybe, just a small can) See in-line.

unikraft/target/target.go
antoineco: I'll keep the UI logic in `cmd` then, to avoid letting the implementation details leak into the UI logic.

nderjung: Ok, I'll add an issue which focuses on providing `Stirng` as an additional mandatory interface in `unikraft.Nameable` and deprecating `unikraft.TypeNameVersion`.

antoineco: See [c9f2aca1..d7ed9336](https://github.com/unikraft/kraftkit/compare/c9f2aca1bf1aabd685e18e65051ae6016a0631d7..d7ed9336b8c74d70e1f3d2a2f9155efd86dc9872)

nderjung (approved):
Works, thanks! :)

unikraft/target/target.go
nderjung: Issue rendered in #268

nderjung:
I should not have merged this yet 😅 optimising merging at its finest.

I can no longer run the target based on the selection.

antoineco:
For me it was always broken. As mentionned in the warning, `project.TargetByName(target)` takes a name, but when this name is always `helloworld, helloworld, helloworld, ...` you rely on the fact that your target will always be the first match. This is (and already was before this PR) a broken behaviour.

nderjung:
Right. I'll have a look with the fixes i have for `kraft run` in-general. The work i was doing was specifically related to outputting the logging and managing the event listener which is broken too.

Looks like we should definitely use canonical name in the return so that we can easily and reliably look it up.

In which case, ignore about optimistic merging, this PR is just extremely focused :D

antoineco:
I agree. Without a unique identifier `kraft run` can't run as-is.

PR #278, opened 2023-03-09
Fix instantiating unikernels and tailing logs

nderjung:
### Prerequisite checklist

### Description of changes

This makes several changes across subsystems to fix the instantiation and tailing of unikernels via `kraft run`. A number of changes were required, particularly:

 - Deprecating the use of a UNIX socket for the logs of a QEMU instance -- since there could be a synchronization issue if the consumer was not fast enough to start receiving logs. Instead, the logs are written verbatim to log file which itself has been made available in `machine.MachineConfig`. Necessary adjustments are made to facilitate "tailing" from a log file from QEMU. This has the added benefit being able to retrieved after the unikernel has ended its life.
 - Fixing the instantiation, tailing and monitoring of a unikernel in `kraft run`. The instantiation of the events monitor is made to be completely silent such that sub-process noise does not spill over into the executor's parent's stdout. The actual instantiation of the unikernel via `driver.Start()` is moved as late-as-possible such that go-routines can be established first. Additionally, the manipulation of the cancalable-context is fixed with respect to the fixed `TailWriter`.
 - Introduce an optional CLI argument for setting the name of the unikernel;
 - Introduce an optional CLI argument for setting the initrd (CPIO) file for the unikernel.

### How to test

I used a simple "Hello, World" C unikernel, with 1 target (kvm-x86_64) and a simple loop in `main.c`, e.g.:

```c
#include <stdio.h>
#include <time.h>
#include <errno.h>

static void millisleep(unsigned int millisec)
{
        struct timespec ts;
        int ret;

        ts.tv_sec = millisec / 1000;
        ts.tv_nsec = (millisec % 1000) * 1000000;
        do
                ret = nanosleep(&ts, &ts);
        while (ret && errno == EINTR);
}


int main(int argc, char *argv[])
{
        for (;;) {
                printf("Hello world!\n");
                millisleep(1000);
        }

        return 0;
}
```

Once built as a unikernel, I could do the following:

```
# this will instantiate `hello1` and immediately detach.
kraft run --detach --name hello1

# i can now see there `kraft events` and `qemu-system-x86_64` running in the background

# starting another unikernel will start to tail the logs
kraft run --name hello2

# Ctrl+C will stop tailing

# i can kill them both
kraft stop --all

# see that they are both exited
kraft ps

# clean up
kraft rm --all

# no lingering process should be on the system (aka `kraft events [...]`)
```

### DIFF
--- a/cmd/kraft/events/events.go
+++ b/cmd/kraft/events/events.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 // Copyright (c) 2022, Unikraft GmbH and The KraftKit Authors.
 // Licensed under the BSD-3-Clause License (the "License").
-// You may not use this file expect in compliance with the License.
+// You may not use this file except in compliance with the License.
 package events
 
 import (
@@ -102,7 +102,7 @@ func (opts *Events) Run(cmd *cobra.Command, args []string) error {
 		cancel()
 	}()
 
-	// TODO: Should we thrown an error here if a process file already exists? We
+	// TODO: Should we throw an error here if a process file already exists?  We
 	// use a pid file for `kraft run` to continuously monitor running machines.
 
 	// Actively seek for machines whose events we wish to monitor.  The thread
--- a/cmd/kraft/events/events.go
+++ b/cmd/kraft/events/events.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"sync"
 	"syscall"
 	"time"
 
@@ -20,6 +19,7 @@ import (
 
 	"kraftkit.sh/cmdfactory"
 	"kraftkit.sh/config"
+	"kraftkit.sh/internal/waitgroup"
 	"kraftkit.sh/log"
 	"kraftkit.sh/machine"
 	machinedriver "kraftkit.sh/machine/driver"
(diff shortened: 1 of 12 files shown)


antoineco (approved):
Looking great!

craciunoiuc (changes requested):
Some things from my side:

```
 1. `kraft stop X` - does not also clean the `.pid` - this makes subsequent calls to `kraft ps` fail.
 2. Ctrl^C should not work and Ctrl^D should stop the unikernel.
 3. `kraft stop/rm --all` should exist
 4. `kraft rm X` only removes an unikernel from the tracked list - if you forget to run `kraft stop` before it is lost forever - `kraft rm` should ask the user if they want to also stop
 5. `kraft events` should have a configurable polling rate and by default poll once a second/half-second
 6. `kraft ps` should run in less than one second
 7. killing an unikernel should make it eventually be removed from `kraft ps` -> 30s timeout maybe?
 8. `kraft stop/rm/etc.` commands should auto-complete ids when they can (so you don't have to input the entire id)
```

In commit: `fix(qemu): tail logs from file`: `porclaining` ->`porcelaining`
In commit: `fix(run): correctly handle executing and tailing logs of a unikernel` `cancalable` -> `cancelable`

nderjung:
Thanks for the feedback @craciunoiuc, i'll take a look :)

craciunoiuc (changes requested):
A few things from my side:

1. Running the `kraft run` command on helloworld, without `--rm` makes the unikernel stop before the kraft can check for it. This puts everything in an 'undefined' state where it just spams:
```
could not stop 1a079215293cce4bb0e7b4e088c847735ccef7563d20d4c88ace18e7b14a0257: dial unix /var/kraftkit/1a079215293cce4bb0e7b4e088c847735ccef7563d20d4c88ace18e7b14a0257_control.sock: connect: no such file or directory
```

Afterwards `kraft ps` and `kraft run` no longer work until you delete `/var/kraftkit`.

My suggestion is to treat the special case of fast exiting VMs, or just switch from `--rm` to `--no-rm`.

2. Should you be able to see logs from stopped machines?

3. Should machines with an exited state show up in in `kraft ps --all`? I can't seem to find them, probably because they exit too fast. Update: Even the ones that I can see no longer show afterwards:
```
➜  doom git:(staging) ✗ sudo ~/kraftkit/dist/kraft ps --all
MACHINE ID    IMAGE                     ARGS  CREATED         STATUS   MEM
838d86fcab74  kernel://doom_kvm-x86_64        10 seconds ago  running  0MB
➜  doom git:(staging) ✗ sudo ~/kraftkit/dist/kraft ps --all
MACHINE ID  IMAGE  ARGS  CREATED  STATUS  MEM
```
4. With `kraft run` the serial interface becomes uni-directional. I can no longer do `scanf` inside the code. This happens probably because it's writing to a file and then just spewing out the file to the console, so you can no longer interact.
Because of this, the unikernel appears to be hanging, but it's just waiting for input.

I think this is enough for now

craciunoiuc (approved):
All good here. Thanks!

PR #310, opened 2023-03-29
fix(kraft): Fix checking for an update

nderjung:
### Prerequisite checklist

### Description of changes

This PR fixes a race condition with KraftKit's use of contexts.  The use of `Pre` is not possible with the root command because the context is populated in `Run` and so invocations and access to the provided `ctx` returns unfufilled values.  In this case, an attempt to retrieve the latest version of KraftKit is made in the `Pre` method.  Since this invocation relies on the `ConfigManager` to be initialized, it has insufficient knowledge to continue.  The result of invoking the `ConfigManager` via `ctx` before its instantiation is that default values would not be properly fufilled.  Following commands, such as `kraft pkg update` would be improperly set.

Additionally, if the version is unset, a version check should not occur.

This PR therefore moves the check for an update after the instantiation of the general context for the main programs invocation and prevents any check from occurring if the version is unknown.

### DIFF
--- a/cmd/kraft/kraft.go
+++ b/cmd/kraft/kraft.go
@@ -86,15 +86,6 @@ func New() *cobra.Command {
 	return cmd
 }
 
-func (k *Kraft) PersistentPre(cmd *cobra.Command, args []string) error {
-	ctx := cmd.Context()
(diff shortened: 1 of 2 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #371, opened 2023-04-16
fix(kraft): Provide access to PackageManager in relevant commands

### LINKED ISSUE #366: Warning about non-initialized package manager in commands that do not require a package manager
### Describe the bug

This is a purely UX-related issue.

Upon executing CLI commands, the user is always greeted with the warning message below if a package manager handler isn't configured (which is the case in the default config):

```console
$ kraft version
[ W ] could not initialize package manager: could not determine handler [format]=oci
kraft v0.5.0 (16f905f) go1.20.3 Sun Apr 16 13:13:25 CEST 2023
```

This is misleading, especially to new users, and we should think about a way to avoid initializing the package manager unless required.

### Steps to reproduce

```
rm ~/.config/kraftkit/config.yaml
kraft version
```

### Expected behavior

No warning, since I didn't perform any action which requires a package manager.

_No response_

_No response_

### Relevant log output

_No response_

nderjung:
### Prerequisite checklist

### Description of changes

This PR migrates the instantiation of the umbrella package manager to subcommands which require access.  This focus prevents commands which would otherwise not use the package manager from calling additional initializers.

GitHub-Fixes: #366

### DIFF
--- a/cmd/kraft/build/build.go
+++ b/cmd/kraft/build/build.go
@@ -68,6 +68,18 @@ func New() *cobra.Command {
 	})
 }
 
+func (*Build) Pre(cmd *cobra.Command, _ []string) error {
+	ctx := cmd.Context()
+	pm, err := packmanager.NewUmbrellaManager(ctx)
+	if err != nil {
+		return err
+	}
+
+	cmd.SetContext(packmanager.WithPackageManager(ctx, pm))
+
+	return nil
+}
+
 func (opts *Build) Run(cmd *cobra.Command, args []string) error {
 	var err error
 	var workdir string
--- a/cmd/kraft/clean/clean.go
+++ b/cmd/kraft/clean/clean.go
@@ -38,6 +38,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"kraftkit.sh/cmdfactory"
+	"kraftkit.sh/packmanager"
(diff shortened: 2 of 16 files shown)


cmd/kraft/build/build.go
```diff
@@ -68,6 +68,18 @@ func New() *cobra.Command {
 	})
 }
 
+func (opts *Build) Pre(cmd *cobra.Command, args []string) error {
```
antoineco: _(putting my linter hat on)_ `opts` is unused. It's not a major issue, but if we don't remove it now, the `unused` linter (which I think we should enable) will eventually ask us to do so.

antoineco (commented):
The approach is simple enough to avoid the warning in top level commands like `kraft -h`, `kraft version` 👍 

I only have one non-blocking comment about unused variables.

cmd/kraft/build/build.go
nderjung: Nice hat! 🤠 Are you saying this is possible?

```go
func (_ *Build) Pre(cmd *cobra.Command, _ []string) error {
```

? I've never seen `func (_ *XXX)` before, but if this is possible, sure it make sense

antoineco: Just remove `opts` entirely.

antoineco (approved):

PR #390, opened 2023-04-18
feat: Add basic login method

nderjung:
### Prerequisite checklist

### Description of changes

This PR introduces a new subcommand `kraft login` which manipulates KraftKit's configuration file such that it populates the list of authenticated services.

### DIFF
--- a/cmd/kraft/login/login.go
+++ b/cmd/kraft/login/login.go
@@ -1,0 +1,86 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2022, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+package login
+
+import (
+	"bufio"
+	"fmt"
+	"strings"
+
+	"golang.org/x/term"
+	"kraftkit.sh/cmdfactory"
+	"kraftkit.sh/config"
+	"kraftkit.sh/iostreams"
+
+	"github.com/spf13/cobra"
+)
+
+type Login struct {
+	User  string `long:"user" short:"u" usage:"Username" env:"KRAFTKIT_LOGIN_USER"`
+	Token string `long:"token" short:"t" usage:"Authentication token" env:"KRAFTKIT_LOGIN_TOKEN"`
+}
+
+func New() *cobra.Command {
+	return cmdfactory.New(&Login{}, cobra.Command{
+		Short: "Provide authorization details for a remote service",
+		Use:   "login [FLAGS] HOST",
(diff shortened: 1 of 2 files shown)


cmd/kraft/login/login.go
```diff
@@ -0,0 +1,76 @@
...
+		Args:  cobra.ExactArgs(1),
+		Annotations: map[string]string{
+			cmdfactory.AnnotationHelpGroup: "misc",
+		},
+	})
+}
+
+func (opts *Login) Pre(cmd *cobra.Command, args []string) error {
+	if opts.User == "" && opts.Token == "" {
+		return fmt.Errorf("must speficy at least `--user` or `--token`")
```
craciunoiuc:
```suggestion
		return fmt.Errorf("must specify at least `--user` or `--token`")
```

cmd/kraft/login/login.go
```diff
@@ -0,0 +1,76 @@
...
+		Use:   "login [FLAGS] HOST",
+		Args:  cobra.ExactArgs(1),
+		Annotations: map[string]string{
+			cmdfactory.AnnotationHelpGroup: "misc",
+		},
+	})
+}
+
+func (opts *Login) Pre(cmd *cobra.Command, args []string) error {
+	if opts.User == "" && opts.Token == "" {
```
craciunoiuc: This only works if the provided argument `--token` is `""`. If left empty it will just assume that the password/token is the next argument.

craciunoiuc: Also setting through env variables does not currently work it seems

The username is caught correctly, but the token is not

craciunoiuc (changes requested):
Hey @nderjung 

I tried to use this to `pull` a private repository.

Not sure if I did something wrong but It did not work. The values are set correctly in the config, but are not used by the pulling mechanism

nderjung:
Hi @craciunoiuc, thanks for the feedback.  The fact that Git cannot be authenticated is a separate issue to this.  This is simply adding an auth entry to the configuration file.

cmd/kraft/login/login.go
nderjung: Good catch, thanks!

cmd/kraft/login/login.go
nderjung: The goal here is to XOR `User` and `Token`.  Env var should work, can you double check? It's definitely set in the structure tag

nderjung: The two usages are
```
kraft -t TOKEN domain.com
```
and
```
kraft -u USER -p PASS domain.com
```

craciunoiuc (approved):
All good here. Thanks!

Issue #370, opened 2023-04-16
Add end-to-end tests for CLI commands


antoineco:
## Feature request summary

There is a lot to be done in terms of testing. For a start, we could focus on high level "e2e" tests and execute each command of the CLI tool against a sample unikernel project.

We could start with one project + one test case per command, then expand as we see fit.

This doesn't provide a feedback loop as quick as unit tests, but would already allow us to catch major regressions in the CI.

```[tasklist]
### E2E tests
- [ ] #811
- [ ] #810
- [ ] #809
- [ ] #808
- [ ] #392
- [ ] #405
- [ ] https://github.com/unikraft/kraftkit/issues/812
- [ ] https://github.com/unikraft/kraftkit/issues/813
- [ ] https://github.com/unikraft/kraftkit/issues/840
```

antoineco:
I think our next focus should be on `kraft build`.

It would be very nice to have an initial test that just 
- initializes a temporary project from a few fixture files (`helloworld.c` might even be enough?)
- writes a `kraft.yaml` in that project
- pulls the dependencies
- builds a few targets

That would cover several subsystems at once. An ideal case for an e2e test.

priyanshsao:
Hello @antoineco , I am Priyansh Sao
I am a computer science student currently in my junior year, who is experienced in the Go programming language but wants to gain complete testing knowledge using Go.

I would like to contribute to `Kraftkit` through GSOC.
If there are any beginning problems or good initial problems associated with this project that I can contribute to while I'm getting familiar with `Kraftkit` and testing with Ginko & Gomega.

This would help me gain experience and make it easier for mentors to evaluate my knowledge.

PR #396, opened 2023-04-21
test(e2e): add helpers to read/write from/to config

antoineco:
### Prerequisite checklist

### Description of changes

Part to #370

Allows each test case to have its own ephemeral configuration file, with which it can conveniently interact by reading/writing arbitrary values before executing `kraft` commands.

The included tests should be self-explanatory.
_(I wrote tests for the test framework, and am currently resisting the urge to post a "yo dawg" meme about it.)_

---

**Notes**

1. I originally wanted the `Write()` method to accept a YAML path and a Go value, but figured that avoiding conversions between the Go and YAML type systems was both:

   - simpler to implement, especially due to potentially nested YAML nodes
   - more appropriate for arbitrary manipulations of the YAML config, for example to set multiple values in a single write

   The downside is that test authors need to understand how to use [yaml.Filters](https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml@v0.14.1/yaml#Filter) and how they are executed in a [yaml.Pipe](https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml@v0.14.1/yaml#Pipe).

   As the number of CLI tests grows, we hopefully will see some patterns emerge and decide to revisit that approach.

1. I usually advocate for writing unit tests using Go's bare `testing` package, and reserve `ginkgo` for the more complex behaviour-driven things that require setup and teardown.

   In this case though, I used `ginkgo` for the unit tests as well. The reason is that the helpers included in the test framework embed a few `ginkgo` functions, so it's easier to test them in the context of `ginkgo` runs.


nderjung (approved):
Awesome! Thanks a lot!

Issue #4, opened 2022-06-14
Configurable output format


nderjung:
This issue tracks adding a configurable output format engine within KraftKit and related commands which produce an informational output, for example `kraft pkg ls` or `kraft pkg info`.

Task list:

 - [ ] Investigate how `kubectl get -o`'s output formatter works, it is quite configurable, e.g.:
   ```
   -o, --output='': Output format. One of:
       json|yaml|wide|name|custom-columns=...|custom-columns-file=...|go-template=...|go-template-file=...|jsonpath=...|jsonpath-file=...
   ```
 
- [ ] The implementation should migrate the use of the existing tool [`tableprinter`](https://github.com/unikraft/kraftkit/blob/staging/utils/table_printer.go) and use instead a more relevant library for outputting formatted data.

- [ ] A global option should be introduced into `kraftkit.sh/config` (`Output`) such that the format can be switched;

- [ ] Log output format be adjusted to the desired format when its set to `json`, overriding `log.type`.

- [ ] All tabular data should use a common interface such that its visible output can be adjusted dynamically and without requiring each instance to be reworked.

mdsahiloss:
@nderjung I would like to work on it, Can you assign it to me :)

nderjung:
Hi @MdSahil-oss please feel free to tackle the issue! When ready, please submit a PR or feel free to reach out on Discord to discuss further.

mdsahiloss:
Hi! @nderjung I've started working on it :)

mdsahiloss:
Note: **How `kubectl get -o` is implemented in kubectl :- https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get.go#L194**

nderjung:
Upon implementation, tests should default to JSON output for better parsing. See https://github.com/unikraft/kraftkit/pull/405/files#r1180682845

PR #405, opened 2023-04-24
test(e2e): add CLI test for pkg update

antoineco:
### Prerequisite checklist

### Description of changes

Part of #370

Tests that, given an empty manifests directory, `kraft pkg update` retrieves the list of components, libraries and packages.
Also tests that the command can print its own help.

### DIFF
--- a/test/e2e/cli/pkg_test.go
+++ b/test/e2e/cli/pkg_test.go
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2023, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package cli_test
+
+import (
+	. "github.com/onsi/ginkgo/v2" //nolint:stylecheck
+	. "github.com/onsi/gomega"    //nolint:stylecheck
+
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+
+	fcmd "kraftkit.sh/test/e2e/framework/cmd"
+	fcfg "kraftkit.sh/test/e2e/framework/config"
+	. "kraftkit.sh/test/e2e/framework/matchers" //nolint:stylecheck
+)
+
+var _ = Describe("kraft pkg", func() {
+	var cmd *fcmd.Cmd
+
+	var stdout *fcmd.IOStream
+	var stderr *fcmd.IOStream
+
+	var cfg *fcfg.Config
+
+	BeforeEach(func() {
+		stdout = fcmd.NewIOStream()
+		stderr = fcmd.NewIOStream()
+
+		cfg = fcfg.NewTempConfig()
+
+		cmd = fcmd.NewKraft(stdout, stderr, cfg.Path())
+		cmd.Args = append(cmd.Args, "pkg")
+	})
+
+	_ = Describe("update", func() {
+		var manifestsPath string
+
+		BeforeEach(func() {
+			cmd.Args = append(cmd.Args, "update")
+
+			manifestsPath = yaml.GetValue(cfg.Read("paths", "manifests"))
+			Expect(manifestsPath).To(SatisfyAny(
+				Not(BeAnExistingFile()),
+				BeAnEmptyDirectory(),
+			), "manifests directory should either be empty or not yet created")
+		})
+
+		Context("implicitly using the default manager type (manifest)", func() {
+			When("invoked without flags or positional arguments", func() {
+				It("should retrieve the list of components, libraries and packages", func() {
+					err := cmd.Run()
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(stderr.String()).To(BeEmpty())
+					// The command sends ANSI escape sequences while updating, such as `\e[2K` (erase entire line).
+					// References:
+					//   https://www.regular-expressions.info/nonprint.html
+					//   https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797
+					Expect(stdout.String()).To(MatchRegexp(`\x1b\[2K\[\+\] Updating\.\.\. \[\d+\.\d+s\]\r\n`), "Quoted output: %q", stdout)
+
+					Expect(manifestsPath).To(ContainFiles("index.yaml", "unikraft.yaml"))
+					Expect(manifestsPath).To(ContainDirectories("libs"))
+				})
+			})
+
+			When("invoked with the --help flag", func() {
+				BeforeEach(func() {
+					cmd.Args = append(cmd.Args, "--help")
+				})
+
+				It("should print the command's help", func() {
+					err := cmd.Run()
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(stderr.String()).To(BeEmpty())
+					Expect(stdout).To(MatchRegexp(`^Retrieve new lists of Unikraft components, libraries and packages.\n`))
+				})
+			})
+		})
+	})
+})
--- a/test/e2e/framework/matchers/matchers.go
+++ b/test/e2e/framework/matchers/matchers.go
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2023, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+// Package matchers contains additional Gomega matchers.
+package matchers
+
+import "github.com/onsi/gomega/types"
+
+// BeAnEmptyDirectory succeeds if a file exists and is a directory that does
+// not contain any file.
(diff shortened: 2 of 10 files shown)


test/e2e/framework/matchers/matchers.go
```diff
@@ -0,0 +1,33 @@
...
+	return &containFilesMatcher{fileNames: files}
+}
+
+// ContainDirectories succeeds if a directory exists and contains
+// sub-directories with the provided names.
+// Actual must be a string representing the absolute path to the directory
+// containing these sub-directories.
+func ContainDirectories(dirs ...string) types.GomegaMatcher {
+	return &containDirectoriesMatcher{dirNames: dirs}
+}
```
antoineco, line 33: I should add simple test cases for those too, just to ensure that our test helpers are asserting the right thing.

antoineco (commented):
~Holding until I add these unit tests.~
**Ready for review**

craciunoiuc (approved):
All good here. Thanks!

Took a deep look and it seems fine + it seems that the test passed

test/e2e/cli/pkg_test.go
```diff
@@ -0,0 +1,81 @@
...
+			It("should retrieve the list of components, libraries and packages", func() {
+				err := cmd.Run()
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(stderr.String()).To(BeEmpty())
+				// The commands sends control characters while updating.
+				Expect(stdout.String()).To(MatchRegexp(`\[\+\].+Updating\.\.\. \[\d+\.\d+s\]\r\n`))
+
+				Expect(manifestsPath).To(ContainFiles("index.yaml", "unikraft.yaml"))
+				Expect(manifestsPath).To(ContainDirectories("libs"))
```
nderjung: These checks are manifest-specific artifacts. Whilst the default functionality of kraftkit is to use manifest Package Manager, it can be configured to use a different ones which would result in alternative output artefacts from `kraft pkg update`. For example, after the introduction of OCI package manager, sourcing from `unikraft.org` does not actually produce any system artefacts.

Perhaps just for greater clarity of the test(s) itself the package manager should be indicated? The flag `-m` should then be used to specifically set the manifest manager and the test name also includes this. 

I can then see additional tests that [try out different providers](https://github.com/unikraft/kraftkit/blob/staging/manifest/provider.go#L34-L91), e.g., starting from a blank list of unikraft manifests:
```
# e2e test 1: GitHub provider
kraft pkg source https://github.com/unikraft/unikraft.git
kraft pkg update
# only a `unikraft.yaml` artefact

# e2e test 2: GitProvider
[...]

# e2e test N: ManifestIndexProvider
kraft pkg source https://manifests.kraftkit.sh/index.yaml
# `index.yaml` and any additional "symlinks"
[...]
```

The current functionality assumes (afaik) that the config will be populated with its default configuration and values, meaning that the manifest provider is an [ManifestIndexProvider](https://github.com/unikraft/kraftkit/blob/staging/manifest/index.go#L33) set to [the default location](https://github.com/unikraft/kraftkit/blob/staging/config/defaults.go#L17).

antoineco: How about having both? Test with a default config, and test with selected manifests?

We want to catch any breaking user facing change, and a change of default is exactly that, that's why I prioritized the default case here.

I'll be busy with other topics for the next few weeks most likely, so we can either block this PR until all cases are added, or benefit from these initial tests and iterate on additional cases later.

antoineco: I just nested the existing test cases inside a context block (see [eedbbb1f..aa9bfc3b](https://github.com/unikraft/kraftkit/compare/eedbbb1f2bc921725d4822cd042166da9efc2bd0..aa9bfc3bbb65259a98466662ab7d62518456aafb)):

```go
Context("implicitly using the default manager type (manifest)", { /* ... */ })
```

This should make it explicit that we are testing with plain defaults, and that future context blocks will contain manager-specific test cases, alongside the default case.

PTAL

nderjung: Your suggestion of updating the test context helps assure that this test is intended for "default" functionality and addresses my concern. Thanks a tonne!

I will open up a separate issue addressing the permutations that can be made by the `--manager` flag which is out-of-scope for this initial e2e of `kraft pkg update`.

test/e2e/cli/pkg_test.go
```diff
@@ -0,0 +1,86 @@
...
+					Expect(manifestsPath).To(ContainDirectories("libs"))
+				})
+			})
+
+			When("invoked with the --help flag", func() {
+				BeforeEach(func() {
+					// The help subsystem is managed by cobra and fails when top-level flags
+					// are passed, so we ensure to keep only the command name and subcommand
+					// from the original cmd.
+					cmd.Args = []string{cmd.Args[0], cmd.Args[len(cmd.Args)-2], cmd.Args[len(cmd.Args)-1], "--help"}
```
nderjung: Non-blocking but this seems quite convoluted approach of injecting the `--help` flag (though the comment preamble does help) and will not scale if the parent `BeforeEach` includes any additional positional arguments or flags.

test/e2e/cli/pkg_test.go
```diff
@@ -0,0 +1,86 @@
...
+				})
+
+				It("should print the command's help", func() {
+					err := cmd.Run()
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(stderr.String()).To(BeEmpty())
+					Expect(stdout).To(MatchRegexp(`^Retrieve new lists of Unikraft components, libraries and packages.\n`))
```
nderjung, line 78: Minor:
```diff
- 					Expect(stdout).To(MatchRegexp(`^Retrieve new lists of Unikraft components, libraries and packages.\n`))
+ 					Expect(stdout).To(MatchRegexp(`^Retrieve new lists of Unikraft component packages.\n`))
```

test/e2e/cli/pkg_test.go
```diff
@@ -0,0 +1,86 @@
...
+				It("should retrieve the list of components, libraries and packages", func() {
+					err := cmd.Run()
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(stderr.String()).To(BeEmpty())
+					// The command sends ANSI escape sequences while updating, such as `\e[2K` (erase entire line).
+					// References:
+					//   https://www.regular-expressions.info/nonprint.html
+					//   https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797
+					Expect(stdout.String()).To(MatchRegexp(`\x1b\[2K\[\+\] Updating\.\.\. \[\d+\.\d+s\]\r\n`), "Quoted output: %q", stdout)
```
nderjung, line 61: This will be eased via https://github.com/unikraft/kraftkit/issues/4 where JSON can be used as the global output format for all tests. Adding a reference to this comment in aforementioned issue.

test/e2e/cli/pkg_test.go
antoineco: Fully agree, this is based on the assumption that we are using a subcommand that is n-deep exactly. Mostly a workaround, the fact that we even need to avoid passing any other flag is a bug. I'll open an issue.

test/e2e/cli/pkg_test.go
antoineco: Hm, if we change the regexp that way it won't match the the help message anymore, and the test will fail.

Do you mean that we should update the help in `cmd/kraft/pkg`?

test/e2e/cli/pkg_test.go
antoineco: I removed that hack entirely from here and it works.
The `--config-dir` flag doesn't cause any issue with `kraft pkg update --help`, but it still does with `kraft version --help` 🤔

Created #430

test/e2e/cli/pkg_test.go
antoineco: Even if we test for alternative output formats, we should always _always_ test for the default output too. The most visible regressions are the ones associated with settings used by the majority of users.

I wouldn't include permutations of all output formats in every single test though. I think we should pick one command with interesting output, and test permutations just for that command. We could even create meta test suites for these kind of global behaviours (output, input parsing, ...).

test/e2e/cli/pkg_test.go
antoineco: I removed that hack entirely and it works with the current staging branch.
The `--config-dir` flag doesn't cause any issue with `kraft pkg update --help`, but it still does with `kraft version --help` 🤔

#427 fixes the issue, even with `kraft version --help`.

test/e2e/cli/pkg_test.go
nderjung: Ah, I see. You're right, the help message should be updated. This can be addressed in a separate PR/issue.

test/e2e/cli/pkg_test.go
nderjung: 100% agree.

nderjung (approved):
Awesome! Looks good!

PR #407, opened 2023-04-25
build: remove Go workspace files

antoineco:
### Prerequisite checklist

### Description of changes

The go.work file is currently misused; it references standalone tool modules which aren't imported by the kraftkit module.

The 'replace' directive inside kraftkit's go.mod file is also superfluous; the entry inside go.work is sufficient for kraftkit to locate the vendored revision.

Ref. https://go.dev/ref/mod#workspaces

```diff
+9
-1,003 
```

🙂

---

### Update

Go workspace files removed entirely. See commit message(s) for the rationale.


craciunoiuc:
Would this also fix `go.work.sum` always updating itself? 😅

antoineco:
@craciunoiuc it might, since the only local module required by kraftkit is git2go, which version is not changing due to manual vendoring.

In fact, we could use a simple `require` inside `go.mod` pointing at the vendored module instead of using a `go.work` file, and things would keep working just fine.

`go.work` is a nice way to iterate on local dependencies without having to push these to Git (like a project-global `replace`), but I've seen several people recommend to exclude it from version control, among them [GitHub themselves](https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/Go.gitignore#L20-L21) (https://github.com/github/gitignore/pull/3884).

craciunoiuc:
Okay makes sense

Also looked over how `go.work` should be used and this PR seems a good idea 👍

antoineco:
Update: I removed the Go workspace files in a second commit. After reading extensively about this feature, it is clear to me that putting `go.work` under version control is both a misuse and an overhead.

nderjung (approved):
`go.work` (and its byproduct `go.work.sum`) are artifacts from creating the `vendor/` directory which was previously renamed to `third_party/`. This change removes a now unnecessary file (and good riddance!) Thanks a lot!

PR #412, opened 2023-04-25
ci(e2e): fix missing build time kraft version

### LINKED ISSUE #411: bug(e2e): `kraft version` fails in GitHub actions workflow
### Describe the bug

The e2e test for `kraft version` fails since the merge of #410. This is due to the fact that `make kraft` introduces a longer (more complex) version pattern that the e2e test is not capable of parsing, which results in a failure.

An argument for a solution/fix is to make the test simpler such that it simply detects that a single line starting with `kraft` is emitted after invoking `kraft version`.

### Steps to reproduce

_No response_

### Expected behavior

The e2e test for `kraft version` should pass with binaries which are built via `make kraft`.

_No response_

_No response_

### Relevant log output

_No response_

antoineco:
### Prerequisite checklist

### Description of changes

The [fetch depth](https://github.com/actions/checkout/blob/f095bcc56b7c2baf48f3ac70d6d6782f4f553222/action.yml#L56-L57) is too shallow for `git describe --tags` to be able to return anything, so we set the `VERSION` variable manually.

This could be something to pay attention to in release workflows in general. We are [already addressing it](https://github.com/unikraft/kraftkit/blob/8fb2ec33764509657df1ceb1c8a8856b06865e83/.github/workflows/release-stable.yaml#L19-L20), but I think the current regex is still a good safeguard against small regressions.

GitHub-Fixes: #411

### DIFF
--- a/.github/workflows/e2e-cli.yaml
+++ b/.github/workflows/e2e-cli.yaml
@@ -41,9 +41,9 @@ jobs:
         run: go install github.com/onsi/ginkgo/v2/ginkgo
 
       - name: Install kraft
-        env:
-          DOCKER: ''
-        run: make kraft DISTDIR="$(go env GOPATH)"/bin
+        run: make kraft DOCKER= VERSION=v0.0.0-ci-e2e DISTDIR="$(go env GOPATH)"/bin
 
       - name: Run e2e tests
+        env:
+          KRAFTKIT_NO_CHECK_UPDATES: true
         run: ginkgo -v -p -randomize-all ./test/e2e/cli/


nderjung:
Hi, this does not address the test locally? (On a dirty tree?). For example, i may have local commits but still a dirty tree and the test will fail since the version, may, for example, return:

```
kraft v0.5.0-93-gb709d64-dirty (b709d64) go1.20.2 Tue Apr 25 18:14:17 UTC 2023
```

antoineco:
Screenshot_20230425-233400

nderjung:
💯

My bad! I can see my issue was resolved from a separate issue which is now resolved locally. :)

nderjung (approved):
Thanks!

PR #427, opened 2023-04-28
fix(cmd): filter out registered flags before parse

antoineco:
### Prerequisite checklist

### Description of changes

Some kraft commands accept flags which registration is delayed using `cmdfactory.RegisterFlag`. Parsing these here results in an expected failure. `kraft pkg --oci-tag` is one example of such flag.

Flags are parsed a second time by `spf13/cobra` inside `cmdfactory.Main`, so if a flag is really not supported, the error will still be reported eventually.

### How it was tested

- `kraft pkg --arch x86_64 --oci-tag test:test` now succeeds
- `kraft pkg --arch x86_64 --oci-tag test:test --foo bar` does not

### DIFF
--- a/cmdfactory/builder.go
+++ b/cmdfactory/builder.go
@@ -96,6 +96,80 @@ func expandRegisteredFlags(cmd *cobra.Command) {
 	}
 }
 
+// filter out registered flags from the given command's args.
+func filterRegisteredFlags(cmd *cobra.Command, args []string) (filteredArgs []string) {
+	for cmdline, flags := range flagOverrides {
+		if !isSameCommand(cmd, cmdline) {
+			continue
+		}
+
+		registeredFlagsNames := make(map[string]struct{}, len(flags))
+		for _, flag := range flags {
+			registeredFlagsNames[flag.Name] = struct{}{}
+		}
+
+		for len(args) > 0 {
+			arg := args[0]
+			args = args[1:]
+
+			switch {
+			// not a flag
+			case arg[0] != '-' || len(arg) == 1:
+				filteredArgs = append(filteredArgs, arg)
+
+			// long flag
+			case arg[1] == '-' && len(arg) > 2:
+				subs := strings.SplitN(arg, "=", 2)
+
+				flagName := strings.TrimPrefix(subs[0], "--")
+				if _, isRegistered := registeredFlagsNames[flagName]; isRegistered {
+					if len(subs) == 1 {
+						args = args[1:]
+					}
+					continue
+				}
+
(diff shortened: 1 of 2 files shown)


nderjung:
Is this ready for review?

antoineco:
@nderjung I just added tests. Go ahead.

.github/workflows/gochecks.yaml
```diff
@@ -84,3 +84,8 @@ jobs:
               git status --short
               exit 1
             fi
+
+      - name: Run tests
+        run: |
+          go run github.com/onsi/ginkgo/v2/ginkgo -v -p -randomize-all \
+            ./cmdfactory/...
```
nderjung: Should this be part of the same commit/PR? Also should this not be part of a test pipeline instead of the linter pipeline? See https://github.com/unikraft/kraftkit/issues/432

nderjung: My confusion results from the pipeline being named `golangci-lint`. This can be resolved just with renaming it to `gochecks` (like the filename).

nderjung: ~~For now, I'd remove this line (though it proves helpful for demonstrating how to test the code) and a separate PR can be opened to add this step "Run tests" and rename the pipeline.~~

antoineco: Done, see last commit.

nderjung: Ignore my last comment! :) Your force push was faster!

antoineco: I can also merge into "gochecks" and have 2 workflows inside. I'd prefer that a little actually :)

nderjung: I'd also move e2e into gotests as a parallel job along with units. Also, is this the right PR for this?

nderjung: Thoughts on merging go checks + go tests (unit + e2e) into a single pipeline? Or keep gochecks and gotests?

antoineco: You tell me, but I think it's critical that we at least run the tests included in this PR.

nderjung: I'd then do two pipelines, gochecks + gotests(unit+e2e), as a separate PR which is merged first and this PR is rebased.

nderjung: https://github.com/unikraft/kraftkit/issues/432

antoineco: Done #433
Will rebase after the merge.

nderjung (approved):
Awesome! Looks good, thanks!

PR #428, opened 2023-04-28
fix(oci): Use correct name referencing when tagging

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/oci/pack.go
+++ b/oci/pack.go
@@ -232,7 +232,7 @@ func NewPackageFromTarget(ctx context.Context, targ target.Target, opts ...packm
 		"tag": ocipack.Name(),
 	}).Debug("oci: saving image")
 
-	_, err = image.Save(ctx, ocipack.Name(), nil)
+	_, err = image.Save(ctx, ocipack.imageRef(), nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


oci/pack.go
```diff
@@ -38,7 +38,7 @@ type ociPackage struct {
 	image  *Image
 
 	// Embedded attributes which represent target.Target
-	name    string
+	// name    string
```
antoineco: If a name is no longer needed, let's remove the field.

oci/pack.go
```diff
@@ -389,7 +387,7 @@ func (ocipack *ociPackage) Pull(ctx context.Context, opts ...pack.PullOption) er
 		return err
 	}
 
-	fullref := ocipack.Name()
+	fullref := fmt.Sprintf("%s:%s", ocipack.Name(), ocipack.Version())
```
antoineco: nit: I see that this format string is used in two places. We might as well create a very small function to encapsulate this format to ensure that all  functions use the same.

antoineco (approved):
Approving with a non-blocking comment.

oci/pack.go
nderjung: I looked internally into [the `name` package](github.com/google/go-containerregistry/pkg/name) to see if we could re-use anything there but there was nothing from a first glance.  I've just done a fixup and will push the change now.

PR #438, opened 2023-04-29
fix(cmd): properly check for help flag in args

### LINKED ISSUE #430: When the `--help` flag is passed, direct sub-commands of `kraft` print "unknown command" instead of their help
### Describe the bug

Requesting a direct sub-command's help while top-level flags are present in the arguments causes that command to print the message "unknown command" followed by a _partial_ help.

Example:

```console
$ kraft --log-type=basic version --help
unknown command "version" for "kraft version"

Usage:  kraft version

Flags:
  -h, --help   help for version
```

It doesn't occur with the top-level `kraft` command, nor with nested sub-commands. For instance, it occurs with

```console
$ kraft --log-type=basic pkg --help
                        ^
                        direct subcommand
```

but not with

```console
$ kraft --log-type=basic pkg update --help
                             ^
                             nested subcommand
```

### Steps to reproduce

_No response_

### Expected behavior

The command's defined help is printed:

```console
$ kraft --log-type=basic version --help
Show kraft version information

USAGE
  kraft version

FLAGS
  -h, --help   help for version

INHERITED FLAGS
  --config-dir string        Path to KraftKit config directory
  --containerd-addr string   Address of containerd daemon socket
  --editor string            Set the text editor to open when prompt to edit a file
  --events-pid-file string   Events process ID used when running multiple unikernels (default "/var/kraftkit/events.pid")
  --git-protocol string      Preferred Git protocol to use (default "https")
  --http-unix-sock string    When making HTTP(S) connections, pipe requests via this shared socket
  --log-level string         Log level verbosity (default "info")
  --log-timestamps           Enable log timestamps
  --log-type string          Log type (default "fancy")
  --manifests-dir string     Path to Unikraft manifest cache
  --no-check-updates         Do not check for updates
  --no-emojis                Do not use emojis in any console output
  --no-parallel              Do not run internal tasks in parallel
  --no-prompt                Do not prompt for user interaction
  --pager string             System pager to pipe output to
  --plugins-dir string       Path to KraftKit plugin directory
  --runtime-dir string       Directory for placing runtime files (e.g. pidfiles) (default "/var/kraftkit")
  --sources-dir string       Path to Unikraft component cache
  --with-manifest strings    Paths to package or component manifests
  --with-mirror strings      Paths to mirrors of Unikraft component artifacts
```

_No response_

_No response_

### Relevant log output

_No response_

antoineco:
### Prerequisite checklist

### Description of changes

Fixes #430

Checks whether the arguments contain a help flag, regardless of this position in the arguments list.

This officially concludes the list of things I wanted to tackle as a follow-up to the initial e2e tests 🙂

### DIFF
--- a/cmdfactory/help.go
+++ b/cmdfactory/help.go
@@ -14,6 +14,7 @@ import (
 	"kraftkit.sh/iostreams"
 	"kraftkit.sh/log"
 
+	"kraftkit.sh/internal/set"
 	"kraftkit.sh/internal/text"
 )
 
@@ -118,7 +119,7 @@ func fullname(cmd *cobra.Command) string {
 }
 
 func rootHelpFunc(cmd *cobra.Command, args []string) {
-	if isRootCmd(cmd.Parent()) && len(args) >= 2 && args[1] != "--help" && args[1] != "-h" {
+	if isRootCmd(cmd.Parent()) && len(args) >= 2 && !set.NewStringSet(args...).ContainsAnyOf("--help", "-h") {
 		nestedSuggestFunc(cmd, args[1])
 		hasFailed = true
 		return
--- a/test/e2e/cli/version_test.go
+++ b/test/e2e/cli/version_test.go
@@ -43,10 +43,7 @@ var _ = Describe("kraft version", func() {
 
 	When("invoked with the --help flag", func() {
 		BeforeEach(func() {
-			// FIXME(antoineco): Sub-commands of the root command handle help flags
-			// improperly when top-level flags exist in the arguments.
-			// Ref. unikraft/kraftkit#430
-			cmd.Args = []string{cmd.Args[0], cmd.Args[len(cmd.Args)-1], "--help"}
+			cmd.Args = append(cmd.Args, "--help")
 		})
 
 		It("should print the command's help", func() {


cmdfactory/help.go
```diff
@@ -268,3 +269,20 @@ func dedent(s string) string {
...
+	}
+
+	for _, input := range inputs {
+		if elemsSet.Contains(input) {
+			return true
+		}
+	}
+
+	return false
+}
```
nderjung: This utility function can be repurposed and imho should be placed within the `internal/set` package as `ContainsOneOf` and part of `stringSet`.

The [`NewStringSet()`](https://github.com/unikraft/kraftkit/blob/staging/internal/set/string_set.go#L33-L38) method's prototype should also be updated to `NewStringSet(base ...string)` for faster use of `AddValues`.

nderjung: Thus:
```diff
-	if isRootCmd(cmd.Parent()) && len(args) >= 2 && !containsOneOf(args, "--help", "-h") {
+	if isRootCmd(cmd.Parent()) && len(args) >= 2 && !set.NewStringSet(args...).ContainsOneOf("--help", "-h") {
```

antoineco: I was thinking exactly the same, the current set implementation is lacking a proper initializer.
Pushing as a separate PR.

nderjung: If you are creating a separate PR, please can you include a replacement (and deprecation of) the more primitive [`utils.Contains`](https://github.com/unikraft/kraftkit/blob/staging/utils/utils.go#L117-L125) method which is used more widely in the project. There is currently little-to-no use of `stringSet` which is more specific and has its own package. The `utils` package is anyway a bit of a odd package and left over from a some verbatim imports a long time ago. Using the `set` package is preferable.

antoineco: String set changes are in #439
~I'm addressing the change to the utils package now in its own PR.~
I just noticed `utils.Contains` could be modified to use the set, sorry if I didn't get that at first. Doing it now in the same PR.

antoineco:
Waiting for #439

nderjung (approved):
Looks good, thanks!

PR #443, opened 2023-04-30
ci: run workflows on merge to the default branch

### LINKED ISSUE #442: Improve caching in pull requests by running workflows on the default branch
### Proposal

Our current caching strategy is of limited use because the first run of any workflow in a PR is performed without a cache (_cache miss_). The reason for that is...

> Workflow runs can restore caches created in either the current branch or the default branch
> -- https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

If we were running our workflows on the default branch (`staging`), the cache would be warmed up for all subsequent PRs, which could be a significant time and space saver.

antoineco:
### Prerequisite checklist

### Description of changes

Aside from providing a better guarantee that merges do not break the default branch, running workflows on it makes caches [readily available to other branches](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache), which prevents creating one cache per branch and speeds up workflow runs significantly.

Closes #442

### DIFF
--- a/.github/workflows/gobuild-protoc.yaml
+++ b/.github/workflows/gobuild-protoc.yaml
@@ -1,6 +1,8 @@
 name: tools/protoc-gen-go-netconn
 
 on:
+  push:
+    branches: [staging]
   pull_request:
     types: [opened, synchronize, reopened]
     branches: [staging]
--- a/.github/workflows/gobuild-qemu-devices.yaml
+++ b/.github/workflows/gobuild-qemu-devices.yaml
@@ -1,6 +1,8 @@
 name: tools/go-generate-qemu-device
 
 on:
+  push:
+    branches: [staging]
   pull_request:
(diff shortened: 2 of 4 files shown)


antoineco:
Requires #450

antoineco:
@nderjung the commit check action is complaining about the long URL in the commit message. We can ignore it right?

nderjung:
We can ignore it for now. I think the eventual solve is to prevent checking lines that start with `\[\d+\]:\shttp`.

nderjung (approved):
Makes sense. We didn't run it because we didn't want to re-build every merge into `staging`. However, now with a more sophisticated caching system (thanks!) this addition makes more sense.

PR #458, opened 2023-05-05
build: upgrade buildenv base to Ubuntu 22.04

### LINKED ISSUE #457: Release binaries are dynamically linked against the outdated libssl1.1
### Description

`make kraft` inside the devenv container produces an executable that is dynamically linked against `libssl1.1`.
By extension this includes **release binaries** (tested with v0.5.1).

This is a problem because the last Ubuntu release that included this version was 20.04 LTS:
- https://packages.ubuntu.com/search?keywords=libssl1.1
- https://packages.ubuntu.com/search?suite=focal&arch=any&searchon=contents&keywords=libssl.so.1.1

```console
$ ldd dist/kraft
        linux-vdso.so.1 (0x00007fffb89b5000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f0128905000)
        libssh2.so.1 => /lib/x86_64-linux-gnu/libssh2.so.1 (0x00007f01288c1000)
        libssl.so.1.1 => not found
        libcrypto.so.1.1 => not found
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f01288bc000)
        libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f01288a7000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f01286a2000)
        libcrypto.so.3 => /lib/x86_64-linux-gnu/libcrypto.so.3 (0x00007f012824f000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f0128231000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f0128914000)
```

`kraft` being a user-oriented program, we must expect people to be running it on modern operating systems.

antoineco:
### Prerequisite checklist

### Description of changes

Allows producing release binaries which are dynamically linked against libssl3 instead of libssl1.1. The latter is no longer available in recent Linux distributions.

Fixes #457

### DIFF
--- a/buildenvs/myself.Dockerfile
+++ b/buildenvs/myself.Dockerfile
@@ -5,19 +5,31 @@
 
 ARG GO_VERSION=1.20.2
 
-FROM golang:${GO_VERSION}-bullseye AS kraftkit-full
+FROM golang:${GO_VERSION}-bullseye AS golang
+
+FROM ubuntu:22.04 AS kraftkit-full
+
+# https://github.com/docker-library/golang/blob/5c6fa890/1.20/bullseye/Dockerfile
+ENV PATH /usr/local/go/bin:$PATH
+ENV GOLANG_VERSION ${GO_VERSION}
+COPY --from=golang /usr/local/go /usr/local/go
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:$PATH
+RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 1777 "$GOPATH"
 
 # Install build dependencies
 RUN set -xe; \
     apt-get update; \
     apt-get install -y --no-install-recommends \
-      build-essential=12.9 \
-      cmake=3.18.4-2+deb11u1 \
-      libssh2-1-dev=1.9.0-2 \
-      libssl-dev=1.1.1n-0+deb11u4 \
-      make=4.3-4.1 \
-      pkg-config=0.29.2-1 \
-      git=1:2.30.2-1+deb11u2; \
+      ca-certificates=20211016ubuntu0.22.04.1 \
+      curl=7.81.0-1ubuntu1.10 \
+      build-essential=12.9ubuntu3 \
+      cmake=3.22.1-1ubuntu1.22.04.1 \
+      libssh2-1-dev=1.10.0-3 \
+      libssl-dev=3.0.2-0ubuntu1.9 \
+      pkg-config=0.29.2-1ubuntu3 \
+      openssh-client=1:8.9p1-3ubuntu0.1 \
+      git=1:2.34.1-1ubuntu1.9; \
     apt-get clean; \
     go install mvdan.cc/gofumpt@v0.4.0; \
     git config --global --add safe.directory /go/src/kraftkit.sh;


nderjung (changes requested):
Thanks for this.

I tried building the container image via `make buildenv-myself-full` then using that build environment to build kraft. There are a lot of warnings when compiling libgit2 now but more importantly:

```
go: cannot find GOROOT directory: /usr/local/go
make: *** [Makefile:150: git2go] Error 2
```

nderjung:
You are too fast 🔥 ignore my last review, trying again with your latest force push.

antoineco:
I realized at the same time as you did that I made that mistake :)

antoineco:
Please note that, with these changes, `kraft` now works on **Ubuntu 22.04 or newer** only.

On Ubuntu 20.04 (jammy) and Debian 11 (bullseye):

```
kraft: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by distkraft/kraft)
kraft: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by distkraft/kraft)
kraft: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by distkraft/kraft)
        linux-vdso.so.1 (0x00007ffe64b88000)
        libssh2.so.1 => not found
        libssl.so.3 => not found
        libcrypto.so.3 => not found
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f969d2ec000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f969d4e2000)
```

nderjung:
For legacy reasons, I'd like to keep Debian 10 still supported.

antoineco:
Then we need to support multiple OS versions, because I can't imagine releasing something that only works _up to_ Debian 11 and Ubuntu 20.04 in 2023.

nderjung:
Free thinking a bit. Before we spend more time implementing this, let's step back. Ultimately, libgit2, libssl, libssh arise from the transport properties found in this file: https://github.com/unikraft/kraftkit/blob/staging/manifest/pack_pull_git.go

The question is whether if it's more work to replace that functionality with https://github.com/go-git/go-git/ which (at a quick glance) looks like it does not have these dependencies VS. having multiple supported OSes.

One argument in favour of supporting multiple OSes whether now or eventually is the idea of providing "base" environments to a user such that they get to use kraft in a familiar environment, e.g.:

```
docker run -it --rm -v $(pwd):/workspace -w /workspace kraftkit.sh/ubuntu:18.04
docker run -it --rm -v $(pwd):/workspace -w /workspace kraftkit.sh/ubuntu:20.04
docker run -it --rm -v $(pwd):/workspace -w /workspace kraftkit.sh/ubuntu:22.04
docker run -it --rm -v $(pwd):/workspace -w /workspace kraftkit.sh/debian:10
// etc
```
Furthermore, they could even use this in the context of a `Dockerfile`'s `FROM` to perform the build and as a drop-in-replacement to get access to a `kraft` binary. However, is it also more work to include these SO dependents...

antoineco:
I'd 100% prefer to avoid these C dependencies than having to create multiple Linux binary releases.

However, developers do use recent OSes (latest LTS for sure), and right now our release artifacts are not usable on these, which is a poor experience.

nderjung:
> I'd 100% prefer to avoid these C dependencies than having to create multiple Linux binary releases.
> 
> However, developers do use recent OSes (latest LTS for sure), and right now our release artifacts are not usable on these, which is a poor experience.

Sounds like even with this idea it is more work to support C/SO dependencies. Looks like we need to re-open https://github.com/unikraft/kraftkit/issues/179

antoineco:
Closing because too controversial. The solution is to remove the problematic dependency on libgit2.

PR #466, opened 2023-05-11
feat(git2go)!: Replace `libgit2` dependency with `go-git`

### LINKED ISSUE #179: Remove libgit2 dependency

### LINKED ISSUE #457: Release binaries are dynamically linked against the outdated libssl1.1

craciunoiuc:
### Prerequisite checklist

### Description of changes

This PR attempts to remove the `libgit2/git2go` dependency that caused so many protability problems and replace it with a native go library named `go-git/go-git`.
The deciding factor was the inability to easily clone git submodules that are part of a repository.

The new library is not perfect. Currently it lacks printing of progress for the better part of the clone process, which is the download itself and unpacking/checking out.
In order to circumvent this, the last ~80% of the clone process is faked with an exponential function to keep the timer/spinner spinning. The progress bar will block at 98%, as going up to 100% would stop the process. This concerns only `--log-type fancy`.

Once this feature is offered by `go-git`, we will have to go back to the function, as represented by the `FIXME` comment.

To sum it up:

### Advantages
 * Native git support - No more need for CGO
 * 2 MB smaller binary
 * Dependencies handled by Go, and up to date - No need for mimicking kraftkit build system in Related Repositories
 * Overall faster cloning
 * Easier to use with simpler interfaces

### Disadvantages
 * No real progress bar in `fancy` mode

Closes: #179
Closes: #457


manifest/pack_pull_git.go
```diff
@@ -6,18 +6,119 @@ package manifest
...
+	// Counting objects step
+	countingObjects = iota
+	// Compressing objects step
+	compressingObjects
+	// Downloading objects step
+	downloadingObjects
+)
+
+func (p cloneProgress) Write(b []byte) (int, error) {
+	if p.onProgress != nil {
```
nderjung:
```diff
- if p.onProgress != nil {
+ if p.onProgress == nil {
+ 	return len(b), nil
+ }
```
Then you can remove an indentation of subsequent lines.

.goreleaser-stable.yaml
```diff
@@ -48,6 +48,4 @@ builds:
       - -X {{ .Env.GOMOD }}/internal/version.buildTime={{ .Date }}
     tags:
       - static
```
antoineco: Can be removed

.goreleaser-staging.yaml
```diff
@@ -48,6 +48,4 @@ builds:
       - -X {{ .Env.GOMOD }}/internal/version.buildTime={{ .Date }}
     tags:
       - static
```
antoineco: Can be removed

Makefile
```diff
@@ -85,7 +85,7 @@ endif
 $(addprefix $(.PROXY), $(BIN)): GO_LDFLAGS += -X "$(GOMOD)/internal/version.version=$(VERSION)"
 $(addprefix $(.PROXY), $(BIN)): GO_LDFLAGS += -X "$(GOMOD)/internal/version.commit=$(GIT_SHA)"
 $(addprefix $(.PROXY), $(BIN)): GO_LDFLAGS += -X "$(GOMOD)/internal/version.buildTime=$(shell date)"
-$(addprefix $(.PROXY), $(BIN)): git2go tidy
+$(addprefix $(.PROXY), $(BIN)): tidy
 $(addprefix $(.PROXY), $(BIN)):
 	$(GO) build \
 		-tags static \
```
antoineco: Can be removed

craciunoiuc: already is 🤷

antoineco: I can still see `-tags static` at the latest commit.

craciunoiuc: Ouch, sorry, I was looking at the modified lines 💫 
image

craciunoiuc: done

antoineco:
@craciunoiuc related to my previous comments, there is also this leftover in the golangci-lint config:
https://github.com/unikraft/kraftkit/blob/a8ba25d9432931c649fb283b9fbc1942b04f3383/.golangci.yaml#L11-L12

Edit:

There is also one here, but it looked like a mistake to start with because neither this program nor its dependencies use build flags:

https://github.com/unikraft/kraftkit/blob/a8ba25d9432931c649fb283b9fbc1942b04f3383/tools/webinstall/Dockerfile#L22-L23

Makefile
```diff
@@ -128,9 +126,6 @@ test: test-unit test-e2e ## Run all tests.
 .PHONY: test-unit
 test-unit: GOTEST_EXCLUDE := third_party/ test/ hack/ buildenvs/ dist/ docs/
 test-unit: GOTEST_PKGS := $(foreach pkg,$(filter-out $(GOTEST_EXCLUDE),$(wildcard */)),$(pkg)...)
-test-unit: git2go ## Run unit tests.
-	$(GO) run github.com/onsi/ginkgo/v2/ginkgo -v -p -randomize-all --tags static $(GOTEST_PKGS)
-
```
antoineco, line 133: Oops. You meant
```diff
-test-unit: git2go ## Run unit tests.
+test-unit: ## Run unit tests.
 	$(GO) run github.com/onsi/ginkgo/v2/ginkgo -v -p -randomize-all --tags static $(GOTEST_PKGS)
```
not
```diff
-test-unit: git2go ## Run unit tests.
-	$(GO) run github.com/onsi/ginkgo/v2/ginkgo -v -p -randomize-all --tags static $(GOTEST_PKGS)
```

go.sum
```diff
@@ -21,11 +21,10 @@ github.com/Microsoft/hcsshim v0.10.0-rc.7/go.mod h1:ILuwjA+kNW+MrN/w5un7n3mTqkws
...
 github.com/OneOfOne/xxhash v1.2.8 h1:31czK/TI9sNkxIKfaUfGlU47BAxQ0ztGgd9vPyqimf8=
-github.com/ProtonMail/go-crypto v0.0.0-20221026131551-cf6655e29de4/go.mod h1:UBYPn8k0D56RtnR8RFQMjmh4KrZzWJ5o7Z9SYjossQ8=
-github.com/ProtonMail/go-crypto v0.0.0-20230214155104-81033d7f4442 h1:OUJ54Fkd+AQXYmr9eOUxZfWNzpK3/e/KD40qa2rKHS4=
-github.com/ProtonMail/go-crypto v0.0.0-20230214155104-81033d7f4442/go.mod h1:I0gYDMZ6Z5GRU7l58bNFSkPTFN6Yl12dsUlAZ8xy98g=
-github.com/acomagu/bufpipe v1.0.3 h1:fxAGrHZTgQ9w5QqVItgzwj235/uYZYgbXitB+dLupOk=
-github.com/acomagu/bufpipe v1.0.3/go.mod h1:mxdxdup/WdsKVreO5GpW4+M/1CE2sMG4jeGJ2sYmHc4=
+github.com/ProtonMail/go-crypto v0.0.0-20230217124315-7d5c6f04bbb8 h1:wPbRQzjjwFc0ih8puEVAOFGELsn1zoIIYdxvML7mDxA=
+github.com/ProtonMail/go-crypto v0.0.0-20230217124315-7d5c6f04bbb8/go.mod h1:I0gYDMZ6Z5GRU7l58bNFSkPTFN6Yl12dsUlAZ8xy98g=
+github.com/acomagu/bufpipe v1.0.4 h1:e3H4WUzM3npvo5uv95QuJM3cQspFNtFBzvJ2oNjKIDQ=
+github.com/acomagu/bufpipe v1.0.4/go.mod h1:mxdxdup/WdsKVreO5GpW4+M/1CE2sMG4jeGJ2sYmHc4=
```
antoineco, line 28: This doesn't look right, the go.sum should only lose things, not replace/downgrade.

craciunoiuc: I think that some other things got merged and my staging is no longer up to date, will check now, thanks!

antoineco: (Maybe the key is to leave the upgrade of `go-git/v5` to dependabot?)

manifest/pack_pull_git.go
```diff
@@ -6,18 +6,99 @@ package manifest
...
+	var scale float64
+
+	// Trim prefix
+	if strings.HasPrefix(string(b), "Counting objects") {
+		prog = strings.TrimPrefix(string(b), "Counting objects: ")
+		scale = 0.1
+	} else if strings.HasPrefix(string(b), "Compressing objects") {
+		prog = strings.TrimPrefix(string(b), "Compressing objects: ")
+		scale = 0.2
+	} else if strings.HasPrefix(string(b), "Total ") {
```
antoineco: nit(style): a good candidate for a `switch`

```suggestion
	switch {
	case strings.HasPrefix(string(b), "Counting objects"):
		prog = strings.TrimPrefix(string(b), "Counting objects: ")
		scale = 0.1
	case strings.HasPrefix(string(b), "Compressing objects"):
		prog = strings.TrimPrefix(string(b), "Compressing objects: ")
		scale = 0.2
	case strings.HasPrefix(string(b), "Total "):
```

manifest/pack_pull_git.go
```diff
@@ -6,18 +6,99 @@ package manifest
...
+	githttp "github.com/go-git/go-git/v5/plumbing/transport/http"
+	gitssh "github.com/go-git/go-git/v5/plumbing/transport/ssh"
 
 	"kraftkit.sh/log"
 	"kraftkit.sh/pack"
 	"kraftkit.sh/unikraft"
 )
 
+type cloneProgress struct {
+	io.Writer
```
antoineco: This seems unused, why is it embedded?

manifest/pack_pull_git.go
```diff
@@ -36,72 +117,62 @@ func pullGit(ctx context.Context, manifest *Manifest, opts ...pack.PullOption) e
...
-
-		copts.FetchOptions.RemoteCallbacks.CredentialsCallback = func(_, usernameFromUrl string, allowedTypes libgit.CredentialType) (*libgit.Credential, error) {
-			if allowedTypes == libgit.CredentialTypeUsername && len(usernameFromUrl) == 0 {
-				return nil, fmt.Errorf("username required in SSH URL")
-			}
-
-			return libgit.NewCredentialSSHKeyFromAgent(usernameFromUrl)
+		copts.Auth, err = gitssh.NewSSHAgentAuth("git")
+		if err != nil {
+			return fmt.Errorf("could not create SSH agent auth: %v", err)
```
antoineco: By wrapping errors we can later discriminate them with `errors.Is`, `errors.Has` and `errors.Unwrap`. In some cases this can be very beneficial, so it's a good practive to always use `%w` with `Errorf`.

```suggestion
			return fmt.Errorf("could not create SSH agent auth: %w", err)
```

manifest/pack_pull_git.go
```diff
@@ -36,72 +117,62 @@ func pullGit(ctx context.Context, manifest *Manifest, opts ...pack.PullOption) e
...
-
-			if auth, ok := manifest.Auths()[u.Host]; ok {
-				if len(auth.User) > 0 && len(auth.Token) > 0 {
-					return libgit.NewCredentialUserpassPlaintext(auth.User, auth.Token)
+		if !strings.HasPrefix(path, "https://") {
+			path = "https://" + path
+		}
+		u, err := url.Parse(path)
+		if err != nil {
+			return fmt.Errorf("could not parse URL: %v", err)
```
antoineco: For the same reason as my previous comment.

```suggestion
			return fmt.Errorf("could not parse URL: %w", err)
```

manifest/pack_pull_git.go
```diff
@@ -6,18 +6,99 @@ package manifest
...
 	"kraftkit.sh/unikraft"
 )
 
+type cloneProgress struct {
+	io.Writer
+
+	once           sync.Once
+	onProgress     func(progress float64)
+	completeParent chan bool
+	completeWorker chan bool
```
antoineco: The boolean value isn't used, so we may as well use zero values (an empty struct is usually considered idiomatic for that).

```suggestion
	completeParent chan struct{}
	completeWorker chan struct{}
```

manifest/pack_pull_git.go
craciunoiuc: The writer is needed by the Progress function. `go-git` passes data to us with the writer

craciunoiuc: works without it, nvm 😅

Makefile
```diff
@@ -128,9 +126,9 @@ test: test-unit test-e2e ## Run all tests.
 .PHONY: test-unit
 test-unit: GOTEST_EXCLUDE := third_party/ test/ hack/ buildenvs/ dist/ docs/
 test-unit: GOTEST_PKGS := $(foreach pkg,$(filter-out $(GOTEST_EXCLUDE),$(wildcard */)),$(pkg)...)
-test-unit: git2go ## Run unit tests.
-	$(GO) run github.com/onsi/ginkgo/v2/ginkgo -v -p -randomize-all --tags static $(GOTEST_PKGS)
 
+test-unit: ## Run unit tests.
+	$(GO) run github.com/onsi/ginkgo/v2/ginkgo -v -p -randomize-all $(GOTEST_PKGS)
```
antoineco: This new line here doesn't look intentional.

antoineco (commented):
Apart from my last comment, and the dependencies versions which I think we should avoid touching in the PR, the code looks fine after a cursory look (not tested myself!).

Makefile
craciunoiuc: well you asked me to add it back :D

go.sum
craciunoiuc: It replaces/upgrades most of them. There is just 1 downgrade. I think this happens because we are replacing it with a very similar library, and they are pinned at different versions

Makefile
antoineco: I'm talking about the new line above. It isn't there on `main`:

https://github.com/unikraft/kraftkit/blob/2609f3a0b160a555525fc5d76f88fa8cec7afa6e/Makefile#L128-L132

craciunoiuc: whoops 😊

antoineco (approved):

nderjung:
Hi. I tried this out locally, this works for me. Good to go after the `go.{mod,sum}` are rebased. Thanks!

nderjung (approved):
😭 Thank you

Issue #34, opened 2022-07-23
Add error tracing in verbose mode


nderjung:
This issue tracks the addition of https://github.com/juju/errors such that errors can be better understood.  Based on the juju errors' README, we need to use the juju `errors` package to replace the internal package's definition of `errors` with juju's, i.e.:
```diff
index 8206a12..4d88304 100644
--- a/internal/errs/defs.go
+++ b/internal/errs/defs.go
@@ -4,7 +4,9 @@
 // You may not use this file expect in compliance with the License.
 package errs

-import "errors"
+import
+       "github.com/juju/errors"
+)
```

**However,** we actually do NOT need to replace the above, in fact, since `internal/errs` declares the same (subset) of `errors`, this can be removed since they will be duplicate. Only when we add new error types (which is possible), would we need to restore this package with custom definitions. Let's hold off for now and simply remove this package.

Instead, to continue this work, I would recommend going through the KraftKit source code and replacing all instances of `return fmt.Errorf` and `errors.New` with valid juju `errors` wrappers. This will allow us to a). classify each error into a category and b). allow us to unwrap each error nicely since the system will homogenise. The moment where errors are actually printed to the console occurs in [`cmdfactory.Main`](https://github.com/unikraft/kraftkit/blob/staging/cmdfactory/builder.go#L181):

https://github.com/unikraft/kraftkit/blob/06bef518ea77d6ff8d80db3c25bf1064d3e4d7ee/cmdfactory/builder.go#L175-L184

For the first iteration, this needs to be replace with juju errors' [`Unwrap`](https://pkg.go.dev/github.com/juju/errors#Unwrap) or [`Trace`](https://pkg.go.dev/github.com/juju/errors#Trace) method which will output the call stack or the list of errors which have occurred.

Finally, the code related to "user" look up can be removed entirely since this is out-of-scope of the project.

b1shtream:
Hello @nderjung 
If someone isn't working on this issue then I would like to play around it and see if I can solve it, could you please assign me this issue ?

craciunoiuc:
Still want to do this @chococandy63? :sweat_smile:

arvganesh:
Hi, I'm a student at the University of Texas at Austin interested in working on this issue as it seems to be available. I'm planning to take on this issue as a part of a project for my virtualization class (CS 360V). Do you have a rough idea of how much effort this would be? @craciunoiuc

craciunoiuc:
[...]

PR #479, opened 2023-05-16
feat: add error tracing

zyllee:
### Prerequisite checklist

### Description of changes

To add the feature related to #34.
The [juju](https://github.com/juju/errors) package for error handling is currently imported. Wrote the test code.

### DIFF
--- a/archive/archive.go
+++ b/archive/archive.go
@@ -8,7 +8,6 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"context"
-	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -16,6 +15,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/juju/errors"
 	"github.com/sirupsen/logrus"
 
 	"kraftkit.sh/log"
@@ -38,12 +38,12 @@ func TarFileWriter(ctx context.Context, src, dst string, tw *tar.Writer, opts ..
 	dst = filepath.ToSlash(dst)
 
 	if dst == "" {
-		return fmt.Errorf("cannot tar file with no specified destination")
+		return errors.New("cannot tar file with no specified destination")
 	} else if dst[0] == '/' {
 		dst = dst[1:]
 	}
 	if strings.HasSuffix(dst, "/") {
-		return fmt.Errorf("attempting to use TarFileWriter with directory")
+		return errors.New("attempting to use TarFileWriter with directory")
 	}
 
 	aopts := ArchiveOptions{}
@@ -55,7 +55,7 @@ func TarFileWriter(ctx context.Context, src, dst string, tw *tar.Writer, opts ..
 
 	fi, err := os.Stat(src)
 	if err != nil {
-		return fmt.Errorf("fail to stat %s: %v", src, err)
+		return errors.Annotatef(err, "fail to stat %s", src)
 	}
 
 	var link string
@@ -68,7 +68,7 @@ func TarFileWriter(ctx context.Context, src, dst string, tw *tar.Writer, opts ..
 
 	header, err := tar.FileInfoHeader(fi, link)
 	if err != nil {
-		return fmt.Errorf("%s: %w", src, err)
+		return errors.Annotatef(err, "%s", src)
 	}
 
 	header.Name = dst
@@ -85,7 +85,7 @@ func TarFileWriter(ctx context.Context, src, dst string, tw *tar.Writer, opts ..
 	}
 
 	if err := tw.WriteHeader(header); err != nil {
-		return fmt.Errorf("tar: %w", err)
+		return errors.Annotate(err, "tar")
 	}
 
 	if mode.IsRegular() {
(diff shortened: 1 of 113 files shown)


nderjung (changes requested):
Hi @zyllee, thanks for this contribution!

The issue you've addressed is out-of-date in terms of its implementation but not its requirements. We have since added [`internal/errs`](https://github.com/unikraft/kraftkit/tree/staging/internal/errs) to KraftKit which now stores error types (instead of `iota`/`int`s, they're `string`s).

As discussed on Discord, the need for a error tracing system is still required and [juju's errors module](https://github.com/juju/errors) still seems like a good fit.

Based on the juju errors' README, you would technically need to use the juju `errors` package to replace the internal package's definition of `errors` with juju's, i.e.:
```diff
index 8206a12..4d88304 100644
--- a/internal/errs/defs.go
+++ b/internal/errs/defs.go
@@ -4,7 +4,9 @@
 // You may not use this file expect in compliance with the License.
 package errs

-import "errors"
+import
+       "github.com/juju/errors"
+)
```

**However,** we actually do NOT need to replace the above, in fact, since `internal/errs` declares the same (subset) of `errors`, this can be removed since they will be duplicate. Only when we add new error types (which is possible), would we need to restore this package with custom definitions. Let's hold off for now and simply remove this package.

Instead, to continue this work, I would recommend going through the KraftKit source code and replacing all instances of `return fmt.Errorf` and `errors.New` with valid juju `errors` wrappers. This will allow us to a). classify each error into a category and b). allow us to unwrap each error nicely since the system will homogenise. The moment where errors are actually printed to the console occurs in [`cmdfactory.Main`](https://github.com/unikraft/kraftkit/blob/staging/cmdfactory/builder.go#L181):

https://github.com/unikraft/kraftkit/blob/06bef518ea77d6ff8d80db3c25bf1064d3e4d7ee/cmdfactory/builder.go#L175-L184

Replace `fmt.Printf(err)` with `errors.ErrorStack(err)`

For the first iteration on your side, this needs to be replace with juju errors' [`Unwrap`](https://pkg.go.dev/github.com/juju/errors#Unwrap) or [`Trace`](https://pkg.go.dev/github.com/juju/errors#Trace) method which will output the call stack or the list of errors which have occurred.



Finally, the code related to "user" look up can be removed entirely since this is out-of-scope of the project.

cmd/kraft/build/build.go
```diff
@@ -91,7 +92,7 @@ func (opts *Build) Run(cmd *cobra.Command, args []string) error {
 	var workdir string
 
 	if (len(opts.Architecture) > 0 || len(opts.Platform) > 0) && len(opts.Target) > 0 {
-		return fmt.Errorf("the `--arch` and `--plat` options are not supported in addition to `--target`")
+		return errors.Errorf("the `--arch` and `--plat` options are not supported in addition to `--target`")
```
zyllee: image-20230525211724259

cmd/kraft/build/build.go
```diff
@@ -116,7 +117,7 @@ func (opts *Build) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	if !app.IsWorkdirInitialized(workdir) {
-		return fmt.Errorf("cannot build uninitialized project! start with: ukbuild init")
+		return errors.Errorf("cannot build uninitialized project! start with: ukbuild init")
```
zyllee: image-20230525214438050

cmd/kraft/build/build.go
```diff
@@ -303,7 +304,7 @@ func (opts *Build) Run(cmd *cobra.Command, args []string) error {
 		}
 
 		if err := treemodel.Start(); err != nil {
-			return fmt.Errorf("could not complete search: %v", err)
+			return errors.Errorf("could not complete search: %v", err)
```
zyllee: image-20230525221207184

cmd/kraft/build/build.go
```diff
@@ -339,7 +340,7 @@ func (opts *Build) Run(cmd *cobra.Command, args []string) error {
 		}
 
 		if err := paramodel.Start(); err != nil {
-			return fmt.Errorf("could not pull all components: %v", err)
+			return errors.Errorf("could not pull all components: %v", err)
```
zyllee: image-20230525220202351

cmd/kraft/build/build.go
```diff
@@ -354,7 +355,7 @@ func (opts *Build) Run(cmd *cobra.Command, args []string) error {
 	)
 
 	if len(selected) == 0 {
-		return fmt.Errorf("no targets selected to build")
+		return errors.Errorf("no targets selected to build")
```
zyllee: image-20230525220353280

zyllee:
The screenshots above(except the first one) are running in cmd `kraft build --arch x86_64 --plat kvm`. I test it by changing `if` condition to satisfied return error to happen.

archive/archive.go
```diff
@@ -85,7 +85,7 @@ func TarFileWriter(ctx context.Context, src, dst string, tw *tar.Writer, opts ..
 	}
 
 	if err := tw.WriteHeader(header); err != nil {
-		return fmt.Errorf("tar: %w", err)
+		return errors.Annotatef(err, "tar:%v", err)
```
zyllee, line 88: juju's error package can't handle %w format. So I try to implement in this way to have same effect.

nderjung:
Closing due to age and it requiring a large rebase.

PR #480, opened 2023-05-17
fix(build): Align pull invocation based on correct package size

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/cmd/kraft/build/build.go
+++ b/cmd/kraft/build/build.go
@@ -46,6 +46,12 @@ type Build struct {
 	Target       string `long:"target" short:"t" usage:"Build a particular known target"`
 }
 
+// The longest word is "configuring" (which is 11 characters long), plus
+// additional space characters (2 characters), brackets (2 characters) the
+// name of the project and the target/plat string (which is variable in
+// length).
+const maxExtraCharacters = 15
+
 func New() *cobra.Command {
 	cmd, err := cmdfactory.New(&Build{}, cobra.Command{
 		Short: "Configure and build Unikraft unikernels",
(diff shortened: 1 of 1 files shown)


cmd/kraft/build/build.go
```diff
@@ -310,6 +310,9 @@ func (opts *Build) Run(cmd *cobra.Command, args []string) error {
 	if len(missingPacks) > 0 {
 		for _, p := range missingPacks {
 			p := p // loop closure
+			if newLen := len(unikraft.TypeNameVersion(p)) + len("pulling "); !norender && newLen > nameWidth {
+				nameWidth = newLen
+			}
```
nderjung: I'd add simply use the integer value to decrease processing time and a comment indicating that it refers to additional length of the string used by the left text name.

Also I believe "`configuring `" is the longest string used.

nderjung: Having said this, shouldn't this be added to [the pre-emptive calculation](https://github.com/unikraft/kraftkit/blob/staging/cmd/kraft/build/build.go#L134-L138)? The issue is probably here?

nderjung: The calculation there is wrong because it does not take into account the version.

craciunoiuc: Not 100% sure why, but in the other cases it works (fetching), I can change there and try out again

It's used only in 2 places anyways

cmd/kraft/build/build.go
```diff
@@ -127,12 +127,17 @@ func (opts *Build) Run(cmd *cobra.Command, args []string) error {
...
+		}
+
 		// The longest word is "configuring" (which is 11 characters long), plus
 		// additional space characters (2 characters), brackets (2 characters) the
 		// name of the project and the target/plat string (which is variable in
 		// length).
-		for _, targ := range project.Targets() {
-			if newLen := len(targ.Name()) + len(target.TargetPlatArchName(targ)) + 15; newLen > nameWidth {
+		for _, component := range comps {
+			if newLen := len(unikraft.TypeNameVersion(component)) + 15; newLen > nameWidth {
```
antoineco: Can we please add a `const` that describes the value `15`?

```go
const maxLengthOfBlah = 15  // or equivalent
```

antoineco: (there is a comment about it above so this is more like a nice to have, not a blocking comment)

craciunoiuc: 🖖

craciunoiuc: done 💪

antoineco (approved):

PR #481, opened 2023-05-17
fix(build): Pull instead if already cloned

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/manifest/pack_pull_git.go
+++ b/manifest/pack_pull_git.go
@@ -6,6 +6,7 @@ package manifest
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


manifest/pack_pull_git.go
```diff
@@ -188,8 +188,25 @@ func pullGit(ctx context.Context, manifest *Manifest, opts ...pack.PullOption) e
 		WithField("branch", popts.Version()).
 		Infof("git clone")
 
-	if _, err = git.PlainCloneContext(ctx, local, false, copts); err != nil {
-		return fmt.Errorf("could not clone repository: %w", err)
+	if _, err := git.PlainCloneContext(ctx, local, false, copts); err != nil {
+		reps, err := git.PlainOpen(local)
```
antoineco: I haven't looked into the lib, but maybe there is a particular error type returned by `PlainClone` that we can switch on, instead of ignoring all errors?

```go
switch {
    case errors.Is(err, git.ErrorRepoExists): // I made that error up
        // open, fetch
    case err != nil:
        // fail
}
```

craciunoiuc: image

clone has only a single error case, as far as I can see

craciunoiuc: I can switch on that, sure, no problem

antoineco: I checked, and there is!

- https://github.com/go-git/go-git/blob/3f1cfde283c93f33218c807602e93d47f72f7b90/repository.go#L416
- https://github.com/go-git/go-git/blob/v5.6.1/repository.go#L52

antoineco: (oops, eventual consistency)

craciunoiuc: updated

manifest/pack_pull_git.go
```diff
@@ -188,7 +188,30 @@ func pullGit(ctx context.Context, manifest *Manifest, opts ...pack.PullOption) e
 		WithField("branch", popts.Version()).
 		Infof("git clone")
 
-	if _, err = git.PlainCloneContext(ctx, local, false, copts); err != nil {
+	_, err = git.PlainCloneContext(ctx, local, false, copts)
+	switch err {
+	case git.ErrRepositoryAlreadyExists:
```
antoineco: This won't work anymore if the returned error is ever wrapped upstream.
Since Go 1.13, the recommended way is to use `errors.Is` to assert a certain error type (or `errors.As` if the underlying dynamic type needs to be accessed).

manifest/pack_pull_git.go
```diff
@@ -188,7 +188,30 @@ func pullGit(ctx context.Context, manifest *Manifest, opts ...pack.PullOption) e
...
+		case git.NoErrAlreadyUpToDate, git.ErrBranchExists, nil:
+			log.G(ctx).Infof("successfully updated %s in %s", path, local)
+			return nil
+		default:
+			return fmt.Errorf("could not clone repository: %w", err)
+		}
+	case nil:
+		break
+	default:
 		return fmt.Errorf("could not clone repository: %w", err)
```
antoineco: We only need one case here:

```suggestion
	case err != nil:
		return fmt.Errorf("could not clone repository: %w", err)
```

If it's `nil`, we "break" out of the switch implicitly.

craciunoiuc: you sure know your switches, updated

antoineco (approved):

PR #489, opened 2023-05-22
feat!: machine api

nderjung:
### Prerequisite checklist

### Description of changes

This PR represents a major refactor to KraftKit's interface for manipulating VM instances.  At its core, a new top-level API package (available at `kraftkit.sh/api`) exposes externally accessible KraftKit internal APIs via the introduction of a new framework, [The Zip API Object Framework](https://api.zip).  The framework allows the definition of APIs in a Kubernetes-like fashion without requiring code generation.  Therefore it is possible to generate traditional CRD-like objects (objects usually defined with a "Spec" and "Status") in a simple way.  The interface is defined simply by introducing a relevant Spec-and-Status structs and using the Zip API generic:

```go
import zip "api.zip"

type (
	// Machine is the mutable API object that represents a machine instance.
	Machine = zip.Object[MachineSpec, MachineStatus]

	// MachineList is the mutable API object that represents a list of machine
	// instances.
	MachineList = zip.ObjectList[MachineSpec, MachineStatus]
)
```

The use of the Zip API Object Framework additionally provides a configurable client which is protocol agnostic, allowing the interface to the object to be defined by the implementer.  One potential client implementation could simply be an inline function call, a remote HTTP request, gRPC call, etc.  Unlike the traditional Kubernetes client, these calls can be inline code function calls.  This allows KraftKit to use a controller-pattern without having to make external calls.  It also paves the way for a future implementation to use the same API with minimal configuration change to invoke the call remotely, if KraftKit is ever daemonized. For example, for the `Create` method of the `MachineService`, we wrap the implementing method with the Zip API client:

```go
create, err := zip.NewMethodClient(ctx, impl.Create, opts...)
if err != nil {
	return nil, err
}
```

The Zip API client is versatile and allows for "before" and "after" wrapping of calls which are defined by the implementer through the [`WithStore` option](https://github.com/api-zip/api-zip/blob/c0ff264a51ca77abf5daa2329019b03c64cf09b9/client_options.go#L47-L145).  This allows us to save the state of API calls through the new embedded store, allowing for content rehydration and local state of the API.  Before, the state was always tailored to the machine. Now, when future APIs are defined, we no longer have to think about its management. This refactor migrates the older store to a new generic embedded store compatible with the Zip API Framework.

This PR introduces the both the definition of the interface for manipulating VMs, or simply "machines" by KraftKit, which essentially defines the configuration used within the context of the current machine driver, as well as necessary implementation methods which manipulate the API in a meaningful way:

```go
// MachineService is the interface of available methods which can be performed
// by an implementing machine platform driver.
type MachineService interface {
	Create(context.Context, *Machine) (*Machine, error)
	Start(context.Context, *Machine) (*Machine, error)
	Pause(context.Context, *Machine) (*Machine, error)
	Stop(context.Context, *Machine) (*Machine, error)
	Update(context.Context, *Machine) (*Machine, error)
	Delete(context.Context, *Machine) (*Machine, error)
	Get(context.Context, *Machine) (*Machine, error)
	List(context.Context, *MachineList) (*MachineList, error)
	Watch(context.Context, *Machine) (chan *Machine, chan error, error)
	Logs(context.Context, *Machine) (chan string, chan error, error)
}
```

This large refactor deprecates the `machine/driver` package in favour of the new top-level API package interface and the implementing interfaces.  To make this switch, the QEMU driver implementation has been adjusted to match the service implementation defined in `api/machine/v1alpha1`.  As a result, all uses of driver in subcommands are adjusted simultaneously to keep the project in working condition. 

Part of this refactor adjusts how the `kraft run` command works due to the nature of the change of the API. Particularly, the split between "kernel arguments" and "application arguments" is made more prevelant through the API.  As such, the args supplied to `kraft run` has changed, which now has an additional `--kernel-args` flag.  Additionally, this refactor introduces a new package, `machine/platform`, aimed at a). naming supported underlying systems using a similar naming convention to Unikraft (already present via `unikraft/plat`) and b). to replace the machine driver model.  Since the goal is to make the choice between the runtime platform highly configurable (and hosts can support different platforms), a mechanism for instantiating and providing choice between platform implementations is provided via "strategy" concept. This strategization of the platform allows for quickly switching between different implementations at a high-level.

Finally, miscellaneous additional fixes are included, including ones which affect `kraft rm` and `kraft stop` such that they are UNIX-like and mimic the output of equivalent commands from `docker`. 

GitHub-Depends: https://github.com/unikraft/kraftkit/pull/503
GitHub-Depends: https://github.com/unikraft/kraftkit/pull/505


machine/platform/detect.go
```diff
@@ -0,0 +1,145 @@
...
+		if i < int(offset) {
+			continue
+		}
+		ret = append(ret, strings.Trim(line, "\n"))
+	}
+
+	return ret, nil
+}
+
+// Detect returns the hypervisor and system mode in the context to the deremined
```
craciunoiuc: `determined`

machine/qemu/qemu.go
```diff
@@ -675,52 +608,51 @@ func (qd *QemuDriver) State(ctx context.Context, mid machine.MachineID) (state m
 	if process, err := processFromPidFile(qcfg.PidFile); err == nil {
 		activeProcess, err = process.IsRunning()
 		if err != nil {
-			state = machine.MachineStateDead
+			fmt.Printf("1111\n")
```
craciunoiuc: `2222`

machine/qemu/qemu.go
```diff
@@ -675,52 +608,51 @@ func (qd *QemuDriver) State(ctx context.Context, mid machine.MachineID) (state m
...
 		}
 
-		// Finally, save the state if it is different from the what we have on record
+		// Finally, save the state if it is different from the what we have on
+		// record
 		if state != savedState {
-			if err = qd.dopts.Store.SaveMachineState(mid, state); err != nil {
-				return
-			}
+			fmt.Printf("seting state\n")
```
craciunoiuc: 👀

craciunoiuc (approved):
All good here. Thanks!

PR #499, opened 2023-06-01
feat: Configurable output format

mdsahiloss:
### Prerequisite checklist

### DIFF
--- a/cmd/kraft/pkg/list/list.go
+++ b/cmd/kraft/pkg/list/list.go
@@ -23,14 +23,15 @@ import (
 )
 
 type List struct {
-	Limit     int  `long:"limit" short:"l" usage:"Set the maximum number of results" default:"50"`
-	NoLimit   bool `long:"no-limit" usage:"Do not limit the number of items to print"`
-	ShowApps  bool `long:"apps" short:"" usage:"Show applications"`
-	ShowArchs bool `long:"archs" short:"M" usage:"Show architectures"`
-	ShowCore  bool `long:"core" short:"C" usage:"Show Unikraft core versions"`
-	ShowLibs  bool `long:"libs" short:"L" usage:"Show libraries"`
-	ShowPlats bool `long:"plats" short:"P" usage:"Show platforms"`
-	Update    bool `long:"update" short:"u" usage:"Get latest information about components before listing results"`
+	Limit     int    `long:"limit" short:"l" usage:"Set the maximum number of results" default:"50"`
+	NoLimit   bool   `long:"no-limit" usage:"Do not limit the number of items to print"`
+	ShowApps  bool   `long:"apps" short:"" usage:"Show applications"`
+	ShowArchs bool   `long:"archs" short:"M" usage:"Show architectures"`
+	ShowCore  bool   `long:"core" short:"C" usage:"Show Unikraft core versions"`
+	ShowLibs  bool   `long:"libs" short:"L" usage:"Show libraries"`
+	ShowPlats bool   `long:"plats" short:"P" usage:"Show platforms"`
+	Update    bool   `long:"update" short:"u" usage:"Get latest information about components before listing results"`
+	Output    string `long:"output" short:"o" usage:"Set output format" default:"table"`
 }
 
 func New() *cobra.Command {
@@ -147,22 +148,28 @@ func (opts *List) Run(cmd *cobra.Command, args []string) error {
 	defer iostreams.G(ctx).StopPager()
 
 	cs := iostreams.G(ctx).ColorScheme()
-	table := tableprinter.NewTablePrinter(ctx)
+	table, err := tableprinter.NewTablePrinter(ctx,
+		tableprinter.WithMaxWidth(iostreams.G(ctx).TerminalWidth()),
+		tableprinter.WithOutputFormatFromString(opts.Output),
+	)
+	if err != nil {
+		return err
+	}
 
 	// Header row
-	table.AddField("TYPE", nil, cs.Bold)
-	table.AddField("PACKAGE", nil, cs.Bold)
-	table.AddField("LATEST", nil, cs.Bold)
-	table.AddField("FORMAT", nil, cs.Bold)
+	table.AddField("TYPE", cs.Bold)
+	table.AddField("PACKAGE", cs.Bold)
+	table.AddField("LATEST", cs.Bold)
+	table.AddField("FORMAT", cs.Bold)
 	table.EndRow()
 
 	for _, pack := range packages {
-		table.AddField(string(pack.Type()), nil, nil)
-		table.AddField(pack.Name(), nil, nil)
-		table.AddField(pack.Version(), nil, nil)
-		table.AddField(pack.Format().String(), nil, nil)
+		table.AddField(string(pack.Type()), nil)
+		table.AddField(pack.Name(), nil)
+		table.AddField(pack.Version(), nil)
+		table.AddField(pack.Format().String(), nil)
 		table.EndRow()
 	}
 
-	return table.Render()
+	return table.Render(iostreams.G(ctx).Out)
 }
--- a/internal/tableprinter/tableprinter.go
+++ b/internal/tableprinter/tableprinter.go
@@ -25,132 +25,98 @@ package tableprinter
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"sort"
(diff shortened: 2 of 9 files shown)


nderjung:
Hi @MdSahil-oss, thanks for opening up the PR! I've converted it to a draft in the mean time as you work on it.

Just an initial note. We already have [`KraftKit.Log.Type`](https://github.com/unikraft/kraftkit/blob/staging/config/config.go#L42) which essentially encompasses the same functionality. Instead, when making the change, use this existing configuration variable to switch the output format. E.g.:
```go
if config.G[config.KraftKit](ctx).Log.Type == "json" {
// ...
}
```

mdsahiloss:
@nderjung Okay! I'll make use of existing `Log.Type` variable and remove `Output` flag but `Log.Type` is set to `fancy` by default, Should I change it to `json` as you mentioned [here](https://github.com/unikraft/kraftkit/issues/4#issuecomment-1527909473).

nderjung:
Hi @MdSahil-oss, no, keep it at its default value. We simply augment internals based on its value. "fancy" is used for CLI users and "json" will be used by machines.

internal/tableprinter/table_printer.go
```diff
@@ -42,22 +44,28 @@ type TablePrinter interface {
 }
 
 type TablePrinterOptions struct {
-	IsTTY bool
+	IsTTY  bool
+	format string
```
nderjung: Switched this to the new type:
```diff
-	format string
+	format TableOutputFormat
```

internal/tableprinter/table_printer.go
```diff
@@ -42,22 +44,28 @@ type TablePrinter interface {
...
 	io := iostreams.G(ctx)
-	if opts.IsTTY {
+	opts := &TablePrinterOptions{}
+	for _, tpo := range topts {
+		tpo(opts)
+	}
+	opts.IsTTY = io.IsStdoutTTY()
+
+	// In case output format is set to JSON or something else except `default` then uses `ttyTablePrinter` by default to print output.
+	if opts.IsTTY || opts.format != string(OutputFormatTable) {
```
nderjung: No need to cast, now that it's the desired type:
```diff
- 	if opts.IsTTY || opts.format != string(OutputFormatTable) {
+ 	if opts.IsTTY || opts.format != OutputFormatTable {
```

internal/tableprinter/table_printer.go
```diff
@@ -42,22 +44,28 @@ type TablePrinter interface {
...
-	io := iostreams.G(ctx)
-	return NewTablePrinterWithOptions(ctx, TablePrinterOptions{
-		IsTTY: io.IsStdoutTTY(),
-	})
+func NewTablePrinter(ctx context.Context, format string) TablePrinter {
+	return NewTablePrinterWithOptions(ctx, WithOutputFormat(format))
 }
 
-func NewTablePrinterWithOptions(ctx context.Context, opts TablePrinterOptions) TablePrinter {
+func NewTablePrinterWithOptions(ctx context.Context, topts ...TablePrinterOption) TablePrinter {
```
nderjung: I would combine these into one function since we are using variadic positional arguments:
```go
func NewTablePrinter(ctx context.Context, tops ...TablePrinterOption) TablePrinter {
```
Since the options are variadic, they can be omitted entirely if desired for "default" functionality.

nderjung (changes requested):
Hi, we're nearly there, please see inline

internal/tableprinter/table_printer.go
mdsahiloss: Done 👍🏽

internal/tableprinter/table_printer.go
mdsahiloss: Done 👍🏽

internal/tableprinter/table_printer.go
mdsahiloss: Done 👍🏽

cmd/kraft/pkg/list/list.go
```diff
@@ -147,7 +149,18 @@ func (opts *List) Run(cmd *cobra.Command, args []string) error {
...
+	switch strings.ToLower(opts.Output) {
+	case "json":
+		table = tableprinter.NewTablePrinter(ctx, tableprinter.OutputFormatJSON)
+	case "yaml":
+		table = tableprinter.NewTablePrinter(ctx, tableprinter.OutputFormatYAML)
+	case "table":
+		table = tableprinter.NewTablePrinter(ctx, tableprinter.OutputFormatTable)
+	default:
+		return fmt.Errorf("Output format %s is Invalid", opts.Output)
+	}
```
nderjung: We're re-instantiating the structure here. Instead, instantiate a slice of options, e.g.:
```go
var topts []tableprinter.TablePrinterOption
switch strings.ToLower(opts.Output) {
case "json":
  topts = append(topts, tableprinter. WithOutputFormat(tableprinter.OutputFormatJSON))
// ...
}
printer, err := tableprinter.NewTablePrinter(ctx, topts...)
```

internal/tableprinter/tableprinter_options.go
```diff
@@ -0,0 +1,17 @@
...
+type TableOutputFormat string
+
+const (
+	OutputFormatTable = TableOutputFormat("table")
+	OutputFormatJSON  = TableOutputFormat("json")
+	OutputFormatYAML  = TableOutputFormat("yaml")
+)
+
+type TablePrinterOption func(*TablePrinterOptions)
+
```
nderjung: Let's add an additional option `WithIsTTY(istty bool)` that sets the internal value for `istty`. Also please can you lowecase this internal attribute. This will allow us to prevent use of `iostreams` within the package.

mdsahiloss: Done 👍🏽

cmd/kraft/pkg/list/list.go
mdsahiloss: Done 👍🏽

internal/tableprinter/tableprinter.go
```diff
@@ -25,132 +25,94 @@ package tableprinter
...
+// NewTablePrinter ...
+func NewTablePrinter(ctx context.Context, topts ...TablePrinterOption) (*TablePrinter, error) {
+	printer := TablePrinter{
+		format:       OutputFormatTable,
+		delimeter:    DefaultDelimeter,
+		truncateFunc: text.Truncate,
+	}
+
+	for _, tpo := range topts {
+		tpo(&printer)
```
nderjung:
```diff
- tpo(&printer)
+ if err := tpo(&printer); err != nil {
+   return nil, err
+ }
```

internal/tableprinter/tableprinter.go
```diff
@@ -25,132 +25,94 @@ package tableprinter
...
+	delimeter    string
+	truncateFunc func(int, string) string
 }
 
-type ttyTablePrinter struct {
-	out      io.Writer
-	maxWidth int
-	rows     [][]tableField
-}
+// NewTablePrinter ...
```
nderjung: Missing constructor comment

nderjung (commented):
Hi, we just need a few more method comments, those that are just `...`

internal/tableprinter/tableprinter.go
mdsahiloss: Done 👍🏽

internal/tableprinter/tableprinter.go
mdsahiloss: Done 👍🏽

internal/tableprinter/tableprinter_options.go
```diff
@@ -0,0 +1,49 @@
...
+// WithOutputFormat returns a function func(opts *TablePrinter) that sets `format` in TablePrinter pointer instance.
+func WithOutputFormat(format TableOutputFormat) TablePrinterOption {
+	return func(opts *TablePrinter) error {
+		opts.format = format
+		return nil
+	}
+}
+
+// WithOutputFormatFromString returns a function func(opts *TablePrinter) that sets `format` in TablePrinter pointer instance of type `TableOutputFormat` from string.
```
nderjung: Great! All we need to do here is line-wrap all comments to 80 characters.

mdsahiloss: Done 👍🏽

PR #502, opened 2023-06-07
fix(packmanager,oci): Correctly handle options

LucaSeri:
An additional positional argument was introduced in ebe06aed9744f334a9fddb0f6d8dafe0f1bc4882 such that the package manager could be instantiated with additional options, but these options were never copied over in order to be passed to the actual constructor.

### Prerequisite checklist


nderjung (approved):
Nice catch! Thanks!

PR #506, opened 2023-06-09
feat: Initial networking support

nderjung:
### Prerequisite checklist

### Description of changes

This PR introduces essential networking support to KraftKit which facilitates the connection of a unikernel "machine" instance to a desired network.  At a high-level, this makes using the `kraft` binary ultimately useful since most unikernel instances are connected to a network. The implementation has been designed to allow for future flexibility in terms of:
1. Providing the machine interface implementation enough information to instantiate the unikernel with networking information (ultimately derived from Unikraft's internal `netdev` interface and its requirements); and,
2. The ability to plug-in different network implementations in the future by defining both prototypes representing actions and structures representing both network systems and their topologies.

This work is the culmination of the changes from #489 which has eased the introducing a new API subsystem (`network/v1alpha1`) without having to worry about a). state management, since this is handled by the [Zip API Object Framework](https://api.zip) and the [internal embedded store implementation](https://github.com/unikraft/kraftkit/blob/c6d9f4cfe336c88961eecb134811f965af1d1f33/machine/store/embedded.go) based on the [`WithStore` Zip API option](https://github.com/api-zip/api-zip/blob/336e3b443cf4a2d684f2b781d2206394150642d3/client_options.go#L32-L146) b). the ability to use the network service via a controller-pattern, enabling its use both via in-line function calls and later external calls.

#### Command-line usage

At the command-line level, networking can now be used in two different ways with the `kraft` binary:

1. **Exposing a port mapping to the host.** In this fashion, we aim to mimic existing container runtime tools like `docker` and `nerdctl` which have a `-p` flag.  In the following example, the Unikraft's NGINX unikernel port of `80` is mapped to the host at `8080`, such that it serves at http://localhost:8080:
   ```
   kraft run -p 8080:80 unikraft.org/nginx:latest
   ```

2. **Using a more complex networking setup with a specific driver and network name.**  With the introduction of a `--network` flag in `kraft run` comes the ability to attach a unikernel machine instance to a desired network.  The flag splits the provided argument into two, the driver name separated with a colon `:` and the network name.  For example, using the bridge driver to connect to the `kraft0` network:
   ```
   kraft run --network bridge:kraft0 unikraft.org/nginx:latest
   ```

For network management, an additional suite of subcommands have been introduced to govern new and existing networks in a CRUD-like fashion following the prototypes delivered by the `api/network/v1alpha1` service.  These are available under the `kraft net` subcommand:

0. **Setting use of a specific driver.** Since the underlying implementation allows for managing different network implementations, or drivers, a top-level flag `--driver` is provided in `kraft net` to switch between implementations, e.g.:
   ```
   kraft net --driver bridge
   ```

1. **Create a new network.** To build a new network, the gateway and subnet are necessary with the `-n|--network` flag which parses the input in CIDR format.  Since the default (and only) driver is `bridge`, omitting the `--driver` flag is possible since this will resolve to the only implementation. Thus, to create a new bridge network we can run:
   ```
   kraft net create -n 172.88.0.1/24 kraft0
   ```
   
2. **Listing existing networks.** Achievable easily through the `ls` subcommand:
   ```
   kraft net ls
   NAME      NETWORK        DRIVER  STATUS
   cni0      10.42.0.1/24   bridge  up
   kraft0    172.88.0.1/24  bridge  up
   nerdctl0  10.4.0.1/24    bridge  up
   ```
3. **State management.** Starting and stopping a network is provided via the `up` and `down` subcommands, respectively:
   ```
   kraft net up kraft0
   kraft net down kraft0
   ```
4. **Deleting a network.** Achievable through the `rm` subcommand.  This will also remove all associated network interfaces:
   ```
   kraft net rm kraft0
   ```

#### Bridge networking

The first implementation provided by this PR is [bridge networking](https://en.wikipedia.org/wiki/Network_bridge) largely facilitated through the highly versatile [`netlink`](https://github.com/vishvananda/netlink) module by @vishvananda.  The implementation reads existing bridges and populates accordingly.

#### API Usage

```go
import (
	"fmt"

	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

	networkapi "kraftkit.sh/api/network/v1alpha1"
	"kraftkit.sh/machine/network"
)

driverName := "bridge"
networkName := "kraft0"

strategy, ok := network.Strategies()[networkName]
if !ok {
	return fmt.Errorf("unsupported network driver strategy: %s", networkName)
}

controller, err := strategy.NewNetworkV1alpha1(ctx)
if err != nil {
	return err
}

network, err := controller.Get(ctx, &networkapi.Network{
	ObjectMeta: metav1.ObjectMeta{
		Name: networkName,
	},
})
if err != nil {
	return err
}
```

#### Acknowledgements

A special mention to @craciunoiuc who originally built an MVP of the networking implementation into KraftKit which was designed around the ideas from Docker's [libnetwork](https://github.com/docker/libnetwork) which ultimately inspired the implementation provided here.


unikraft/export/v0/uknetdev/params.go
```diff
@@ -0,0 +1,22 @@
...
+package uknetdev
+
+import (
+	"kraftkit.sh/unikraft/export/v0/ukargparse"
+)
+
+var (
+	ParamIpv4Addr       = ukargparse.ParamStr("netdev", "ipv4_addr", nil)
+	ParamIpv4SubnetMask = ukargparse.ParamStr("netdev", "ipv4_subnet_mask", nil)
+	ParamIpv4GwAddr     = ukargparse.ParamStr("netdev", "ipv4_gw_addr", nil)
```
craciunoiuc, line 14: I know there's a lot more arguments for netdev (dns/ipv6, etc.). What's the status on them?

craciunoiuc (commented):
Took an initial look over all the code.

It may come as a surprise, but I did not find any typos, lingering comments, etc.

I'll have to test all the commands to make sure they actually work and come back with a review then

nderjung:
I am also amazed 🥳

Regarding the arguments: I hope we can generate these automatically one day, post Unikraft v1.0, into directories representing the Unikraft version. We can also generate C-Go bindings too in the same package. For now, this is all done manually via v0.

unikraft/export/v0/uknetdev/params.go
nderjung: They need to be added, for now this is enough. Hopefully a more systematic way in the future.

craciunoiuc (approved):
All good here. Thanks!

PR #532, opened 2023-06-20
fix(run): Fix miscellaneous `kraft run` issues

nderjung:
### Prerequisite checklist

### Description of changes

This PR adjusts the `kraft run` command in several ways which improve its usability:
* Correctly propagate the application arguments when using the kernel runner;
* Fix an issue where the network was pre-emptively removed before the invocation of the unikernel.  This is because the `defer` clean up was migrated into the network parser.  Of course, at the end of this function it is invoked.  Instead, it should be part of the main method that handles the creation and start of the unikernel such that in this lifecycle it is handled correctly;
* Optimize the execution flow by reading the `--as` flag early: when it is set to "kernel", we can skip the instantiating
of the umbrella package manager which would otherwise invoke needless cycles on a service that is not used.  Instead, only call this on runners that rely on this necessary KraftKit subsystem;
* Move the `runner` interface to its own file to keep the `run.go` code base simpler.

### DIFF
--- a/cmd/kraft/run/run.go
+++ b/cmd/kraft/run/run.go
@@ -188,14 +188,6 @@ func (opts *Run) Pre(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	// Set use of the global package manager.
-	pm, err := packmanager.NewUmbrellaManager(ctx)
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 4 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #542, opened 2023-06-25
fix(unikraft): Better handling of schema transformations

nderjung:
### Prerequisite checklist

### Description of changes

This PR fixes a number of issues related to parsing the schema and the values of attributes related to the `version`.  Particularly:

* Fixes additional types which can be set within the Kraftfile's specification.
* Fixes an issue where a directory can be specified as the source.  This is useful in circumstances where the user is directly manipulating the library's source code or it is not a public/upstream microlibrary/component.  Setting the path will correctly resolve the manifest's `DirectoryProvider`. 

This PR can be tested by specifying a source and version transformations.  For example, setting a path on disk of a microlibrary or a commit version that only contains numerical value.

### DIFF
--- a/unikraft/component/component.go
+++ b/unikraft/component/component.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strconv"
 
 	"kraftkit.sh/kconfig"
 	"kraftkit.sh/unikraft"
@@ -59,7 +60,14 @@ func TranslateFromSchema(props interface{}) (map[string]interface{}, error) {
 		for key, prop := range entry {
 			switch key {
 			case "version":
-				component["version"] = prop.(string)
+				switch tprop := prop.(type) {
+				case string:
+					component["version"] = tprop
+				case int:
+					component["version"] = strconv.Itoa(tprop)
+				default:
+					component["version"] = fmt.Sprint(prop)
+				}
 
 			case "source":
 				for k, v := range parseStringProp(prop.(string)) {
--- a/unikraft/core/transform.go
+++ b/unikraft/core/transform.go
@@ -6,10 +6,12 @@ package core
 
 import (
 	"context"
+	"os"
 
 	"kraftkit.sh/kconfig"
 	"kraftkit.sh/unikraft"
 	"kraftkit.sh/unikraft/component"
+	"kraftkit.sh/utils"
 )
 
 // TransformFromSchema parses an input schema and returns an instantiated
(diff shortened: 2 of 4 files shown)


unikraft/component/component.go
```diff
@@ -59,7 +60,14 @@ func TranslateFromSchema(props interface{}) (map[string]interface{}, error) {
...
 			switch key {
 			case "version":
-				component["version"] = prop.(string)
+				switch tprop := prop.(type) {
+				case string:
+					component["version"] = tprop
+				case int:
+					component["version"] = strconv.Itoa(tprop)
+				default:
+					component["version"] = fmt.Sprintf("%s", prop)
```
antoineco: Is the format string really required here?

```suggestion
					component["version"] = fmt.Sprint(prop)
```

unikraft/core/transform.go
```diff
@@ -33,6 +35,14 @@ func TransformFromSchema(ctx context.Context, props interface{}) (interface{}, e
...
 	if source, ok := c["source"]; ok {
 		core.source = source.(string)
+
+		if f, err := os.Stat(core.source); err == nil && f.IsDir() {
+			if uk != nil && uk.UK_BASE != "" {
+				core.path = utils.RelativePath(uk.UK_BASE, core.source)
+			} else {
+				core.path = core.source
+			}
+		}
```
antoineco: From looking at this code, I'm not sure whether not setting `core.path` due to a non-nil `err` would cause any issue.
IMO that intention could be better conveyed by handling the error, with the cases that make we deem acceptable explicitly defined:

```golang
	f, err := os.Stat(core.source)
	if err != nil && !os.IsNotExist(err) {
		return fmt.Errorf("getting file information for %s: %w", core.source, err)
	}
	// in the following code, it's now clear that os.ErrNotExist is non-fatal
	if f != nil && f.IsDir() {
		if uk != nil && uk.UK_BASE != "" {
			core.path = utils.RelativePath(uk.UK_BASE, core.source)
		} else {
			core.path = core.source
		}
	}
```

unikraft/lib/transform.go
```diff
@@ -38,6 +40,16 @@ func TransformFromSchema(ctx context.Context, name string, props interface{}) (L
...
+
+		if f, err := os.Stat(lib.source); err == nil && f.IsDir() {
+			if f, err := os.Stat(lib.source); err == nil && f.IsDir() {
+				if uk != nil && uk.UK_BASE != "" {
+					lib.path = utils.RelativePath(uk.UK_BASE, lib.source)
+				} else {
+					lib.path = lib.source
+				}
+			}
+		}
```
antoineco: Same comment

unikraft/core/transform.go
nderjung: In this scenario, we don't want to set `core.path = core.source` _unless_  `core.source` is a directory. When it is not a directory, the `path` is the location on disk of the extracted microlibrary, e.g. `.unikraft/libs/lwip`. This means that the suggested first error would fail on all build scenarios where the `source:` attribute in the Kraftfile is not a directory.

nderjung: I've added a comment inline to clarify the reasoning.

antoineco: It wouldn't fail if we handle the error: `if err != nil && !os.IsNotExist(err)`

antoineco (approved):

PR #543, opened 2023-06-25
fix(app): Create the `BuildDir` if it does not exist

nderjung:
### Prerequisite checklist

### Description of changes

This PR is a fix related to a recent change (50d2b33) wherein the build directory was migrated into `.unikraft/build`.  Since this directory will not exist before invoking Unikraft's build system, create it.  Without this change, invocations to Unikraft's build system fail.

### DIFF
--- a/unikraft/app/project.go
+++ b/unikraft/app/project.go
@@ -19,6 +19,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"kraftkit.sh/kconfig"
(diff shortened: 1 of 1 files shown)


unikraft/app/project.go
```diff
@@ -114,6 +115,12 @@ func NewProjectFromOptions(ctx context.Context, opts ...ProjectOption) (Applicat
 		BUILD_DIR: popts.RelativePath(outdir),
 	}
 
+	if _, err := os.Stat(uk.BUILD_DIR); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(uk.BUILD_DIR, 0o755); err != nil {
+			return nil, err
```
antoineco: nit: it's a good practice to wrap some context around the raw error:

```suggestion
			return nil, fmt.Errorf("creating build directory: %w", err)
```

When the user encounters an error, it's easier for them to understand the original intention of the code, as opposed to just `error creating /some/obsure/path`.

It's a very general comment, and I've seen that pattern missing across the code base, but it's never too late to start applying these small qulity of life improvements.

antoineco (approved):

PR #544, opened 2023-06-25
fix(run): Fix issues related to project runner

nderjung:
### Prerequisite checklist

### Description of changes

This PR fixes two issues related to using the project runner in `kraft run`:

* Firstly, a new check occurs whether the runner-type has been pre-emptively set to "project" or "kernel".  These runners do not require access to the package manager which would otherwise be instantiated, slowing down the instantiation process via `kraft run`. In the future, it may make sense to simply embbed the user of the package manager in the relevant runner.
* Secondly, correctly set application arguments in project runner which stems from misparsing the cmdline arguments provided within `kraft run` as well as not setting them in the machine specification.

### DIFF
--- a/cmd/kraft/run/runner_project.go
+++ b/cmd/kraft/run/runner_project.go
@@ -38,14 +38,20 @@ func (runner *runnerProject) String() string {
 func (runner *runnerProject) Runnable(ctx context.Context, opts *Run, args ...string) (bool, error) {
 	var err error
 
+	cwd, err := os.Getwd()
+	if err != nil {
+		return false, fmt.Errorf("getting current working directory: %w", err)
+	}
+
 	if len(args) == 0 {
-		runner.workdir, err = os.Getwd()
-		if err != nil {
-			return false, err
-		}
+		runner.workdir = cwd
 	} else {
-		runner.workdir = args[0]
-		runner.args = args[1:]
+		runner.workdir = cwd
+		runner.args = args
+		if f, err := os.Stat(args[0]); err == nil && f.IsDir() {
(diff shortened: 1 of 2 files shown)


cmd/kraft/run/runner_project.go
```diff
@@ -38,14 +38,21 @@ func (runner *runnerProject) String() string {
 func (runner *runnerProject) Runnable(ctx context.Context, opts *Run, args ...string) (bool, error) {
 	var err error
 
+	cwd, err := os.Getwd()
+	if err != nil {
+		return false, err
```
antoineco: Let's return the error with its context for easier troubleshooting:

```suggestion
		return false, fmt.Errorf("getting current working directory: %w", err)
```

cmd/kraft/run/runner_project.go
```diff
@@ -38,14 +38,21 @@ func (runner *runnerProject) String() string {
...
 	} else {
-		runner.workdir = args[0]
-		runner.args = args[1:]
+		if f, err := os.Stat(args[0]); err == nil && f.IsDir() {
+			runner.workdir = args[0]
+			runner.args = args[1:]
+		} else {
+			runner.workdir = cwd
+			runner.args = args
+		}
```
antoineco: Style suggestion to avoid the `else` (feel free to discard)
```suggestion
		runner.workdir = cwd
		runner.args = args
		if f, err := os.Stat(args[0]); err == nil && f.IsDir() {
			runner.workdir = args[0]
			runner.args = args[1:]
		}
```

antoineco (approved):

PR #545, opened 2023-06-26
fix: flag attribution to struct fields of composite type

### LINKED ISSUE #537: Non primitive types aren't parsed correctly
### Describe the bug

Another `cmdfactory` parsing bug is preventing the usage of composite flag types (`[]string`, `map`, etc.).

Flags which aren't primitive types (`string`, `integer`, `boolean`) are handled incorrectly, resulting in their value remaining empty, even after parsing.

It affects:
- `kraft pkg --volume`
- `kraft run --kernel-arg`
- `kraft run --port`
- `kraft --with-mirror`
- `kraft --with-manifest`

### Steps to reproduce

Let's take the example of a config type `Config` with one primitive flag, and one composite flag:

```golang
type Config struct {
	TraceSampleRate string   `long:"trace-sample-rate" usage:"Tracing sample rate."`
	TraceExporters  []string `long:"trace-exporter" usage:"Tracing exporters."`
}
```

Call `cmdfactory.AttributeFlags` with that config type as argument:

https://github.com/unikraft/kraftkit/blob/fef0f266d1cb62f9dbfc102ca88bf995f05d30dc/internal/cli/options.go#L132-L134

Run the `dlv` debugger, passing the following flags to the command: `--trace-exporter=otlp --trace-sample-rate=10`.

Inside `cmdfactory.AttributeFlags`, set a breakpoint at the end of the function:

https://github.com/unikraft/kraftkit/blob/fef0f266d1cb62f9dbfc102ca88bf995f05d30dc/cmdfactory/builder.go#L330-L332

All flags _(just 1 in this example for simplification)_ are present in the `c.pflags.actual` list (_with_ "p"), except `trace-exporter`:

```gdb
(dlv) p c.pflags.actual
map[github.com/spf13/pflag.NormalizedName]*github.com/spf13/pflag.Flag [
        "trace-sample-rate": *{
                Name: "trace-sample-rate",
                Shorthand: "",
                Usage: "Tracing sample rate.",
                Value: github.com/spf13/pflag.Value(*github.com/spf13/pflag.stringValue) ...,
                DefValue: "",
                Changed: true,
                NoOptDefVal: "",
                Deprecated: "",
                Hidden: false,
                ShorthandDeprecated: "",
                Annotations: map[string][]string nil,},
        ...
]
```

```gdb
(dlv) p c.pflags.actual["trace-exporter"]
Command failed: key not found
```

However, it is the opposite for the  `c.flags.actual` list (_without_ "p"); here `trace-exporter` is the only entry:

```gdb

(dlv) p c.flags.actual
map[github.com/spf13/pflag.NormalizedName]*github.com/spf13/pflag.Flag [
        "trace-exporter": *{
                Name: "trace-exporter",
                Shorthand: "",
                Usage: "Tracing exporters.",
                Value: github.com/spf13/pflag.Value(*github.com/spf13/pflag.stringSliceValue) ...,
                DefValue: "[]",
                Changed: true,
                NoOptDefVal: "",
                Deprecated: "",
                Hidden: false,
                ShorthandDeprecated: "",
                Annotations: map[string][]string nil,},
]
```

In both cases, **the `Value` field holds the expected parsed value**:

```gdb
(dlv) p c.flags.actual["trace-exporter"].Value
github.com/spf13/pflag.Value(*github.com/spf13/pflag.stringSliceValue) *{
        value: *[]string len: 1, cap: 1, ["otlp"],
        changed: true,}
```
```gdb
(dlv) p c.pflags.actual["trace-sample-rate"].Value
github.com/spf13/pflag.Value(*github.com/spf13/pflag.stringValue) *"10"
```

However, only the primitive flag `trace-sample-rate` made it to the `Config` struct, whereas the composite flag `trace-exporter` retained an empty value:

```gdb
=>  132:         if err := cmdfactory.AttributeFlags(cmd, cfgm.Config, os.Args[1:]...); err != nil {
    133:                 return err
    134:         }
(dlv) p cfgm
*myproject/pkg/config.Config] {
        Config: *myproject/pkg/config.Config {
                TraceSampleRate: "10",
                TraceExporters: []string len: 0, cap: 0, nil,},
        ConfigFile: "",
        Feeders: []kraftkit.sh/config.Feeder len: 0, cap: 0, nil,}
```

### Expected behavior

_No response_

_No response_

_No response_

### Relevant log output

_No response_

antoineco:
### Prerequisite checklist

### Description of changes

Copied from https://github.com/unikraft/kraftkit/issues/537#issuecomment-1608264469:

`cmdfactory` calls `AttributeFlags` with the **root** command (and all args + flags). The `RunE` function, overridden by `bind()`, is therefore assigned to the **root** command.

However, the root command itself is never executed by cobra, only whichever _sub_ command was selected by the use, so `bind()` has absolutely no effect. Its result is never called.

This PR ensures that `AttributeFlags` is called with the subcommand invoked by the user, instead of the root command.

GitHub-Fixes #537

### DIFF
--- a/cmdfactory/builder_test.go
+++ b/cmdfactory/builder_test.go
@@ -6,11 +6,105 @@
 package cmdfactory
 
 import (
+	"os"
 	"testing"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
+
+func TestAttributeFlags_StructFields(t *testing.T) {
+	attributeFlags := func(t *testing.T, obj any, args ...string) {
+		t.Helper()
+
+		rootCmd, subCmds := []string{"kraft"}, []string{"cmd1", "subcmd1"}
+		cmd := makeCommand(append(rootCmd, subCmds...)...)
+
+		allArgs := append(args, subCmds...)
+		os.Args = append([]string{os.Args[1]}, allArgs...)
+
+		// AttributeFlags also populates cmd's private flag fields.
+		if err := AttributeFlags(cmd, obj, allArgs...); err != nil {
+			t.Fatal("Failed to associate flags with struct fields:", err)
+		}
+
+		// Execute command to invoke cmd.RunE. This executes all middlewares injected
+		// by bind(), which role is to assign parsed flag values to struct fields.
+		if _, err := executeC(cmd); err != nil {
+			t.Fatal("Failed to execute command:", err)
+		}
+	}
+
+	type TestObj struct {
+		String string            `long:"string" usage:"String arg"`
+		Int    int               `long:"int" usage:"Integer arg"`
+		Bool   bool              `long:"bool" usage:"Boolean arg"`
+		Slice  []string          `long:"slice" usage:"Slice arg"`
+		Map    map[string]string `long:"map" usage:"Map arg"`
+		Nested struct {
+			String string            `long:"n-string" usage:"Nested string arg"`
+			Int    int               `long:"n-int" usage:"Nested integer arg"`
+			Bool   bool              `long:"n-bool" usage:"Nested boolean arg"`
+			Slice  []string          `long:"n-slice" usage:"Nested slice arg"`
+			Map    map[string]string `long:"n-map" usage:"Nested map arg"`
+		}
+	}
+
+	t.Run("String fields", func(t *testing.T) {
+		obj := &TestObj{}
+		attributeFlags(t, obj, "--string=val", "--n-string=n-val")
+		if expect, got := "val", obj.String; expect != got {
+			t.Errorf("Unexpected value for string struct field after flags attribution. Expected %q, got %q", expect, got)
+		}
+		if expect, got := "n-val", obj.Nested.String; expect != got {
+			t.Errorf("Unexpected value for nested string struct field after flags attribution. Expected %q, got %q", expect, got)
+		}
+	})
+
+	t.Run("Integer fields", func(t *testing.T) {
+		obj := &TestObj{}
+		attributeFlags(t, obj, "--int=1", "--n-int=2")
+		if expect, got := 1, obj.Int; expect != got {
+			t.Errorf("Unexpected value for int struct field after flags attribution. Expected %d, got %d", expect, got)
+		}
+		if expect, got := 2, obj.Nested.Int; expect != got {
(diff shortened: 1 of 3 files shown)


cmdfactory/builder_test.go
```diff
@@ -12,6 +12,98 @@ import (
...
+		obj := &TestObj{}
+		attributeFlags(t, obj, "--map=key=val", "--n-map=key=val")
+		if expect, got := map[string]string{"key": "val"}, obj.Map; !equalMaps(got, expect) {
+			t.Errorf("Unexpected value for map struct field after flags attribution. Expected %v, got %v", expect, got)
+		}
+		if expect, got := map[string]string{"key": "val"}, obj.Nested.Map; !equalMaps(got, expect) {
+			t.Errorf("Unexpected value for nested map struct field after flags attribution. Expected %v, got %v", expect, got)
+		}
+	})
+}
```
antoineco, line 107: These two cases won't work, and addressing these will most likely require a partial redesign of `cmdfactory`.

To summarize my investigation:
- cobra executes the `RunE` of the _subcommand_
- cmdfactory sets (though `bind()`) the `RunE` of the _parent_ command

Unfortunately, a fix wouldn't be as simple as moving the initializations performed by cmdfactory to the subcommand, because that would break other things, such as the registration of global flags, etc.

cmdfactory/builder_test.go
```diff
@@ -12,6 +12,99 @@ import (
...
+	t.Run("Slice fields", func(t *testing.T) {
+		obj := &TestObj{}
+		attributeFlags(t, obj, "--slice=val1", "--slice=val2", "--n-slice=val1,val2")
+		if expect, got := []string{"val1", "val2"}, obj.Slice; !equalSlices(got, expect) {
+			t.Errorf("Unexpected value for slice struct field after flags attribution. Expected %v, got %v", expect, got)
+		}
+		if expect, got := []string{"val1", "val2"}, obj.Nested.Slice; !equalSlices(got, expect) {
+			t.Errorf("Unexpected value for nested slice struct field after flags attribution. Expected %v, got %v", expect, got)
+		}
+	})
```
antoineco, line 95: This test is still failing.

Because flags are being parsed twice (once in cmdfactory, once in cobra), we are getting duplicated values in string slices.

craciunoiuc (approved):
All good here. Thanks!

PR #567, opened 2023-07-01
fix(manifest): Pull based also on unikraft versions

### LINKED ISSUE #561: Version number format for libraries doesn't work
### Describe the bug

The build process failed to find libraries if they have a version number in the `kraft.yml` file instead of `stable` even though the version number is valid.

### Steps to reproduce

```bash
git clone --depth=1 https://github.com/unikraft/app-nginx.git;
cd app-nginx;
# Modify the `kraft.yml` file and change `stable` version for `musl` to version number format( e.g. "0.11.0")
# I checked version number and it exists in the manifest file of `kraftkit`
kraft build --plat qemu --arch x86_64;
# It will end up like this
```
image

The same it's happens with `app-helloworld-go`.

### Expected behavior

The build process should run without any issues.

x86_64

linux/debian

### Relevant log output

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

This fix adds checks for pulling based on unikraft versions also. Support was already in place, but probably was broken in the last releases.

Closes: #561

### DIFF
--- a/manifest/manager.go
+++ b/manifest/manager.go
@@ -285,6 +285,10 @@ func (m *manifestManager) Catalog(ctx context.Context, qopts ...packmanager.Quer
 					versions = append(versions, v.Version)
 					break
 				}
+				if v.Unikraft == version {
+					versions = append(versions, v.Unikraft)
+					break
+				}
 			}
 			if len(versions) == 0 {
 				for _, channel := range manifest.Channels {
--- a/manifest/pack.go
+++ b/manifest/pack.go
@@ -37,7 +37,7 @@ func NewPackageFromManifestWithVersion(manifest *Manifest, version string, opts 
 	}
 
 	for _, ver := range manifest.Versions {
-		if ver.Version == version {
+		if ver.Version == version || ver.Unikraft == version {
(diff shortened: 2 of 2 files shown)


antoineco (approved):

nderjung:
🙃

nderjung:
This addition means that both the Unikraft version and the library's version are possible in the `version` attribute of the `Kraftfile`.

I fear this has an unintended consequence. Specifying the version based on Unikraft, in my opinion, should not be used as the library's version. This can create a confusion between when a library's version is the same as the Unikraft version.

Whilst not yet implemented, [the versioning WG track on the Unikraft core](https://docs.google.com/document/d/1A-CAss5RvgYapg3YO8GNCdMki6cgq_7XG5om8nVWWGk/edit) aims to enable passing the actual origin library's specific version.  When an origin library and Unikraft both have a same version, kraftkit will not know which to select, or will select the origin library's version -- since it is the first to be checked in the sequence this modifies.

The used attribute Unikraft is meant to be an indicator of the minimum version of Unikraft that the library supports at `version`.

PR #595, opened 2023-07-05
fix(run): Fix the order of runners

nderjung:
### Prerequisite checklist

### Description of changes

This PR fixes an out-of-order issue which would occassionally misdetermine what to run.  The issue is found with the use of hashmap which does not maintain order vs. a slice which does.  Necessary adjustments are made to rework the code to use a slice for the list of available runners.  The order is important because the runners based on a package manager should come last, as these can require network connectivity and would otherwise slow down the check for more lightweight runners that are local to the system.


craciunoiuc (approved):
All good here. Thanks!

PR #596, opened 2023-07-05
refactor(build): Move target selection to beginning of execution

nderjung:
### Prerequisite checklist

### Description of changes

This commit moves the dynamic selection of a target from a user input to the beginning of the `Run` method such that it is the first operation performed.  This not only makes the output slightly cleaner when using debug/trace-logging, but also solves an issue where the width of each process can be correctly determined.  In the latter fix, the width was incorrectly determined based on all project targets, and not the narrower focus based on the prompt or user-provided input.


craciunoiuc (approved):
All good here. Thanks!

PR #615, opened 2023-07-07
feat(webinstall): Create kraftkit group and well known paths

### LINKED ISSUE #585: Create `kraftkit` group and chown runtime dir
### Feature request summary

This issue tracks creating a new group `kraftkit` in the [webinstall script](https://github.com/unikraft/kraftkit/blob/staging/tools/webinstall/install.sh) and relevant distribution scripts when detecting non-root users. The current user should be added to this group and the directory `/var/kraftkit` which represents the KraftKit runtime directory should be `chown`ed to this group and `chmod`ed to `775` such that user can read and write to this directory. This will prevent runtime errors with `kraft run` in case the user needs to use `sudo` with `kraft run` to access a privileged mode with `/dev/kvm`.

### Describe alternatives

_No response_

### Related architectures

None

### Related platforms

None

### Additional context

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

The script now creates a `kraftkit` group and adds the user running it to the group.
It then creates all paths used inside kraftkit that need permissions (currently just `/var/kraftkit`, open to suggestions).

❗ The script does not change existing files in those paths, for that you can use `sudo find /var/kraftkit -exec chgrp kraftkit {} \;` ❗
❗ The shell requires restarting for changes to take effect ❗

Closes: #585


tools/webinstall/install.sh
```diff
@@ -1153,6 +1160,85 @@ install_linux_manual() {
     _CLEANUP_VERSION=""
 }
 
+# check_environment checks the environment for the kraftkit group.
+# If the group already exists, it means that it owns all default files.
+# $1: the group name
+# Returns:
+# Code: 0 if group doesn't exist, 1 if it exists
+check_environment() {
```
nderjung: I'd rename `environment` (in the two newly introduced methods) to `usergroup` to prevent confusion with the idea of environmental variables.

tools/webinstall/install.sh
```diff
@@ -1290,13 +1376,22 @@ main() {
...
     # Install kraftkit for the given architecture
     install_kraftkit "$_main_arch" "$_main_auto_install"
     say "kraftkit was installed successfully to $PREFIX/kraft"
 
     # Check if kraft is installed and working
     kraft -h
 
+    # Check if we need to ask for restartqq
+    if [ "$_main_new_group" -eq 0 ]; then
+        say "please restart your shell to make kraftkit run commands available"
```
nderjung: https://unix.stackexchange.com/questions/18796/how-to-apply-changes-of-newly-added-user-groups-without-needing-to-reboot
or
https://superuser.com/questions/272061/reload-a-linux-users-group-assignments-without-logging-out

tools/webinstall/install.sh
```diff
@@ -1153,6 +1160,85 @@ install_linux_manual() {
...
+    do_cmd "$GROUPADD $_cre_group"
+
+    say_debug "detecting current user"
+    _cre_user="$($WHOAMI)"
+
+    say "adding $_cre_user to $_cre_group group"
+    do_cmd "$USERMOD -a -G $_cre_group $_cre_user"
+
+    say "creating default directories"
+    _cre_default_dirs="/var/kraftkit"
```
nderjung: Can you make this a global program variable, `KRAFTKIT_RUNTIME_DIR` (which matches [the existing env var](https://github.com/unikraft/kraftkit/blob/staging/config/config.go#L26)). We also need a way to set this automatically if it is changed in the `~/.config/kraftkit/config.yaml` file. But let's leave that latter part for another day (e.g. when we have something like `kraft config set ...`)

nderjung:
Overall looks good! Thanks for tackling this. I still need to try this out locally and will get back to you.

craciunoiuc:
Fixed your requests. The script now prompts for restart and skips the restart for non-interactive shells.

There are some (big) changes from the previous version, so go over the code again please 🙏

tools/webinstall/install.sh
```diff
@@ -1153,6 +1163,116 @@ install_linux_manual() {
...
+            say_debug "moving to new shell"
+            do_cmd "$SU -p $_sug_user"
+        elif printf "%s" "$_sug_answer" | "$GREP" -q -E "$_NO_ANS"; then
+            say_debug "keeping current shell"
+        else
+            err "error: choose either yes or no."
+        fi
+    else
+        say_debug "non-interactive shell, not switching"
+    fi
```
nderjung:
```
say_debug "skipping shell restart"
```

PR #629, opened 2023-07-15
feat: Native volume support

nderjung:
### Prerequisite checklist

### Description of changes

This PR introduces native volume support into KraftKit which ultimately allows users to bind devices to the unikernel in a simple and expressive manner. At a high-level, this allows users to specify devices they would like to mount to a unikernel in `kraft run` with the introduction of a new flag `-v|--volume`.

Practically, the implementation is built around the same API definitions that are used to express both a machine instance but also the networking setup.

QEMU is the only machine driver implementation to recognise and act on the updated machine specification that describes a volume.  Additionally, the only driver implementation for volumes is 9pfs, which is a lightweight passthrough mechanism for mounting a source path on the host in a bidirectional manner.  The utility of this is that a user can make adjustments on their host and have those immediately reflected within the instance.

GitHub-Depends: #545

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -13,12 +13,14 @@ import (
 
 	machinev1alpha1 "kraftkit.sh/api/machine/v1alpha1"
 	networkv1alpha1 "kraftkit.sh/api/network/v1alpha1"
+	volumev1alpha1 "kraftkit.sh/api/volume/v1alpha1"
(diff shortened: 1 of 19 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #646, opened 2023-07-16
fix(manifest): Improve manifest Catalog caching

nderjung:
### Prerequisite checklist

### Description of changes

This PR improves in-memory caching when using the `Catalog` method.  Often, the `Catalog` method is called sequentially (for example, when finding individual components, such as alibraries, as part of a build).  In the manifest package manager implementation, the call to `Catalog` would repeatedly hit the remote manifest index and subsequentially download all latest manifests.  For larger projects that have many libraries, this is quite an expensive operation since multiple look-ups and downloads occur.  In this PR, fix this by using a temprorary in-memory index that can be referenced.  It is populated the first time when either `Update` or `Catalog` are called.

### DIFF
--- a/manifest/git.go
+++ b/manifest/git.go
@@ -96,12 +96,12 @@ func NewGitProvider(ctx context.Context, path string, opts ...ManifestOption) (P
 		return nil, fmt.Errorf("could not list remote git repository: %v", err)
 	}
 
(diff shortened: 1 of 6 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #649, opened 2023-07-17
feat: Better detection for machines when running

### LINKED ISSUE #634: `kraft run`: Consider built-in (not as module) KVM support
### Describe the bug

Currently, KraftKit checks `/proc/modules` for KVM support, assuming KVM is built as a kernel module and then loaded in the kernel.

This will cause a failure in case of kernel built-in support.

The solution to this would be to check KVM support via querying `/dev/kvm`. Note that simply checking for the existence of `/dev/kvm` is not the rigorous way, since that file can be created without KVM support. The proper way is to query KVM support via the [KVM API](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt); one way could be via `ioctl` by passing the `KVM_GET_API_VERSION` command.

### Steps to reproduce

Boot into a kernel with built-in KVM support.

### Expected behavior

The `kraft run` command should cause a running of a Unikraft instance.

x86_64

linux/debian

### Relevant log output

```shell
$ kraft run
could not determine hypervisor and system mode
```

### LINKED ISSUE #636: `kraft run`: `-W` option should not look for KVM / acceleration support
### Describe the bug

When `kraft run -W` is used (to disable KVM acceleration), there are still checks for KVM support, such as looking in `/proc/modules`. These should be disabled when `-W` is used.

This is unlikely to cause errors, as there is a fallback to check for QEMU support. But, as a good practice, when using `-W`, KVM support checks should not happen.

### Steps to reproduce

There is no error, due to the fallback check.

### Expected behavior

_No response_

x86_64

_No response_

### Relevant log output

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

GitHub-Fixes: #636
GitHub-Fixes:  #634

### DIFF
--- a/cmd/kraft/events/events.go
+++ b/cmd/kraft/events/events.go
@@ -74,12 +74,8 @@ func (opts *Events) Run(cmd *cobra.Command, args []string) error {
 	platform := mplatform.PlatformUnknown
 
 	if opts.platform == "" || opts.platform == "auto" {
-		var mode mplatform.SystemMode
-		platform, mode, err = mplatform.Detect(ctx)
-		if mode == mplatform.SystemGuest {
-			cancel()
-			return fmt.Errorf("nested virtualization not supported")
-		} else if err != nil {
+		platform, _, err = mplatform.Detect(ctx)
+		if err != nil {
 			cancel()
 			return err
 		}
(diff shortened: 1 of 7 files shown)


craciunoiuc:
this does not actually fix that issue, it needs one more commit to check if `/dev/kvm` exists and is `ioctl`-able

nderjung (approved):
Looks good! Thanks!

PR #659, opened 2023-07-19
fix(cli): Select targets only once

### LINKED ISSUE #658: kraft build appends desired target twice in selection screen
### Describe the bug

When executing the command kraft build -p xen -m arm64 -t target1 on a kraft.yaml file containing a unique target described by the platform, architecture, and name combination, the expected behavior is for the command to match and select only the target "target1 (xen/arm64)".

However, the current behavior is that the desired target (xen/arm64) gets appended twice during the filtering process, and it does not get selected exclusively. This is inconsistent with the expected behavior, as the platform, architecture, and name combination describes the target uniquely.

### Steps to reproduce

Given this sample kraft.yaml file:

```yaml
specification: '0.5'
name: helloworld
unikraft:
  version: stable
targets:
  - architecture: x86_64
    platform: kvm
    name: target1
  - architecture: arm64
    platform: xen
    name: target1
```

Run the command: 

```sh
kraft build -p xen -m arm64 -t target1
```

The user is taken to the selection screen and sees:

```sh
▸ target1 (qemu/x86_64)
target1 (xen/arm64)
target1 (xen/arm64)
```

### Expected behavior

The command should match only the target "target1 (xen/arm64)". The user should not see the selection screen.

_No response_

_No response_

### Relevant log output

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

If the loop is not exited, there is a good chance that a target meets several conditions at the same time and is added multiple times, as described in the issue.

GitHub-Fixes: #658

### DIFF
--- a/internal/cli/target.go
+++ b/internal/cli/target.go
@@ -95,6 +95,7 @@ func FilterTargets(targets []target.Target, arch, plat, targ string) []target.Ta
 			}
 
 			selected = append(selected, t)
+			break
 		}
 	}
 
--- a/internal/cli/target.go
+++ b/internal/cli/target.go
@@ -66,24 +66,39 @@ func FilterTargets(targets []target.Target, arch, plat, targ string) []target.Ta
 			return len(targ) == 0 && len(arch) == 0 && len(plat) == 0
 		},
 
-		// If the `targ` is supplied and the target name match
+		// If only `targ` is supplied and the target name match
 		func(t target.Target, arch, plat, targ string) bool {
-			return len(targ) > 0 && t.Name() == targ
+			return len(targ) > 0 && len(plat) == 0 && len(arch) == 0 && t.Name() == targ
 		},
 
 		// If only `arch` is supplied and the target's arch matches
 		func(t target.Target, arch, plat, targ string) bool {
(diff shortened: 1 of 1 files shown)


jake-ciolek:
Good work on addressing the multiple addition of targets as described in the original issue. When I used the kraftfile from the original bug report, it appears that your fix solved the duplication problem.

However, the issue with the selection screen still popping up persists. It prompts me to choose between these options:

```sh
[?] select target:
  ▸ target1 (qemu/x86_64)
    target1 (xen/arm64)
```

Even though the kraft build -p xen -m arm64 -t target1 command should exclusively match "target1 (xen/arm64)" without any user interaction. I believe we should consider this issue as partially resolved, and it might be beneficial to investigate why the selection screen still appears despite providing a unique target identifier.

craciunoiuc:
makes sense, looking into now

craciunoiuc:
@jake-ciolek check again

I think previously target was to be used separately from arch/plat, I can't quite remember, anyways, I made it so that both 3 can be used

jake-ciolek (approved):
Thanks

PR #662, opened 2023-07-20
fix: Alias platform argument

### LINKED ISSUE #657: kraft build fails to correctly filter targets with "kvm" platform
### Describe the bug

Filtering using the -p flag and targetting "kvm" specifically, doesn't match targets from the kraft.yaml file.

Before filtering, "kvm" gets aliased to "qemu". As a result, the filtering matches no targets.

### Steps to reproduce

Given this sample kraft.yaml file:

```yaml
specification: '0.5'
name: helloworld
unikraft:
  version: stable
targets:
  - architecture: x86_64
    platform: kvm
  - architecture: arm64
    platform: xen
```

Run the command:

```sh
kraft build -p kvm
```

Build exits with the error:

```sh
no choices provided.
 ```

### Expected behavior

Command should select the target with platform "kvm".

_No response_

_No response_

### Relevant log output

```shell
no choices provided.
```

craciunoiuc:
### Prerequisite checklist

### Description of changes

GitHub-Fixes: #657

### DIFF
--- a/cmd/kraft/build/build.go
+++ b/cmd/kraft/build/build.go
@@ -18,6 +18,7 @@ import (
 	"kraftkit.sh/config"
 	"kraftkit.sh/exec"
 	"kraftkit.sh/internal/cli"
+	"kraftkit.sh/machine/platform"
 	"kraftkit.sh/pack"
 	"kraftkit.sh/unikraft"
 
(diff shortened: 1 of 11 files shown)


jake-ciolek:
Thanks. Would it be possible to define that code once and use it as a method or a function?

craciunoiuc:
of course :magic_wand:

craciunoiuc:
@jake-ciolek check again

Changed the logic a bit. Now it returns the same string if not found (it will fail anyways at a later step if wrong)

jake-ciolek (approved):
Thanks. I tried it out and when using "-p kvm", filtering works as expected.

PR #664, opened 2023-07-20
fix(manifest): Remove duplicate pull argument

craciunoiuc:
### Prerequisite checklist

### Description of changes

The argument to use git was duplicated by the `--git-protocol` argument and is now removed in favor of that. To use git the remaining one needs to be called like `--git-protocol git`.

The argument was also removed because it caused `cmdfactory` to behave wrongly and waaas also confusing.

### DIFF
--- a/manifest/git.go
+++ b/manifest/git.go
@@ -17,6 +17,7 @@ import (
 	gitssh "github.com/go-git/go-git/v5/plumbing/transport/ssh"
 	giturl "github.com/kubescape/go-git-url"
 
+	"kraftkit.sh/config"
 	"kraftkit.sh/log"
 	"kraftkit.sh/pack"
 	"kraftkit.sh/unikraft"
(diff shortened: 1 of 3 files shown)


nderjung:
Hmmm. I understand where you're coming from but in my mind these flags serve two different purposes. The one, `--git-protocol`, is used to switch any internal git method ultimately between https and ssh. In the end, with this flag, git is still used. This is different from this flag `kraft pkg pull --git` which indicates whether to pull a git repo or just a tarball. The latter is set if the user wants access to the git history or not.

craciunoiuc:
But is there any difference between a tarball archive and an https git link?

As far as I know, kraftkit pulls https links as tarballs

craciunoiuc:
So as far as I understand, this should not be removed, just renamed. The bug is also fixed so it should be fine.

craciunoiuc:
So as far as I understand, this should not be removed, just renamed. The bug is also fixed so it should be fine.

PR #672, opened 2023-07-25
feat(oci): Add option to package initrd directories

### LINKED ISSUE #669: `--initrd` flag is not working for directories when running `kraft pkg`
### Describe the bug

While trying to include a directory as an OCI layer, I tried using the `--initrd` flag but that doesn't seem to work with directories.
After unpacking the image, the `initrd` directory was empty
```bash
$ tree unikraft 
unikraft
└── bin
    ├── initrd
    └── kernel
2 directories, 1 file
```
kraftkit version:
```bash
$ kraft version   
kraft v0.6.4-78-g38a5be2 (38a5be2) go1.20.2 Fri Jul 21 21:40:42 UTC 2023
```

### Steps to reproduce

To reproduce the error, take any unikraft project and try to package it using `kraft pkg --as=oci` with `-i` or `--initrd` flag. 

```bash
$ KRAFTKIT_LOG_TYPE=basic kraft pkg --as=oci \
    --oci-tag=<image>:<tag> \
    -i <folder to be included> \
    --containerd-addr=/run/containerd/containerd.sock \
    --log-level=trace
```

### Expected behavior

After pulling and unpacking the OCI image, the `unikraft/bin/initrd` should not be empty

x86_64

linux/debian

### Relevant log output

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

GitHub-Fixes: #669

### DIFF
--- a/oci/pack.go
+++ b/oci/pack.go
@@ -163,9 +163,31 @@ func NewPackageFromTarget(ctx context.Context, targ target.Target, opts ...packm
 
 	if popts.Initrd() != "" {
 		log.G(ctx).Debug("oci: including initrd")
+		initRdPath := popts.Initrd()
+		if f, err := os.Stat(initRdPath); err == nil && f.IsDir() {
+			cwd, err2 := os.Getwd()
+			if err2 != nil {
+				return nil, err2
+			}
+
+			file, err := os.CreateTemp("", "kraftkit-oci-archive-*")
+			if err != nil {
+				return nil, err
+			}
+			defer os.Remove(file.Name())
+
+			cfg, err2 := initrd.NewFromMapping(cwd,
+				file.Name(),
+				fmt.Sprintf("%s:/", initRdPath))
+			if err2 != nil {
+				return nil, err2
+			}
+
+			initRdPath = cfg.Output
+		}
 		layer, err := NewLayerFromFile(ctx,
 			ocispec.MediaTypeImageLayer,
-			popts.Initrd(),
+			initRdPath,
 			WellKnownInitrdPath,
 			WithLayerAnnotation(AnnotationKernelInitrdPath, WellKnownInitrdPath),
 		)


oci/pack.go
```diff
@@ -443,11 +459,18 @@ unpack:
...
 			if err != nil {
 				return err
 			}
+		} else if err == nil && f.IsDir() {
+			ocipack.initrd, err = initrd.NewFromMapping(popts.Workdir(),
+				filepath.Join(filepath.Dir(initrdPath), "initramfs.cpio"),
+				fmt.Sprintf("%s:/", initrdPath))
+			if err != nil {
+				return err
+			}
```
nderjung: A few thoughts on this part specifically. The general implementation of this PR allows for serializing a filesystem into a single layer (which in itself outside of the context of using it as initramfs is very useful), and here we dynamically serialize it into a initramfs when provided with a directory AFTER unpacking it. I imagined when a user serialized a filesystem as a "normal" OCI layer (tarball of a directory) that this may not necessarily be initramfs (because it can be both initramfs OR it can act like a normal filesystem (9pfs) and be extracted somewhere. even one day multiple layers can be "overlayed" like how the normal OCI spec dictates). I think the tarball work is useful but should be opened up as a separate PR.

To simplify this, i would move this `initrd.NewFromMapping` to much earlier, where `NewLayerFromDirectory` is actually called. This way we pack the initramfs at build time. This has the added benefit of speeding up general instantiation because we don't have to do any "post processing" before instantiating the unikernel (untarring then turning into a CPIO).

craciunoiuc: Applied your changes.

Did not open an new PR because It's more or less the same thing 🪄

oci/pack.go
```diff
@@ -163,9 +163,25 @@ func NewPackageFromTarget(ctx context.Context, targ target.Target, opts ...packm
...
 		log.G(ctx).Debug("oci: including initrd")
+		initRdPath := popts.Initrd()
+		if f, err := os.Stat(initRdPath); err == nil && f.IsDir() {
+			cwd, err2 := os.Getwd()
+			if err2 != nil {
+				return nil, err2
+			}
+
+			cfg, err2 := initrd.NewFromMapping(cwd,
+				filepath.Join(filepath.Dir(initRdPath), "initramfs.cpio"),
```
nderjung: I'd turn `filepath.Dir(initRdPath), "initramfs.cpio"` into a file in tempfs, we don't want to overwrite user content potentially.

oci/pack.go
```diff
@@ -163,9 +163,30 @@ func NewPackageFromTarget(ctx context.Context, targ target.Target, opts ...packm
...
 	if popts.Initrd() != "" {
 		log.G(ctx).Debug("oci: including initrd")
+		initRdPath := popts.Initrd()
+		if f, err := os.Stat(initRdPath); err == nil && f.IsDir() {
+			cwd, err2 := os.Getwd()
+			if err2 != nil {
+				return nil, err2
+			}
+
+			file, err := os.CreateTemp("", "kraftkit-oci-archive-*")
```
nderjung, line 173: I just noticed we somehow have to clean this file up once it has been packaged.

Maybe introduced a higher-level string slice, `cleanUpFiles`, that on `defer` is iterated on and each file is removed in this slice.

craciunoiuc: We never clean up after ourselves, when I package stuff tmp it's always filled with junk :D

nderjung: I'll open an issue that tackles looking for and handling cleanups. In the mean time, we can solve the first here :)

nderjung: Rendered in #679

craciunoiuc: check again

nderjung (approved):
Thanks!

Issue #579, opened 2023-07-03
Added features to KraftKit devcontainer


nderjung:
### Feature request summary

This issue tracks several additional features to the devcontainer for the KraftKit repository, specifically:
* Adding the `dist/` folder in the environmental PATH variable, such that following a successful `make kraft`, the binary is immediately usable;
* Adding Golang extension to access the language server;
* Adding the build tools from the base environment and the QEMU binaries into the `myself-full` environment such that using `kraft run -W` works without added installation.

### Related architectures

None

### Related platforms

None

craciunoiuc:
This requires a new Dockerfile that copies all the content of `myself-full` and `base` into a new image. The other two items should be easy to do.

PR #692, opened 2023-08-01
refactor: Misc base workflow improvements

nderjung:
### Prerequisite checklist

### Description of changes

This PR improves the general workflows for "base environments" (the "base" image and the "github-actions" image) such that better caching is utilized in the former and a quality-of-life improvement for maintainers by removing the GCC build which is no longer necessary.

The biggest change is the removal of GCC which was built from source. This step is unnecessary because it provided little value-add in the end and is a product from earlier days of pykraft where manipulating the build of GCC allowed for customizing the use of certain feature flags was important for also enabling certain features in Unikraft (i.e. choosing specific languages via `--enable-language`) or certain CPU tuning (`--with-tune` which was never upstreamed).

Removing GCC reduces overall CI time since it is not built and removed the headache of building cross-compile versions for different architectures. Instead, rely on Debian to distribute a relevant version of `gcc`.


buildenvs/base.Dockerfile
```diff
@@ -2,114 +2,79 @@
...
+      make \
+      openssh-client \
+      patch \
+      python3 \
+      socat \
+      unzip \
+      uuid-runtime \
+      wget \
+      xz-utils \
+      ; \
```
craciunoiuc: Don't forget to also pin the versions so we make sure nothing changes

buildenvs/base.Dockerfile
```diff
@@ -2,114 +2,79 @@
...
+    ln -s /usr/bin/aarch64-linux-gnu-lto-tool-12            /usr/bin/aarch64-linux-gnu-lto-tool; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-cpp-12        /usr/bin/gcc-12-arm-linux-gnueabihf-cpp; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-gcc-12        /usr/bin/gcc-12-arm-linux-gnueabihf-gcc; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-gcc-ar-12     /usr/bin/gcc-12-arm-linux-gnueabihf-gcc-ar; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-gcc-nm-12     /usr/bin/gcc-12-arm-linux-gnueabihf-gcc-nm; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-gcc-ranlib-12 /usr/bin/gcc-12-arm-linux-gnueabihf-gcc-ranlib; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-gcov-12       /usr/bin/gcc-12-arm-linux-gnueabihf-gcov; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-gcov-dump-12  /usr/bin/gcc-12-arm-linux-gnueabihf-gcov-dump; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-gcov-tool-12  /usr/bin/gcc-12-arm-linux-gnueabihf-gcov-tool; \
+    ln -s /usr/bin/gcc-12-arm-linux-gnueabihf-lto-tool-12   /usr/bin/gcc-12-arm-linux-gnueabihf-lto-tool;
```
craciunoiuc, line 77: Why is this needed if we install from packages?

buildenvs/github-action.Dockerfile
```diff
@@ -3,17 +3,14 @@
...
+ARG DEBIAN_VERSION=bookworm-20230725
 ARG KRAFTKIT_VERSION=latest
 ARG QEMU_VERSION=7.2.4
 ARG REGISTRY=kraftkit.sh
-ARG UK_ARCH=x86_64
 
 FROM golang:${GO_VERSION}-bullseye AS build
 
-COPY tools/github-action /go/src/kraftkit.sh/tools/github-action
+COPY . /go/src/kraftkit.sh
```
craciunoiuc, line 13: Should we copy the whole project instead of just the tool?

buildenvs/github-action.Dockerfile
```diff
@@ -25,75 +22,14 @@ RUN set -xe; \
         --add safe.directory /go/src/kraftkit.sh/tools/github-action; \
     go build \
         -o /github-action \
-        -tags static \
-        -a \
-        -ldflags='-s -w' \
-        -ldflags '-extldflags "-static"' .;
+        -a;
```
craciunoiuc, line 25: Why not static :(

buildenvs/github-action.Dockerfile
```diff
@@ -106,10 +42,12 @@ RUN set -xe; \
       ca-certificates \
       flex \
       gawk \
+      gcc-12 \
+      gcc-12-aarch64-linux-gnu \
+      gcc-12-arm-linux-gnueabihf \
       git \
       libncursesw5 \
       libncursesw5-dev \
```
craciunoiuc, line 50: If we refactor we should set the versions also

tools/github-action/go.mod
```diff
@@ -7,19 +7,22 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.7.0
 	k8s.io/api v0.27.3
-	k8s.io/apimachinery v0.27.3
+	k8s.io/apimachinery v0.27.4
 	kraftkit.sh v0.6.4-108-g28b7051
 )
 
+replace kraftkit.sh => ../..
```
craciunoiuc, line 14: This did not work for me, why shouldn't we just set a version and bump through dependabot?

craciunoiuc (changes requested):
Not sure how ready the PR is, but here are some comments from my side

buildenvs/base.Dockerfile
nderjung: I specifically removed pinning as we want to be able to easily upgrade dependencies simply by re-running the container build. This makes it easier from a maintenance point of view.

buildenvs/base.Dockerfile
nderjung: Unfortunately all programs are suffixed with `-12`. These are simply aliases to what is installed.

buildenvs/github-action.Dockerfile
nderjung: This commit requires the whole source directory: https://github.com/unikraft/kraftkit/pull/692/commits/0315bb5ed2f38bf6040b79b45d5301021dee9768

buildenvs/github-action.Dockerfile
nderjung: The `kraft` binary is not static actually in its build, nor is it entirely possible since it is dependent on a libc in order to get access to host system information (groups, user IDs, etc.) A static build results in build-time warnings (and in fact, it crashes for me sometimes).

buildenvs/base.Dockerfile
craciunoiuc: Hmm, sure, but if something were to break, we won't know from sure what package caused it. It's fine from me, but Codacy was complaining a while ago

buildenvs/github-action.Dockerfile
nderjung: What do you mean?

buildenvs/base.Dockerfile
craciunoiuc: Don't forget to also add `curl` while at it, doesn't hurt to have it in

tools/github-action/go.mod
nderjung: This just might be temporary until we do a next release. Right now, i want to get access to the current staging.

buildenvs/base.Dockerfile
nderjung: I can understand why Codacy might complain but until there's a dependabot workflow that can specifically upgrade this, then this is the easier way to move forward.

I can add `curl`.

buildenvs/github-action.Dockerfile
craciunoiuc: image

Lies. But yes, it's not fully static. For the action I suppose it's fine as we provide the whole dockerfile anyways.

buildenvs/github-action.Dockerfile
craciunoiuc: Makes sense

buildenvs/github-action.Dockerfile
craciunoiuc: Same as above, but you said you don't want version pinning

nderjung: We don't want version pinning. This just ensures we get gcc v12.X.

craciunoiuc:
As you are touching workflows and dockerfiles, fancy doing <https://github.com/unikraft/kraftkit/issues/579> also? 😉

nderjung:
> As you are touching workflows and dockerfiles, fancy doing #579 also? 😉

It's similar but i won't address those requirements in this PR. I am just trying to get github-actions to work properly for CI contexts.

craciunoiuc (approved):
All good here. Thanks!

PR #719, opened 2023-08-09
feat!: Implement OCI Index manipulation

### LINKED ISSUE #704: Package multiple targets
### Feature request summary

This issue tracks the ability to package multiple targets of a project into a single package. This is ultimately useful in contexts where the user wishes to distribute variants of the same application with different built-time properties. For example, the same application but to be executed on different host architectures or platforms.

At present, KraftKit only supports packaging unikernels application format as an OCI Image. This implementation via the [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md#example-image-index) supports this ability inherently through the index manifest, making this implementation relatively straight forward.

To approach this task, the overlaying interface `PackageManager` must be updated in order to support passing multiple targets:

```diff
diff --git a/packmanager/manager.go b/packmanager/manager.go
index 6e2b0a9..ee0db34 100644
--- a/packmanager/manager.go
+++ b/packmanager/manager.go
@@ -26,7 +26,7 @@ type PackageManager interface {
        // components can comprise of other components, it is possible to return more
        // than one package.  It is possible to disable this and "flatten" a component
        // into a single package by setting a relevant `pack.PackOption`.
-       Pack(context.Context, component.Component, ...PackOption) ([]pack.Package, error)
+       Pack(context.Context, []component.Component, ...PackOption) ([]pack.Package, error)
 
        // Unpack turns a given package into a usable component.  Since a package can
        // compromise of a multiple components, it is possible to return multiple
```

The directory handler for the OCI package manager must also be updated to support indexes. The new structure should use the OCI Image reference (e.g. `unikraft.org/helloworld:latest`) for the index and not the manifest. Manifests should be referenceable via their digest, resulting in a structure similar to the following:

```
.
|-- configs
|   `-- sha256
|       `-- 3c423c3ffecae5ab0928650c85335d15711913857cdf8ff465b9be6ab43ef6de
|-- layers
|   `-- sha256
|       `-- 289c1b99428b9e3be7f9d92bd64c0173c02ee7721542e78138048b57ae743375
|-- manifests
|   `-- sha256
|       `-- 289c1b99428b9e3be7f9d92bd64c0173c02ee7721542e78138048b57ae743375
`-- indexes
    `-- unikraft.org
        `-- helloworld
            `-- latest.json
```

### Additional tasks

- `kraft pkg` must be updated to handle correctly filtering and passing the slice of `target.Target` to the Package Manager's `Pack` method;
- The OCI implementation must handle "subsequent calls" to the same image reference such that the index is updated with new manifests representing different architecture/platform/kconfig manifest.

craciunoiuc:
### Prerequisite checklist

### Description of changes

This PR offers the ability to work with index oci artifacts.
This translates in pushing a single artifact `helloworld:latest` for multiple platform/architecture permutations.

The outstanding issue of this PR is that containerd support breaks as it still tries to work with images instead of indexes.
Moreover, existing images in the registry might be now unusable and need to be pushed again. Single artifact applications should still have an index.

GitHub-Closes: #704

### DIFF
--- a/cmd/kraft/pkg/pkg.go
+++ b/cmd/kraft/pkg/pkg.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/mattn/go-shellwords"
@@ -23,6 +24,8 @@ import (
 	"kraftkit.sh/packmanager"
 	"kraftkit.sh/tui/processtree"
 	"kraftkit.sh/unikraft/app"
+	"kraftkit.sh/unikraft/component"
+	"kraftkit.sh/unikraft/target"
 
 	"kraftkit.sh/cmd/kraft/pkg/list"
 	"kraftkit.sh/cmd/kraft/pkg/pull"
@@ -33,19 +36,19 @@ import (
 )
 
 type Pkg struct {
-	Architecture string `local:"true" long:"arch" short:"m" usage:"Filter the creation of the package by architecture of known targets"`
-	Args         string `local:"true" long:"args" short:"a" usage:"Pass arguments that will be part of the running kernel's command line"`
-	Dbg          bool   `local:"true" long:"dbg" usage:"Package the debuggable (symbolic) kernel image instead of the stripped image"`
-	Force        bool   `local:"true" long:"force-format" usage:"Force the use of a packaging handler format"`
-	Format       string `local:"true" long:"as" short:"M" usage:"Force the packaging despite possible conflicts" default:"auto"`
-	Initrd       string `local:"true" long:"initrd" short:"i" usage:"Path to init ramdisk to bundle within the package (passing a path will automatically generate a CPIO image)"`
-	Kernel       string `local:"true" long:"kernel" short:"k" usage:"Override the path to the unikernel image"`
-	Kraftfile    string `long:"kraftfile" usage:"Set an alternative path of the Kraftfile"`
-	Name         string `local:"true" long:"name" short:"n" usage:"Specify the name of the package"`
-	Output       string `local:"true" long:"output" short:"o" usage:"Save the package at the following output"`
-	Platform     string `local:"true" long:"plat" short:"p" usage:"Filter the creation of the package by platform of known targets"`
-	Target       string `local:"true" long:"target" short:"t" usage:"Package a particular known target"`
-	WithKConfig  bool   `local:"true" long:"with-kconfig" usage:"Include the target .config"`
+	Architecture string   `local:"true" long:"arch" short:"m" usage:"Filter the creation of the package by architecture of known targets"`
+	Args         string   `local:"true" long:"args" short:"a" usage:"Pass arguments that will be part of the running kernel's command line"`
+	Dbg          bool     `local:"true" long:"dbg" usage:"Package the debuggable (symbolic) kernel image instead of the stripped image"`
+	Force        bool     `local:"true" long:"force-format" usage:"Force the use of a packaging handler format"`
+	Format       string   `local:"true" long:"as" short:"M" usage:"Force the packaging despite possible conflicts" default:"auto"`
+	Initrd       string   `local:"true" long:"initrd" short:"i" usage:"Path to init ramdisk to bundle within the package (passing a path will automatically generate a CPIO image)"`
+	Kernel       string   `local:"true" long:"kernel" short:"k" usage:"Override the path to the unikernel image"`
+	Kraftfile    string   `long:"kraftfile" usage:"Set an alternative path of the Kraftfile"`
+	Name         string   `local:"true" long:"name" short:"n" usage:"Specify the name of the package"`
+	Output       string   `local:"true" long:"output" short:"o" usage:"Save the package at the following output"`
+	Platform     string   `local:"true" long:"plat" short:"p" usage:"Filter the creation of the package by platform of known targets"`
+	Targets      []string `local:"true" long:"target" short:"t" usage:"Package a particular known target"`
+	WithKConfig  bool     `local:"true" long:"with-kconfig" usage:"Include the target .config"`
 }
 
 func New() *cobra.Command {
@@ -66,7 +69,17 @@ func New() *cobra.Command {
 		`, "`"),
 		Example: heredoc.Doc(`
 			# Package a project as an OCI archive and embed the target's KConfig.
-			$ kraft pkg --as oci --name unikraft.org/nginx:latest --with-kconfig`),
+			$ kraft pkg --as oci --name unikraft.org/nginx:latest --with-kconfig
+
+			# Package a project with a given target name
+			$ kraft pkg --as oci --name unikraft.org/nginx:latest --target my_target
+
+			# Package a project with a given platform and architecture
+			$ kraft pkg --as oci --name unikraft.org/nginx:latest --plat qemu --arch x86_64
+
+			# Package a project with multiple targets
+			$ kraft pkg --as oci --name unikraft.org/nginx:latest --target my_target --target my_other_target
+		`),
 		Annotations: map[string]string{
 			cmdfactory.AnnotationHelpGroup: "pkg",
 		},
@@ -86,7 +99,7 @@ func New() *cobra.Command {
 }
 
 func (opts *Pkg) Pre(cmd *cobra.Command, _ []string) error {
-	if (len(opts.Architecture) > 0 || len(opts.Platform) > 0) && len(opts.Target) > 0 {
+	if (len(opts.Architecture) > 0 || len(opts.Platform) > 0) && len(opts.Targets) > 0 {
 		return fmt.Errorf("the `--arch` and `--plat` options are not supported in addition to `--target`")
 	}
 
@@ -100,6 +113,22 @@ func (opts *Pkg) Pre(cmd *cobra.Command, _ []string) error {
 	opts.Platform = platform.PlatformByName(opts.Platform).String()
 
 	return nil
+}
+
+func matchOption(targets []string, option string) bool {
+	for _, target := range targets {
+		name := target
+		if strings.Contains(target, "/") {
+			name = string(platform.PlatformByName(strings.Split(target, "/")[0])) +
+				"/" +
+				strings.Split(target, "/")[1]
+		}
+		if name == option {
+			return true
+		}
+	}
+
+	return false
 }
 
 func (opts *Pkg) Run(cmd *cobra.Command, args []string) error {
@@ -134,11 +163,13 @@ func (opts *Pkg) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	var tree []*processtree.ProcessTreeItem
+	var targets []target.Target
+	var packageOpts []packmanager.PackOption
 
 	parallel := !config.G[config.KraftKit](ctx).NoParallel
 	norender := log.LoggerTypeFromString(config.G[config.KraftKit](ctx).Log.Type) != log.FANCY
 
-	// Generate a package for every matching requested target
+	// Generate a package with all matching requested targets
 	for _, targ := range project.Targets() {
 		// See: https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable
 		targ := targ
@@ -146,13 +177,17 @@ func (opts *Pkg) Run(cmd *cobra.Command, args []string) error {
 		switch true {
 		case
 			// If no arguments are supplied
-			len(opts.Target) == 0 &&
+			len(opts.Targets) == 0 &&
 				len(opts.Architecture) == 0 &&
 				len(opts.Platform) == 0,
 
 			// If the --target flag is supplied and the target name match
-			len(opts.Target) > 0 &&
-				targ.Name() == opts.Target,
+			len(opts.Targets) > 0 &&
+				matchOption(opts.Targets, targ.Name()),
+
+			// If the --target flag is supplied and the target is a plat/arch combo
+			len(opts.Targets) > 0 &&
+				matchOption(opts.Targets, target.TargetPlatArchName(targ)),
 
 			// If only the --arch flag is supplied and the target's arch matches
 			len(opts.Architecture) > 0 &&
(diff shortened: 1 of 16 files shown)


oci/index.go
```diff
@@ -0,0 +1,143 @@
...
+		return ocispec.Descriptor{}, err
+	}
+
+	// General annotations
+	index.annotations[ocispec.AnnotationRefName] = ref.Context().String()
+	index.annotations[ocispec.AnnotationRevision] = ref.Identifier()
+	index.annotations[ocispec.AnnotationCreated] = time.Now().UTC().Format(time.RFC3339)
+
+	// containerd compatibility annotations
+	index.annotations[images.AnnotationImageName] = ref.String()
```
nderjung, line 105: I wonder if this should only be set by the containerd implementation.

oci/manager.go
```diff
@@ -123,6 +130,19 @@ func (manager *ociManager) Catalog(ctx context.Context, qopts ...packmanager.Que
...
 	qversion := query.Version()
+	platform := &v1.Platform{}
+
+	if query.Architecture() != "" {
+		platform.Architecture = query.Architecture()
+	}
+
+	if query.OS() != "" {
+		platform.OS = query.OS()
+	}
```
nderjung: We retrofit platform into the OS, so we should keep our naming conventions. Please rename the `OS` attribute to `Platform` of the query.

oci/manager.go
```diff
@@ -123,6 +130,19 @@ func (manager *ociManager) Catalog(ctx context.Context, qopts ...packmanager.Que
...
+
+	if query.OSVariant() != "" {
+		platform.Variant = query.OSVariant()
+	}
```
nderjung: No need to use `OSVariant` imho. Instead `OSVersion` and `OSFeatures`. 

We should use semver to do a comparison between `OSVersion` and the requested version (this is the Unikraft version).

We retrofit KConfig options into OSFeatures. We should use `KConfig` instead and turn this into a slice representing mappings in the form`CONFIG_XYZ=y` for example.

cmd/kraft/pkg/pull/pull.go
```diff
@@ -49,7 +49,7 @@ func New() *cobra.Command {
 		Example: heredoc.Doc(`
 			# Pull the dependencies for a project in the current working directory
 			$ kraft pkg pull
-			
+
```
nderjung: Can you remove this from the commit

cmd/kraft/run/runner_package.go
```diff
@@ -77,6 +77,8 @@ func (runner *runnerPackage) Prepare(ctx context.Context, opts *Run, machine *ma
 		packmanager.WithTypes(unikraft.ComponentTypeApp),
 		packmanager.WithName(runner.packName),
 		packmanager.WithCache(true),
+		packmanager.WithOS(string(opts.platform)),
```
nderjung: `s/WithOS/WithPlatform/g`

cmd/kraft/run/runner_package.go
```diff
@@ -90,6 +92,8 @@ func (runner *runnerPackage) Prepare(ctx context.Context, opts *Run, machine *ma
 			packmanager.WithTypes(unikraft.ComponentTypeApp),
 			packmanager.WithName(runner.packName),
 			packmanager.WithCache(false),
+			packmanager.WithOS(string(opts.platform)),
```
nderjung: `s/WithOS/WithPlatform/g`

cmd/kraft/pkg/push/push.go
```diff
@@ -122,6 +122,7 @@ func (opts *Push) Run(cmd *cobra.Command, args []string) error {
 	if pm, compatible, err := pmananger.IsCompatible(ctx, ref); err == nil && compatible {
 		packages, err := pm.Catalog(ctx,
 			packmanager.WithCache(true),
+			packmanager.WithMultiple(true),
```
nderjung: The commit introducing this (66057790) should be moved later since this method does not yet exist at this point.

packmanager/query.go
```diff
@@ -31,6 +31,18 @@ type Query struct {
...
 	auths map[string]config.AuthConfig
+
+	// Multiple indicates that the query should return merged results if multiple
+	multiple bool
+
+	// Architecture specifies the architecture of the package
+	architecture string
+
+	// OS specifies the operating system of the package
+	os string
```
nderjung: `os` => `platform`

packmanager/query.go
```diff
@@ -31,6 +31,18 @@ type Query struct {
...
+
+	// OSVariant specifies the operating system variant of the package
+	osVariant string
```
nderjung: `osVariant` => `kconfig []string`

packmanager/query.go
```diff
@@ -53,6 +65,26 @@ func (query *Query) Version() string {
...
+	return query.multiple
+}
+
+// Architecture specifies the architecture of the package
+func (query *Query) Architecture() string {
+	return query.architecture
+}
+
+// OS specifies the operating system of the package
+func (query *Query) OS() string {
```
nderjung: `OS` => `Platform`

packmanager/query.go
```diff
@@ -53,6 +65,26 @@ func (query *Query) Version() string {
...
+	return query.os
+}
+
+// OSVariant specifies the operating system variant of the package
+func (query *Query) OSVariant() string {
```
nderjung: `OSVariant` => `KConfig` and `[]string`

oci/index.go
```diff
@@ -0,0 +1,143 @@
...
+	descriptor := content.NewDescriptorFromBytes(
+		ocispec.MediaTypeImageManifest,
+		manifestJson,
+	)
+
+	descriptor.Platform = &ocispec.Platform{
+		Architecture: image.Architecture,
+		OS:           image.OS,
+		Variant:      image.Variant,
+		OSFeatures:   image.OSFeatures,
```
nderjung: `Variant` can be removed.

cmd/kraft/pkg/push/push.go
craciunoiuc: I did a top-down commit order so everything you see in the history is basically reversed let's say

I think it made a bit more sense than bottom-up for reviewing

oci/index.go
craciunoiuc: well if we set it only there then pulling with containerd won't work I think, so better to set it

oci/manager.go
craciunoiuc: sure thing, but I think it won't be possible to select based on it, or at least not before pulling

oci/index.go
nderjung: This needs to be checked somehow, it's not typical to actually see these types of annotations in standard OCI images. We can open up a separate issue to track this.

oci/handler/directory.go
```diff
@@ -271,15 +311,21 @@ func (handle *DirectoryHandler) ResolveImage(ctx context.Context, fullref string
...
+	OS := strings.Split(platform, "/")[0]
+	arch := strings.Split(platform, "/")[1]
+
 	ref, err := name.ParseReference(fullref)
 	if err != nil {
 		return err
 	}
-	img, err := remote.Image(ref,
+
+	fmt.Println(platform)
```
nderjung: Lingering `fmt.Println`

oci/manager.go
```diff
@@ -123,6 +124,19 @@ func (manager *ociManager) Catalog(ctx context.Context, qopts ...packmanager.Que
...
+
+	if query.Architecture() != "" {
+		platform.Architecture = query.Architecture()
+	}
+
+	if query.Platform() != "" {
+		platform.OS = query.Platform()
+	}
+
+	if len(query.Kconfig()) > 0 {
```
nderjung: `s/Kconfig/KConfig/g` to keep it consistent with other instances in the project.

oci/pack.go
```diff
@@ -17,7 +17,6 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/name"
-	v1 "github.com/google/go-containerregistry/pkg/v1"
```
nderjung: This commit (a1967b153196a4e4966fd0440d32a77b7d2c6e06) seems to only remove this package instead of "[Adding] method to create packages from index specifications".

oci/pack.go
```diff
@@ -672,11 +718,19 @@ unpack:
 	return nil
 }
 
-// Pull implements pack.Package
+func (ociidx ociIndex) Pull(ctx context.Context, opts ...pack.PullOption) error {
+	return fmt.Errorf("can't pull index")
```
nderjung: `cannot pull OCI image index`

packmanager/query.go
```diff
@@ -31,6 +31,18 @@ type Query struct {
 
 	// Auth contains required authentication for the query.
 	auths map[string]config.AuthConfig
+
+	// Multiple indicates that the query should return merged results if multiple
+	multiple bool
```
nderjung: "merged" results? This sounds a bit confusing

cmd/kraft/pkg/pull/pull.go
nderjung: you never removed );

oci/manager.go
```diff
@@ -216,80 +236,183 @@ func (manager *ociManager) Catalog(ctx context.Context, qopts ...packmanager.Que
...
-
-	for _, manifest := range manifests {
-		// Check if the OCI image has a known annotation which identifies if a
-		// unikernel is contained within
-		if _, ok := manifest.Annotations[AnnotationKernelVersion]; !ok {
-			log.G(ctx).
-				WithField("ref", manifest.Config.Digest.String()).
-				Trace("skipping non-unikernel digest")
-			continue
+	if !query.Multiple() {
```
nderjung: Can you explain what the difference between when `query.Multiple` is `true` and `false`?

craciunoiuc: I wanted to have an implementation-agnostic name

in our case:
* false: push/pull single manifest
* true: push/pull index containing multiple manifests

packmanager/query.go
craciunoiuc: I can change

oci/pack.go
craciunoiuc: image

something really strange happened with that line in commits, can't explain but I'll pass through the history again

I lost 20 minutes on that single line trying to figure it out :D

nderjung: weird it shows this for me too but still showed up in the git commit history on github? 🤷

packmanager/query.go
nderjung: It's a larger question, see below about the diff. between true/false

oci/manager.go
nderjung: Ah i see. I guess it should then be renamed to `AllTargets` (similar to `ctr i pull --all-platforms`), for me this is a clearer name. Can you please adjust appropriately and also any comment preambles?

oci/pack.go
craciunoiuc: removed it

cmd/kraft/pkg/pkg.go
```diff
@@ -33,19 +36,19 @@ import (
...
+	Force        bool     `local:"true" long:"force-format" usage:"Force the use of a packaging handler format"`
+	Format       string   `local:"true" long:"as" short:"M" usage:"Force the packaging despite possible conflicts" default:"auto"`
+	Initrd       string   `local:"true" long:"initrd" short:"i" usage:"Path to init ramdisk to bundle within the package (passing a path will automatically generate a CPIO image)"`
+	Kernel       string   `local:"true" long:"kernel" short:"k" usage:"Override the path to the unikernel image"`
+	Kraftfile    string   `long:"kraftfile" usage:"Set an alternative path of the Kraftfile"`
+	Name         string   `local:"true" long:"name" short:"n" usage:"Specify the name of the package"`
+	Output       string   `local:"true" long:"output" short:"o" usage:"Save the package at the following output"`
+	Platform     string   `local:"true" long:"plat" short:"p" usage:"Filter the creation of the package by platform of known targets"`
+	Target       []string `local:"true" long:"target" short:"t" usage:"Package a particular known target"`
```
nderjung: Internally this attribute should be named `Targets`, the `long` and `short` can remain the same.

cmd/kraft/pkg/pkg.go
```diff
@@ -153,7 +184,11 @@ func (opts *Pkg) Run(cmd *cobra.Command, args []string) error {
 
 			// If the --target flag is supplied and the target name match
 			len(opts.Target) > 0 &&
-				targ.Name() == opts.Target,
+				matchOption(opts.Target, targ.Name()),
+
+			// If the --target flag is supplied and the target is a plat/arch combo
+			len(opts.Target) > 0 &&
+				matchOption(opts.Target, targ.Platform().Name()+"/"+targ.Architecture().Name()),
```
nderjung:
```diff
- matchOption(opts.Target, targ.Platform().Name()+"/"+targ.Architecture().Name()),
+ matchOption(opts.Target, target.TargetPlatArchName(targ)),
```

machine/qemu/init.go
```diff
@@ -267,7 +267,7 @@ func init() {
 	// gob.Register(QemuDevicePcTestdev{})
 	// gob.Register(QemuDevicePciTestdev{})
 	// gob.Register(QemuDevicePcm3680Pci{})
-	// gob.Register(QemuDevicePvpanic{})
+	gob.Register(QemuDevicePvpanic{})
```
nderjung: Lingering from another PR?

machine/qemu/v1alpha1.go
```diff
@@ -240,6 +240,7 @@ func (service *machineV1alpha1Service) Create(ctx context.Context, machine *mach
 		}),
 		WithDisplay(QemuDisplayNone{}),
 		WithParallel(QemuHostCharDevNone{}),
+		WithDevice(QemuDevicePvpanic{}),
```
nderjung: Here as well.

machine/qemu/v1alpha1.go
```diff
@@ -675,6 +676,13 @@ func (service *machineV1alpha1Service) Watch(ctx context.Context, machine *machi
 				machine.Status.State = machinev1alpha1.MachineStateExited
 				events <- machine
 
+				if !qcfg.NoShutdown {
+					break accept
+				}
+			case qmpapi.EVENT_GUEST_PANICKED:
+				machine.Status.State = machinev1alpha1.MachineStateErrored
+				events <- machine
+
```
nderjung: Also

machine/qemu/v1alpha1.go
```diff
@@ -841,7 +849,11 @@ func (service *machineV1alpha1Service) Get(ctx context.Context, machine *machine
 
 	// Map the QMP status to supported machine states
 	switch status.Return.Status {
-	case qmpapi.RUN_STATE_GUEST_PANICKED, qmpapi.RUN_STATE_INTERNAL_ERROR, qmpapi.RUN_STATE_IO_ERROR:
+	case qmpapi.RUN_STATE_GUEST_PANICKED:
+		state = machinev1alpha1.MachineStateErrored
+		exitCode = 1
+
+	case qmpapi.RUN_STATE_INTERNAL_ERROR, qmpapi.RUN_STATE_IO_ERROR:
```
nderjung: And also here

oci/handler/directory.go
```diff
@@ -323,88 +426,154 @@ func (handle *DirectoryHandler) FetchImage(ctx context.Context, fullref, platfor
...
+	var manifests []v1.Descriptor
+	for _, descriptor := range parsableManifest.Manifests {
+		if descriptor.Platform.OS == OS && descriptor.Platform.Architecture == arch {
+			log.G(ctx).Tracef("Fetching %s/%s manifest %s",
+				descriptor.Platform.Architecture,
+				descriptor.Platform.OS,
+				descriptor.Digest)
+			manifests = append(manifests, descriptor)
+		} else {
+			log.G(ctx).Tracef("Skip fetching %s/%s manifest %s",
```
nderjung: Log messages should start with a lowercase letter. Let's also use `WithFields` too, for example:
```go
log.G(ctx).
  WithField("platform", fmt.Sprintf("%s/%s", descriptor.Platform.OS, descriptor.Platform.Architecture)).
  WithField("digest", descriptor.Digest).
  Tracef("fetching")
```

oci/handler/directory.go
```diff
@@ -441,72 +610,104 @@ func (handle *DirectoryHandler) PushImage(ctx context.Context, fullref string, t
 }
 
 // UnpackImage implements ImageUnpacker.
-func (handle *DirectoryHandler) UnpackImage(ctx context.Context, ref string, dest string) (err error) {
-	img, err := handle.ResolveImage(ctx, ref)
+func (handle *DirectoryHandler) UnpackImage(ctx context.Context, ref string, platform string, dest string) (err error) {
+	OS := strings.Split(platform, "/")[0]
+	arch := strings.Split(platform, "/")[1]
```
nderjung: lowercase letters for scoped variables, `s/OS/os/g`

machine/qemu/init.go
craciunoiuc: how did this happen. I'm never reordering commits again, so strange

craciunoiuc:
This was superseded

PR #736, opened 2023-08-22
fix(unikraft): Hardcode language libraries before libcs

craciunoiuc:
### Prerequisite checklist

### Description of changes

This commit hardcodes all known language libraries that need to sit before musl/newlib.
This is a temporary fix as mentioned in the comment above them.
Before that is fixed, all libraries known to cause problems should be added here.

### DIFF
--- a/unikraft/app/application.go
+++ b/unikraft/app/application.go
@@ -347,6 +347,29 @@ func (app application) MakeArgs(tc target.Target) (*core.MakeArgs, error) {
 		unformattedLibraries[k] = v
 	}
 
+	// Add language libraries that we know require a specific ordering.
+	// Currently, these are C++-related libraries. Others may be added as required.
+	if unformattedLibraries["libcxxabi"] != nil {
+		libraries = append(libraries, unformattedLibraries["libcxxabi"].Path())
+		delete(unformattedLibraries, "libcxxabi")
+	}
+	if unformattedLibraries["libcxx"] != nil {
+		libraries = append(libraries, unformattedLibraries["libcxx"].Path())
+		delete(unformattedLibraries, "libcxx")
+	}
+	if unformattedLibraries["libunwind"] != nil {
+		libraries = append(libraries, unformattedLibraries["libunwind"].Path())
+		delete(unformattedLibraries, "libunwind")
+	}
+	if unformattedLibraries["compiler-rt"] != nil {
+		libraries = append(libraries, unformattedLibraries["compiler-rt"].Path())
+		delete(unformattedLibraries, "compiler-rt")
+	}
+	if unformattedLibraries["libgo"] != nil {
+		libraries = append(libraries, unformattedLibraries["libgo"].Path())
+		delete(unformattedLibraries, "libgo")
+	}
+
 	// All supported libCs right now
 	if unformattedLibraries["musl"] != nil {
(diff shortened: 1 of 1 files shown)


unikraft/app/application.go
```diff
@@ -347,6 +347,24 @@ func (app application) MakeArgs(tc target.Target) (*core.MakeArgs, error) {
 		unformattedLibraries[k] = v
 	}
 
+	// All known language libraries right now
```
razvand: This comment seems a bit cryptic to me. What does it mean?

razvand (commented):
Left a small comment.

unikraft/app/application.go
craciunoiuc: Basically, all libraries that are needed to compile different languages usually sit in front of libc.

`known` -> all that I know of 😅

razvand: I would suggest a fuller comment such as "Add language libraries that we know require a specific ordering. Currently, these are C++-related libraries. Other may be added as required."

StefanJum: `usually sit` sounds very scary. Basically, if we do this there is no (easy) way to change the libraries ordering (placing a libc in front of a language library), right?

Is there a reason to change the library ordering from the `kraftkit` internals? Why not just go with the ordering from the `kraft.yaml` file? That sounds to me like the easiest way to go, leave some control to the user, do not hardcode library ordering from within kraftkit.

unikraft/app/application.go
```diff
@@ -347,6 +347,25 @@ func (app application) MakeArgs(tc target.Target) (*core.MakeArgs, error) {
 		unformattedLibraries[k] = v
 	}
 
+	// Add language libraries that we know require a specific ordering.
+	// Currently, these are C++-related libraries. Other may be added as required
```
razvand:
```suggestion
	// Currently, these are C++-related libraries. Others may be added as required.
```

unikraft/app/application.go
razvand: @StefanJum , as in the PR comment: " This is a temporary fix as mentioned in the comment above them."

@craciunoiuc , I would add some "FIXME" or "TODO" item in front of the comment, together with the sentence "This is a temporary fix", as something to be handled properly in the future.

craciunoiuc: there is a previous todo in there above
image

craciunoiuc: @StefanJum because the libraries in the `kraft.yaml` are saved in a map so the order is not preserved

craciunoiuc: Alex wants us to parse the dependency chain in unikraft and detect the order from there

razvand (approved):
Thanks, @craciunoiuc. This is OK with me.

Reviewed-by: Razvan Deaconescu <razvand@unikraft.io>

StefanJum:
Should we also add [`libgo`](https://github.com/unikraft/lib-libgo) to the list (it's also put [before the libc](https://github.com/unikraft/app-helloworld-go/blob/staging/Makefile#L4))?

craciunoiuc:
sure thing

craciunoiuc:
added

StefanJum (approved):
All good now, thanks.
Reviewed-by: Stefan Jumarea <stefanjumarea02@gmail.com>

PR #740, opened 2023-08-23
fix(run): Fix race-condition related to short-lived instances

craciunoiuc:
### Prerequisite checklist

### Description of changes

This PR switches the current event-based logging to a hybrid poll-event system.
The main takeaway is that every 10 ms kraftkit checks if the machine has exited and wants to shutdown.
When that is the case, it will check for two subsequent calls where nothing was printed.
The only case when the machine wants to shutdown (nothing more will be outputted) and nothing was printed for 20ms is when it really finished.

The only case where output would be missed is when the console used for ouput writes less than 1 character per 10ms.

This poll method infers ~0.4% stress on a single core, which means 0.1% on a 4-core CPU. (4 GHz)

### DIFF
--- a/cmd/kraft/run/run.go
+++ b/cmd/kraft/run/run.go
@@ -7,10 +7,12 @@ package run
 import (
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/MakeNowJust/heredoc"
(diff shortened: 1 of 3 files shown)


marcrittinghaus (approved):
Reviewed-by: Marc Rittinghaus <marc.rittinghaus@unikraft.io>
Approved-by: Marc Rittinghaus <marc.rittinghaus@unikraft.io>

PR #751, opened 2023-08-28
fix(initrd): Do not follow symlinks when creating archives

craciunoiuc:
### Prerequisite checklist

### Description of changes

Creating an initrd through kraftkit currently fails if it contains symlinks

`symlink -> dir`

If the symlink points to a directory it will consider it as a directory but then fail because it is actually a file
The error is something like `not a directory`.

To get the error just unpack the rootfs in `app-python3` and try to use it with `--initrd ./fs0`

### DIFF
--- a/initrd/initrd.go
+++ b/initrd/initrd.go
@@ -197,15 +197,26 @@ func NewFromMapping(workdir, output string, maps ...string) (*InitrdConfig, erro
 
 				initrd.Files[path] = internal
 
-				data, err := os.ReadFile(path)
-				if err != nil {
-					return err
+				var data []byte
+				targetLink := ""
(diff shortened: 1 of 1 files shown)


nderjung:
Hi! Could you please provide a reason for this change? I.e. what issue/error happened to prompt this PR and why disabling symbolic links solves it. Thanks!

craciunoiuc:
@nderjung 

Creating an initrd through kraftkit currently fails if it contains symlinks

`symlink -> dir`

If the symlink points to a directory it will consider it as a directory but then fail because it is actually a file
The error is something like `not a directory`.

To get the error just unpack the rootfs in `app-python3` and try to use it with `--initrd ./fs0`

nderjung (approved):
Thanks!

PR #752, opened 2023-08-29
fix(kconfig): Correctly handle integer conversions from map interface

nderjung:
### Prerequisite checklist

### Description of changes

When the `NewKeyValueMapFromMap` was used with integers, this would result in a Golang value of `(int=5)%s` for example.  Correctly handle by doing a type check.

### DIFF
--- a/kconfig/config.go
+++ b/kconfig/config.go
@@ -24,7 +24,13 @@ func NewKeyValueMapFromSlice(values ...interface{}) KeyValueMap {
 	mapping := KeyValueMap{}
 
 	for _, value := range values {
-		str := fmt.Sprintf("%s", value)
+		var str string
+		switch t := value.(type) {
+		case string:
+			str = t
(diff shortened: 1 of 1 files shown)


kconfig/config.go
```diff
@@ -24,7 +24,13 @@ func NewKeyValueMapFromSlice(values ...interface{}) KeyValueMap {
 	mapping := KeyValueMap{}
 
 	for _, value := range values {
-		str := fmt.Sprintf("%s", value)
+		var str string
+		switch t := value.(type) {
+		case string:
+			str = t
+		case int, uint, int64, uint64:
```
jake-ciolek: It might be good to cover all the remaining integer types - (u)int(8|16|32).

jake-ciolek (approved):

kconfig/config.go
nderjung: Good shout! Thanks, I've updated the PR. :)

PR #738, opened 2023-08-23
feat(login): Add option to login to container registries

craciunoiuc:
### Prerequisite checklist

  - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
  - [x] Tested your changes against relevant architectures and platforms;
  - [x] Ran `make fmt` on your commit series before opening this PR;
  - [ ] Updated relevant documentation.

### Description of changes

This PR adds the posibility to login directly to docker through kraftkit when using `--registry`
This remove the usage of workarounds to get push to works in kraftkit.

This is a workaround in itself and the correct method to do this would be to pass the saved login data to the auth method.

PR #759, opened 2023-08-29
fix: Handle registry authentication

nderjung:
### Prerequisite checklist

### Description of changes

This PR essentially reverts #738 for two reasons as well as correctly handling authentication in the context of GitHub Actions:
    
- We should not be manipulating the `~/.docker/config.json` file as it is not owned by KraftKit.  Modifying this file can be seen as malicious or unintended behaviour from users.
- Ultimately, this does not solve the underlying issue of communicating with registries in a pragmatic way: KraftKit already has an internal "auths" section which can be referenced and reading the docker config file should simply be properly supplemented in the existing implementation which looks up user credentials.

Included in this PR is a fix which makes it possible to use the typical action, [`docker/action-login`](https://github.com/docker/login-action), for logging into a container registry, making the KraftKit action a drop-in replacement to existing pipelines.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -77,5 +77,30 @@ inputs:
     default: false
 
 runs:
-  using: docker
-  image: docker://kraftkit.sh/github-action:v0
+  # Use a composite type so we can execute a bash script before the invocation
+  # of the container image.
+  using: composite
+  steps:
+  - shell: bash
+    run: ${{ github.action_path }}/tools/github-action/pre.sh
+  - uses: docker://kraftkit.sh/github-action:v0
+    # This is a bit of a hack, since we are using a composite-type, the actions
+    # input are not passed when the container image is instantiated.  Instead,
+    # we simply pass all inputs as environmental variables which are already
+    # compatible with the action itself.
+    env:
+      INPUT_LOGLEVEL: ${{ inputs.loglevel }}
+      INPUT_WORKDIR: ${{ inputs.workdir }}
+      INPUT_KRAFTFILE: ${{ inputs.kraftfile }}
+      INPUT_ARCH: ${{ inputs.arch }}
+      INPUT_PLAT: ${{ inputs.plat }}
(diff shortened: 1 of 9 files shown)


oci/handler/directory.go
```diff
@@ -410,21 +414,38 @@ func (handle *DirectoryHandler) FetchImage(ctx context.Context, fullref, platfor
 	return nil
 }
 
+// directoryHandlerAuthorization is used handle looking up the already populated
```
craciunoiuc: `handle` -> `to handle`

I can't stop correcting typos

tools/github-action/pre.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+RUNNER_HOME=/home/runner/work/_temp/_github_home
```
craciunoiuc, line 3: we need to keep an eye out whether this ever changes

action.yml
```diff
@@ -77,5 +77,30 @@ inputs:
...
 
 runs:
-  using: docker
-  image: docker://kraftkit.sh/github-action:v0
+  # Use a composite type so we can execute a bash script before the invocation
+  # of the container image.
+  using: composite
+  steps:
+  - shell: bash
+    run: ${{ github.action_path }}/tools/github-action/pre.sh
```
craciunoiuc, line 85: In commit message 3:

` the pwth that` ->` the path that`

craciunoiuc (changes requested):
Look good, just 2 typos (sorry :( )

tools/github-action/pre.sh
nderjung: This is true, but there's nothing we can do until it does change.

oci/handler/directory.go
nderjung: I need to install a spell checker, thanks!

craciunoiuc (approved):
All good here. Thanks!

PR #770, opened 2023-09-04
fix(config): Remove several flags from the default config

craciunoiuc:
### Prerequisite checklist

### Description of changes

Out of the two commits, the one removing the config from the default `config.yaml` file is crucial.
As the path is already known, this flag is partially redundant.
Because this flag can point to a different config, it can make the config go into an inconsistent state, which leads to the flags being loaded from the current config and sub-flags from the one pointed to, whilst also ignoring env and parameter flags.


nderjung (approved):
Thanks!

PR #782, opened 2023-09-06
fix(machine): Switch to IEC format memory sizes

### LINKED ISSUE #773: Incorrect computation of guest memory
### Describe the bug

`kraft run` incorrectly interprets memory suffixes `G` and `T` to be `1000M` and `1000000M`, respectively.
The correct values are `1024M` and `1048576M`, respectively.
A similar issue exists for larger suffixes, but those are outside the realm of feasible RAM sizes today.
This is due to a confusion between SI and IEC units, see below for a detailed discussion.

### Steps to reproduce

`kraft run --plat qemu --log-level debug -M 1G` on any unikernel image and observe the `-m size=XXX` argument in the resulting qemu commandline.

### Expected behavior

Expected `-m size=1024M` for `-M 1G` and `-m size=1048576M` for `-M 1T`.

#### TL;DR
Memory works best handled with IEC (base-2) suffixes due to how it's built and accessed at the hardware level. Using SI units will lead to memory fragmentation and inability to use cool performance features like hugepages.

#### Long explanation (grab a coffee):
Historically there has been much confusion about how to correctly express multiples of bits and bytes. Early computer engineers inconsistently piggy-backed on top of metric prefixes: sometimes 'kilo' meant 1000 (base-10; 10^3), sometimes 1024 (base-2, 2^10), whichever was most convenient for the task at hand -- some things are naturally base-2, some not. The marketing department probably got involved at some point too.

This confusion only got exponentially worse when larger multipliers appeared ('mega', 'giga', and so on), with absolutely no standard as to what each meant. 1M could mean 1000 * 1000 (e.g. 1 Mbps in networking), or 1024 * 1000 (e.g., 1.44MB floppies), or 1024 * 1024 (a 1MB file as declared by Windows). This became progressively less sustainable, as the relative difference between interpretations grew ever larger. Misinterpreting 1K is only a 2.4% error; for 1T can be as much as 10% -- enough for non-techies to notice and vocally complain, possibly to lawyers.

To solve this issue once and for all, two systems are agreed upon for measuring multiples of bytes:
- SI (metric): kB, MB, TB, PB, EB, etc., representing powers of 1000
- IEC: KiB, MiB, TiB, PiB, EiB, etc., representing powers of 1024

Note that case is important here! There's no such thing as KB; it's either kB or KiB (just like we have kg and km, not Kg and Km).

So why use one over the other? Common sense says to stick to what's most widely known: metric prefixes, and indeed most user-facing software has made that transition for storage media (with the notable exception of Windows shell, likely for obscure backwards-compat reasons). This makes sense as the size of a storage medium is not directly determined by powers of two. Networking too uses SI prefixes since forever, because telecommunications often don't care about powers of two either.

Where things change greatly is RAM -- the one place where IEC units are not only preferrable, but often mandated by hardware design and performance considerations. Without getting _too deep_ into details, DRAM, even though it appears as one contiguous byte-addressable array (mostly), is actually addressed using a higher-dimensional tuple of addresses: `<channel, DIMM, rank, bank, row, column>`, with such a tuple addressing a _memory word_ (which in DDR* is 64 bits).

The smallest usable unit of DRAM memory is the _bank_ which consists of _exactly_ 65536 rows by 1024 columns of 64-bit memory words. JEDEC DDR standards further impose the number of banks a DRAM device (or _rank_) must provide -- DDR3 has exactly 8 banks, DDR4 16. All DRAM of a particular standard has the _exact_ same geometry. The way you get more DRAM is by pooling multiple devices in parallel, then MUXing between the lot: two ranks on the same DIMM with a rank select bit, two DIMMs on the same channel w/ a DIMM select bit, two channels on the same controller addressed independently, etc.
This MUXing operation further entrenches powers of two as the natural addressing mode of DRAM -- indeed most RAM configurations in the wild are pure powers of two for this exact reason, and the rare examples that aren't must perform various addressing tricks inside the memory controller in order to cleanly map physical DRAM into a linear address space.
AMD used to publish their memory controller mapping functions for microarchitectures prior to zen; feel free to look them up for a concrete example of this.

Besides all that we must consider how RAM is typically accessed in a modern system: paging. Because page translation of a virtual address must be done as fast as possible in hardware, pages too are naturally aligned on power-of-two boundaries so that bit-slicing may be used instead of more complex arithmetic. This goes doubly so for hugepages.
Allocating without regard to natural page boundaries can lead to wasted unaligned memory, or can prevent the use of more performant hugepages altogether.

Summing all up: DRAM is _aggressively_ parallelized as well as addressed in powers of two, and conscientious allocation of memory resources should also follow powers of two if one wants to make the most of memory performance.

_No response_

_No response_

### Relevant log output

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

This PR fixes several other SI -> IEC cases

GitHub-Fixes: #773

### DIFF
--- a/machine/firecracker/v1alpha1.go
+++ b/machine/firecracker/v1alpha1.go
@@ -36,8 +36,9 @@ import (
 )
 
 const (
-	FirecrackerBin      = "firecracker"
-	DefaultClientTimout = time.Second * 5
+	FirecrackerBin         = "firecracker"
+	DefaultClientTimout    = time.Second * 5
+	FirecrackerMemoryScale = 1024 * 1024
 )
 
 // machineV1alpha1Service ...
@@ -103,7 +104,7 @@ func (service *machineV1alpha1Service) Create(ctx context.Context, machine *mach
 	}
 
 	if machine.Spec.Resources.Requests.Memory().Value() == 0 {
-		quantity, err := resource.ParseQuantity("64M")
+		quantity, err := resource.ParseQuantity("64Mi")
 		if err != nil {
 			machine.Status.State = machinev1alpha1.MachineStateFailed
 			return machine, err
@@ -278,7 +279,7 @@ watch:
 	// Set the machine's resource configuration.
 	if _, err := client.PutMachineConfiguration(ctx, &models.MachineConfiguration{
 		VcpuCount:  firecracker.Int64(machine.Spec.Resources.Requests.Cpu().Value()),
-		MemSizeMib: firecracker.Int64(machine.Spec.Resources.Requests.Memory().Value() / 1000000),
+		MemSizeMib: firecracker.Int64(machine.Spec.Resources.Requests.Memory().Value() / FirecrackerMemoryScale),
 	}); err != nil {
 		return machine, err
 	}
--- a/machine/qemu/qemu_memory.go
+++ b/machine/qemu/qemu_memory.go
@@ -25,6 +25,7 @@ type QemuMemory struct {
 
 const (
 	QemuMemoryDefault = 64
+	QemuMemoryScale   = 1024 * 1024
 )
 
 func (qm QemuMemory) String() string {
--- a/machine/qemu/v1alpha1.go
+++ b/machine/qemu/v1alpha1.go
@@ -153,7 +153,7 @@ func (service *machineV1alpha1Service) Create(ctx context.Context, machine *mach
 	}
 
 	if machine.Spec.Resources.Requests.Memory().Value() == 0 {
-		quantity, err := resource.ParseQuantity("64M")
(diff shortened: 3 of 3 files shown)


jake-ciolek:
Thanks!

I tried it out and, qemu side, 1G results in:

` -m size=1000M `
 
 but 1Gi ends up being:

`-m size=1073M`
 
 I think for 1Gi it should result in:
 
 `-m size=1024M`

craciunoiuc:
that's strange, I'll have to do more tests tomorrow

All good, confusion cleared up

jake-ciolek:
I used the wrong branch, 1Gi is 1024M and 1G results in 953M

jake-ciolek (approved):

PR #800, opened 2023-09-13
feat(workflows): Automate stable release tags

craciunoiuc:
### Prerequisite checklist

### Description of changes

Generating a new tag will also start the release pipeline.
Monday was chosen as it's 1 day after dependabot runs and is before the week starts.

### DIFF
--- a/.github/workflows/release-stable.yaml
+++ b/.github/workflows/release-stable.yaml
@@ -2,7 +2,7 @@ name: release-stable
 
 on:
   push:
-    tags: ["v[1-9]+.[0-9]+.[0-9]+"]
+    tags: ["v[0-9]+.[0-9]+.[0-9]+"]
     branches: [stable]
 
 permissions:
--- a/.github/workflows/release-stable.yaml
+++ b/.github/workflows/release-stable.yaml
@@ -3,7 +3,6 @@ name: release-stable
 on:
   push:
     tags: ["v[0-9]+.[0-9]+.[0-9]+"]
-    branches: [stable]
 
 permissions:
   contents: write
--- a/.github/workflows/tag-stable-weekly.yaml
+++ b/.github/workflows/tag-stable-weekly.yaml
@@ -1,0 +1,53 @@
+name: tag-stable-weekly
+
+on:
+  schedule:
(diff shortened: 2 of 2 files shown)


.github/workflows/release-stable.yaml
```diff
@@ -2,7 +2,7 @@ name: release-stable
 
 on:
   push:
-    tags: ["v[1-9]+.[0-9]+.[0-9]+"]
+    tags: ["v[0-9]+.[0-9]+.[0-9]+"]
```
nderjung: Weird, how did this work before? :D

craciunoiuc: It did not 🪄 

I had to do it by hand 😢

craciunoiuc: either that or we pushed to stable and that picked it up

What I was doing was to generate a tag + also push to stable and that would pick it up

nderjung (commented):
I wonder, this should actually also check if there is actually a diff between staging/HEAD and stable. If we have a holiday or there is a week where nothing changes, a new version will come out regardless. 

The pipeline should also merge staging into stable, this is part of the general release mechanism.

craciunoiuc:
I can add some more checks, sure
Things left to do:
* Checkout full staging branch
* Check if commit difference between latest release and the new one is bigger than 0
* Push to stable branch
* Disable releases on stable push

Edit: Fixed all here

.github/workflows/tag-stable-weekly.yaml
```diff
@@ -0,0 +1,53 @@
...
+          NEW_TAG="v${array[0]}.${array[1]}.${array[2]}"
+
+          git checkout staging
+          git tag -a "$NEW_TAG" -m "KraftKit $NEW_TAG"
+          git push origin "$NEW_TAG"
+
+          # Rebase stable on staging
+          git checkout stable
+          git rebase staging
+          git push origin stable
```
nderjung: can you reverse the order here: merge and push to stable, then tag

craciunoiuc: oke

nderjung (approved):
Thanks!

PR #801, opened 2023-09-13
fix(workflows): Skip tidy checks if dependabot bumps

craciunoiuc:
### Prerequisite checklist

### Description of changes

Most, if not all, dependency bumps are now done by dependabot. They are single-commit PRs and have fail because of the `replace` statement inside `github-action`.

Because of this, it's safe to just remove the check for all gomod cases.


craciunoiuc:
Not a good fix, it would break afterwards

PR #826, opened 2023-09-18
feat(e2e): Add tests for the 'net' set of commands

craciunoiuc:
### Prerequisite checklist

### Description of changes

To run the tests you will need to be root or run with `sudo` because the commands manipulate network interfaces.

Hopefully this works in the actions, otherwise we can just try to skip them in the action.

### DIFF
--- a/test/e2e/cli/net_down_test.go
+++ b/test/e2e/cli/net_down_test.go
@@ -27,7 +27,7 @@ var _ = Describe("kraft net down", func() {
 
 		cfg = fcfg.NewTempConfig()
 
-		cmd = fcmd.NewKraft(stdout, stderr, cfg.Path())
+		cmd = fcmd.NewKraftPrivileged(stdout, stderr, cfg.Path())
 		cmd.Args = append(cmd.Args, "net", "down", "--log-level", "info", "--log-type", "json")
 	})
 
@@ -63,7 +63,7 @@ var _ = Describe("kraft net down", func() {
 			// Create the network
 			stdoutCreate := fcmd.NewIOStream()
 			stderrCreate := fcmd.NewIOStream()
-			cmdCreate := fcmd.NewKraft(stdoutCreate, stderrCreate, cfg.Path())
+			cmdCreate := fcmd.NewKraftPrivileged(stdoutCreate, stderrCreate, cfg.Path())
 			cmdCreate.Args = append(cmdCreate.Args, "net", "create", "--log-level", "info", "--log-type", "json")
 			cmdCreate.Args = append(cmdCreate.Args, "--driver", "bridge")
 			cmdCreate.Args = append(cmdCreate.Args, "--network", "172.48.0.1/24")
@@ -81,7 +81,7 @@ var _ = Describe("kraft net down", func() {
 		AfterEach(func() {
 			stdoutRm := fcmd.NewIOStream()
 			stderrRm := fcmd.NewIOStream()
-			cmdRm := fcmd.NewKraft(stdoutRm, stderrRm, cfg.Path())
+			cmdRm := fcmd.NewKraftPrivileged(stdoutRm, stderrRm, cfg.Path())
 			cmdRm.Args = append(cmdRm.Args, "net", "rm", "test-down-1")
 
 			err := cmdRm.Run()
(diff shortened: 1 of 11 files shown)


test/e2e/cli/net_rm_test.go
```diff
@@ -0,0 +1,151 @@
...
+			err := cmd.Run()
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(stderr.String()).To(BeEmpty())
+			Expect(stdout.String()).To(MatchRegexp(`^Remove a network\n`))
+		})
+	})
+
+	// Requires root privileges
+	When("invoked with the --help flag", func() {
```
antoineco: Duplicate spec

test/e2e/cli/net_rm_test.go
```diff
@@ -0,0 +1,151 @@
...
+			cmd.Args = append(cmd.Args, "--driver", "bridge")
+			cmd.Args = append(cmd.Args, "--network", "172.46.0.1/24")
+			cmd.Args = append(cmd.Args, "test-rm-0")
+
+			err := cmd.Run()
+
+			Expect(err).ToNot(HaveOccurred())
+			Expect(stderr.String()).To(BeEmpty())
+
+			// Reset args
```
antoineco: (Applies all tests suites)
This same command is set in the global `BeforeEach`, so instead of overriding it, then reset it, couldn't we simply use another var for the other setup commands and avoid those reset steps?

test/e2e/cli/net_up_test.go
```diff
@@ -0,0 +1,156 @@
...
+		AfterEach(func() {
+			stdout = fcmd.NewIOStream()
+			stderr = fcmd.NewIOStream()
+			cmd = fcmd.NewKraftPrivileged(stdout, stderr, cfg.Path())
+			cmd.Args = append(cmd.Args, "net", "rm", "test-up-1")
+			err := cmd.Run()
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should bring the network up and print it's name", func() {
```
antoineco: its

test/e2e/cli/net_down_test.go
```diff
@@ -0,0 +1,129 @@
...
+		AfterEach(func() {
+			stdout = fcmd.NewIOStream()
+			stderr = fcmd.NewIOStream()
+			cmd = fcmd.NewKraftPrivileged(stdout, stderr, cfg.Path())
+			cmd.Args = append(cmd.Args, "net", "rm", "test-down-1")
+			err := cmd.Run()
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should bring the network down and print it's name", func() {
```
antoineco: its

antoineco (commented):
General comments:

- All these commands are intertwined inside the test suites. `inspect` depends on `up`, `down` and `rm` for example. It seems like we could have easily covered most of them in a single suite.

- It would have been interesting to test that a unikernel can actually make use of the network created by those commands (e.g. nginx from our catalog).

test/e2e/cli/net_rm_test.go
craciunoiuc: Makes sense, will do for all

craciunoiuc:
Yes, for the first point, I wanted to use `ifconfig` to do what kraft does but then I realised that it's not really a good, generic way to test. So, better to just throw all commands in there and see where it breaks.

For the second point, it will surely be covered in the `kraft run` tests. Here we can at most add a ping to the bridge to see if it's up

antoineco:
Regarding the first point, what I meant is that you could pretty much keep only the file that's currently called `net_inspect`, and still cover almost all relevant functionalities.
If the goal is to test a lot of combinations of flags I agree that we need a tests suite for every subcommand. However I would focus only on the edge cases in these, to avoid the redundancy of the "happy path".

This is not a request for change btw, just an observation to reduce redundancy in the future. The code is written so let's keep it.

craciunoiuc:
Took a bit more to solve all stuff

I now suffixed every set of commands with the command name. Should make things a bit less confusing + less chance for errors because input/output won't (potentially) spill between commands

craciunoiuc:
aaand tests fail now online

craciunoiuc:
image

Even version fails, strange

craciunoiuc:
and now they seem to pass, wut

antoineco (approved):

PR #827, opened 2023-09-19
feat(goreleaser): Suggest compiler packages for different arch cases

craciunoiuc:
### Prerequisite checklist

### Description of changes

If someone with arm64 installs the kraftkit package they will only get the arm build tools through the recommended packages.
The main, most stable, target of Unikraft is still x86 so we add suggestions to explicitly install the toolchains for building x86_64 targets.

### DIFF
--- a/.goreleaser-stable.yaml
+++ b/.goreleaser-stable.yaml
@@ -58,6 +58,9 @@ nfpms:
       - socat
       - unzip
       - wget
+    suggests:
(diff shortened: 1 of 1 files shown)


craciunoiuc:
It seems that to use this people need to have the `--install-suggests` added. Also the install script needs to be changed.

I can move this to recommends if you want

antoineco (approved):

PR #830, opened 2023-09-19
fix(goreleaser): Enable multi-binary releases

craciunoiuc:
### Prerequisite checklist

### Description of changes

This PR fixes 3 aspects:
1. Arm64 does not work for runu so we disable for now (it would require including the gcc toolchain in the workflow)
1. Publishers had different ids from the nfpms so it lead to it not doing anything
1. We now have multiple binaries so we need to specify the archives for each one and what to do with them

### DIFF
--- a/.goreleaser-stable.yaml
+++ b/.goreleaser-stable.yaml
@@ -39,6 +39,7 @@ release:
 
 nfpms:
   - vendor: Unikraft
+    id: nfpm-default
     maintainer: Alexander Jung <alex@unikraft.io>
     description: Build and use highly customized and ultra-lightweight unikernels.
     license: BSD 3-clause
--- a/.goreleaser-stable.yaml
+++ b/.goreleaser-stable.yaml
@@ -120,14 +120,15 @@ builds:
 #@ end
 #@ end
 #@ targets = {
-#@   "linux-amd64": {"os": "linux", "arch": "amd64"},
-#@   "linux-arm64": {"os": "linux", "arch": "arm64"}
+#@   "linux-amd64": {"os": "linux", "arch": "amd64"}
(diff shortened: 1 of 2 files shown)


.goreleaser-stable.yaml
```diff
@@ -66,6 +67,16 @@ publishers:
     cmd: curl -F package=@{{ .ArtifactName }} https://{{ .Env.FURY_TOKEN }}@push.fury.io/{{ .Env.FURY_USER }}/
 aurs:
   - homepage: https://kraftkit.sh
+    ids:
+#@ targets = {
+#@   "linux-amd64": {"os": "linux", "arch": "amd64"},
+#@   "linux-arm64": {"os": "linux", "arch": "arm64"}
+#@ }
+#@ for binary in ["kraft"]:
+#@ for target, specs in targets.items():
```
nderjung: `specs` looks like it is not used? the list of targets could be simplified to just:
```
#@ targets = [
#@   "linux-amd64",
#@   "linux-arm64",
#@ ]
```
?

.goreleaser-stable.yaml
```diff
@@ -82,6 +93,16 @@ aurs:
       email: monkey+aur@unikraft.io
 nix:
   - name: kraftkit
+    ids:
+#@ targets = {
+#@   "linux-amd64": {"os": "linux", "arch": "amd64"},
+#@   "linux-arm64": {"os": "linux", "arch": "arm64"}
+#@ }
+#@ for binary in ["kraft"]:
+#@ for target, specs in targets.items():
```
nderjung: same here with "specs"

antoineco (approved):

Issue #820, opened 2023-09-16
`kraft rm --all` ends up with an error


razvand:
### Describe the bug

After doing KraftKit testing, `kraft rm --all` end up with an error.

### Steps to reproduce

I used KraftKit and, then, at some point I wanted to remove the created VMs. I used a normal account to remove those created by the privileged account.

I'm using Ubuntu 22.04. KraftKit version `kraft 0.6.7 (b139ae2a52608f2e4a910225be0c981f671fe75f) go1.20.6 2023-09-14T09:00:05Z`.

### Expected behavior

`kraft rm --all` should remove all Unikernel images / VMs, not end up in error.

### Which architectures were you using or does this bug affect?

x86_64

### Which operating system were you using or does this bug affect?

linux/debian

### Relevant log output

```shell
razvan@yggdrasil:~/projects/unikraft/maintainer-tools/workdir/apps/elfloader$ sudo kraft ps
NAME                      KERNEL                                 ARGS                                                                                 CREATED     STATUS  MEM   PLAT
agitated_joemartin        project://elfloader:kraftcloud-x86_64  /bin/python3.11 -m http.server 8080                                                  2 days ago  exited  0MiB  fc/x86_64
agitated_lucy             project://elfloader:kraftcloud-x86_64  /python -m http.server 8080                                                          1 day ago   exited  0MiB  fc/x86_64
charming_faben            project://elfloader:kraftcloud-x86_64  /lib64/ld-linux-x86-64.so.2 --library-path /lib /bin/python3.11 -m http.server 8080  2 days ago  exited  0MiB  fc/x86_64
charming_frodo            project://elfloader:kraftcloud-x86_64  /lib64/ld-linux-x86-64.so.2 --library-path /lib /bin/python3.11 -m http.server 8080  2 days ago  exited  0MiB  fc/x86_64
crazy_edgar               project://elfloader:kraftcloud-x86_64  /lib64/ld-linux-x86-64.so.2 --library-path /lib /bin/python3.11 http_server.py       2 days ago  exited  0MiB  fc/x86_64
eager_ayumu               project://elfloader:kraftcloud-x86_64  /usr/local/nginx/sbin/nginx -c /usr/local/nginx/conf/nginx.conf                      2 days ago  exited  0MiB  fc/x86_64
elegant_loon              project://elfloader:kraftcloud-x86_64  /python http_server.py                                                               1 day ago   exited  0MiB  fc/x86_64
[...]

PR #832, opened 2023-09-20
fix(machine): Remove link from store if already deleted

craciunoiuc:
### Prerequisite checklist

### Description of changes

This fixes the case where someone creates a network with kraftkit and removes it with something else like `ifconfig`/`ip link del` and the entry stays in the list.

Partial fix for: #820

### DIFF
--- a/machine/network/bridge/v1alpha1.go
+++ b/machine/network/bridge/v1alpha1.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
@@ -18,9 +19,12 @@ import (
 	"github.com/vishvananda/netlink/nl"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apiserver/pkg/storage"
 
 	networkv1alpha1 "kraftkit.sh/api/network/v1alpha1"
+	"kraftkit.sh/config"
 	"kraftkit.sh/machine/network/macaddr"
+	"kraftkit.sh/machine/store"
 )
 
 type v1alpha1Network struct{}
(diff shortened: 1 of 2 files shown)


machine/network/bridge/v1alpha1.go
```diff
@@ -356,42 +360,82 @@ func (service *v1alpha1Network) Update(ctx context.Context, network *networkv1al
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	var existingNetwork networkv1alpha1.Network
+	err = embeddedStore.Get(ctx, network.Name, storage.GetOptions{}, &existingNetwork)
+	if err != nil && !strings.Contains(err.Error(), "not found") {
+		return nil, err
+	}
```
nderjung, line 377: [The store is already wrapping the v1alpha1 implementation](https://github.com/unikraft/kraftkit/blob/staging/machine/network/register_linux.go#L24-L43), meaning that this obfuscates its use.

nderjung (commented):
In fact the `Delete` method should not be updated. The `Get`/[`List` method](https://github.com/unikraft/kraftkit/blob/staging/machine/network/bridge/v1alpha1.go#L474-L482) should be checked to ensure they only return valid entries from the host.

machine/network/bridge/v1alpha1.go
craciunoiuc: I think the wrapping itself makes everything feel obfuscated, and it's fine. As long as it works correctly >:D

craciunoiuc:
Ok then the fix would be to remove from the store on list?

craciunoiuc:
It will still fail if you run `ip link del` + `kraft net rm` without `kraft net ls` in between

nderjung:
When `kraft net ls` is run, eventually the network controller/v1alpha1 implemention will be called which calls `List`. This prototype will receive a "stale" list of networks in the positional parameter `networks`:
```go
func (service *v1alpha1Network) List(ctx context.Context, networks *networkv1alpha1.NetworkList) (*networkv1alpha1.NetworkList, error) {
                                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
The value of `networks` was populated by going through the store. [During the iteration of this list](https://github.com/unikraft/kraftkit/blob/staging/machine/network/bridge/v1alpha1.go#L478-L486), items need to be [added](https://github.com/unikraft/kraftkit/blob/staging/machine/network/bridge/v1alpha1.go#L513) and removed based on whether they exist or not. Clearly this is not working correctly here.

The result `List` should return an up-to-date list of networks which then is passed back to the store and saved.

The store essentially provides us with "enriched" metadata about items.

craciunoiuc:
Ok I can try to fix there. Hopefully the store is also updated by the returned value, otherwise things will never be removed there

craciunoiuc:
Converting to draft as not really ready

The bug is still there by the way

PR #835, opened 2023-09-21
fix(cmdfactory): Set arguments during initial object attribution

nderjung:
### Prerequisite checklist

### Description of changes

This commit fixes an issue where arguments were not attributed to the initial object (representing the command invoked by `New`).  In KraftKit, this did not matter since the arguments are passed again by `AttributeFlags` when the internal config is associated with the root command during the bootstrapping of the core application.  However, when used externally, and without using the `AttributeFlags` method independently from `New`, this causes a late-parsing issue where flags were not fully parsed before relevant subsystems were invoked.  For example, when calling the `--help` flag would not work because the system to recognise this has already executed.


craciunoiuc:
@nderjung maybe you can also revisit this if you are touching cmdfactory

PR #839, opened 2023-09-27
feat(workflows): Add multi-os stable tests

craciunoiuc:
### Prerequisite checklist

### Description of changes

This adds testing when pushing to stable on all OSes available through GitHub on Linux and Mac.
Of course we should catch this before pushing to stable, but it will take A Lot of time ti run all those so it's better if we only do it ~once a week.

I only copied the existing tests and added some duplication + matrixes.

### DIFF
--- a/test/e2e/cli/net_down_test.go
+++ b/test/e2e/cli/net_down_test.go
@@ -7,6 +7,7 @@ package cli_test
 
 import (
 	"fmt"
+	"runtime"
 
 	. "github.com/onsi/ginkgo/v2" //nolint:stylecheck
 	. "github.com/onsi/gomega"    //nolint:stylecheck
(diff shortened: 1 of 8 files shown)


antoineco:
Can we temporarily enable this workflow for this PR, just to at least ensure that it completes? Then we can revert back to only running on `stable`.

craciunoiuc:
sure, we can remove it after it completes

craciunoiuc:
Added, let's see

antoineco (approved):

test/e2e/cli/net_down_test.go
```diff
@@ -24,6 +25,10 @@ var _ = Describe("kraft net down", func() {
 	var cfg *fcfg.Config
 
 	BeforeEach(func() {
+		if runtime.GOOS != "linux" {
+			Skip("This test requires Linux right now")
```
nderjung: Could you please cite this issue in the message:
```
Skip("This test only supports Linux. See here for more information: https://github.com/unikraft/kraftkit/issues/840")
```

nderjung: (same with other instances)

PR #847, opened 2023-09-29
fix(cmdfactory): Do not set `defValue` for integer attributes

nderjung:
### Prerequisite checklist

### Description of changes

This commit fixes an issue with setting the value associated with an integer attribute after parsing a struct and how the default value is also set.

The issue is discovered where an attribute that is associated with flag attributes has a non-zero default value:

```go
type MyStruct struct {
  IntAttr int `long:"int-attr" usage:"Int flag" default:"1"`
}
```

In the above, the default value when read by `cobra.Run` is set to `0` instead of `1` when the flag is not set.  This is because the value from `strValue` overwrites the default value.  Unlike an empty string, `flags.Set` recognizes `0` as a real value as opposed to an "unset value".

Removing the logic which sets this for an integer remedies this issue and the value of `IntAttr` without disruping a new value when the flag is set with an alternative value from the default.


craciunoiuc (approved):
All good here. Thanks!

PR #848, opened 2023-09-29
fix(initrd): trimmed file mode on symlinks

antoineco:
### Prerequisite checklist

### Description of changes

`d.Type().Perm()` trimms the actual _mode_ bits in the CPIO header, therefore resulting in regular zero-sized files instead of symlinks.

I instead applied the same logic as in https://pkg.go.dev/github.com/cavaliergopher/cpio#FileInfoHeader and set `h.Mode |= TypeSymlink`, etc. based on the file type.

Additionally:
- I refactored the logic a little for readability.
- ~I skipped writing directories, since those are irrelevant inside a CPIO archive (unless specific perms need to be applied to some directories).~

### DIFF
--- a/initrd/initrd.go
+++ b/initrd/initrd.go
@@ -177,74 +177,73 @@ func NewFromMapping(workdir, output string, maps ...string) (*InitrdConfig, erro
 					return nil
 				}
 
-				internal = "." + strings.ReplaceAll(filepath.Join(initrdPath, internal), "//", "/")
+				if internal, err = filepath.Rel("/", internal); err != nil {
+					return fmt.Errorf("trimming leading separator from path %q: %w", internal, err)
+				}
 
 				info, err := d.Info()
 				if err != nil {
 					return err
 				}
 
-				if d.Type().IsDir() {
-					if err := writer.WriteHeader(&cpio.Header{
-						Name: internal,
-						Mode: cpio.TypeDir,
-					}); err != nil {
-						return err
+				header := &cpio.Header{
+					Name:    internal,
+					Mode:    cpio.FileMode(info.Mode().Perm()),
+					ModTime: info.ModTime(),
+				}
+
+				switch {
+				case info.Mode().IsDir():
+					header.Mode |= cpio.TypeDir
+
+					if err := writer.WriteHeader(header); err != nil {
+						return fmt.Errorf("writing cpio header for %q: %w", internal, err)
 					}
 
-					return nil
+				case info.Mode().IsRegular():
+					header.Mode |= cpio.TypeReg
+					header.Size = info.Size()
+
+					if err := writer.WriteHeader(header); err != nil {
+						return fmt.Errorf("writing cpio header for %q: %w", internal, err)
+					}
+
+					if err := copyData(func() (io.ReadCloser, error) { return os.Open(path) }, writer); err != nil {
+						return fmt.Errorf("copying file %q to cpio archive: %w", path, err)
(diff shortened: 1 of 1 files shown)


craciunoiuc:
@antoineco

> (unless specific perms need to be applied to some directories)

Does this mean that directories will now always have default permissions?

I am okay with this, as I can't find any case where special permissions are needed (we have a single user in Unikraft anyways)

antoineco:
@craciunoiuc that's right, there will not be any explicit inode created for directories, so whatever extracts the archive will assume default permissions.

This is the same as tar archives by the way. It is common to find tarballs without directory entries, but when you extract them you still get directories.

antoineco:
@craciunoiuc I restored the writing of directories. Unikraft doesn't currently support the automated creation of directories I was referring to.

The PR should be ready to go.

initrd/initrd.go
```diff
@@ -177,74 +177,73 @@ func NewFromMapping(workdir, output string, maps ...string) (*InitrdConfig, erro
 					return nil
 				}
 
-				internal = "." + strings.ReplaceAll(filepath.Join(initrdPath, internal), "//", "/")
+				if internal, err = filepath.Rel("/", internal); err != nil {
```
craciunoiuc, line 180:
```suggestion
				if internal, err = filepath.Rel(string(filepath.Separator), internal); err != nil {
```

initrd/initrd.go
```diff
@@ -177,74 +177,73 @@ func NewFromMapping(workdir, output string, maps ...string) (*InitrdConfig, erro
...
-						Mode: cpio.TypeDir,
-					}); err != nil {
-						return err
+				header := &cpio.Header{
+					Name:    internal,
+					Mode:    cpio.FileMode(info.Mode().Perm()),
+					ModTime: info.ModTime(),
+				}
+
+				switch {
```
craciunoiuc, line 195: I would also add a default case so we can see if people ever have any other type of files

craciunoiuc (changes requested):
just two comments from my side, will try now

initrd/initrd.go
craciunoiuc: Or you can leave like that and I modify it in the WIndows PR

initrd/initrd.go
antoineco: And what would it do?

craciunoiuc: Just print something like `Unknown file type` and print it

initrd/initrd.go
antoineco: `string(filepath.Separator)` might not make a difference on Windows because we are generating UNIX file paths here (relative to a literal `/`). For this to work on Windows, the entire path would need to be transformed before writing the CPIO file header.

Example:
```
C:\\my\src\dir\etc\config.json -> /my/src/dir/etc/config.json
```
And then the file name we write using `rel("/")` is `my/src/dir/etc/config.json`.

initrd/initrd.go
antoineco: The only issue is that this method does not accept a `Context` so we cannot access the logger.
If I use `fmt.Print` it's going to mess the output up.

It is really critical to change the signature just to accept a logger? Any other file type would be irrelevant inside a unikernel.

antoineco:
Closing in favour of #850.
The fix included in this PR is writing files with the correct file mode, but symlink targets are still empty due to an upstream issue.

PR #856, opened 2023-10-03
fix(internal): Attribute CLI flags after reading config

craciunoiuc:
### Prerequisite checklist

### Description of changes

Because we read the config file at a path, the previously read CLI flags were overwritten. To fix we attribute the flags again over the newly-read config.

### DIFF
--- a/internal/cli/options.go
+++ b/internal/cli/options.go
@@ -126,9 +126,7 @@ func WithDefaultConfigManager(cmd *cobra.Command) CliOption {
 		if err != nil {
 			return err
 		}
-		if err := cmdfactory.AttributeFlags(cmd, cfg, args...); err != nil {
-			return err
-		}
+		configDir := config.FetchConfigDirFromArgs(args)
 
 		// Did the user specify a non-standard config directory?  The following
 		// check is possible thanks to the attribution of flags to the config file.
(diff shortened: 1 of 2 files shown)


jake-ciolek (approved):

cmdfactory/builder.go
```diff
@@ -478,6 +478,28 @@ func AttributeFlags(c *cobra.Command, obj any, args ...string) error {
...
+		} else {
+			if !strings.HasPrefix(args[idx+1], "-") {
+				path = args[idx+1]
+			}
+		}
+		break
+	}
+	return
+}
+
```
nderjung: Since this is KraftKit-specific, i'd migrate it into `kraftkit.sh/config` as this is where the definition of `--config-dir` exists as well.

Additionally, since this is parsing args and to be more explicit in its method name, let's renamed it to `FetchConfigDirFromArgs`.

internal/cli/options.go
```diff
@@ -126,17 +126,15 @@ func WithDefaultConfigManager(cmd *cobra.Command) CliOption {
 		if err != nil {
 			return err
 		}
-		if err := cmdfactory.AttributeFlags(cmd, cfg, args...); err != nil {
-			return err
-		}
+		cpath := cmdfactory.FetchAlternateConfigDirectory(args)
```
nderjung:
```go
configDir := config.FetchConfigDirFromArgs(args)
```

cmdfactory/builder.go
craciunoiuc: Hope the file is correct :crossed_fingers:

nderjung (approved):
Awesome thanks!

PR #868, opened 2023-10-09
refactor: Use dependency injection for configuration passing

jake-ciolek:
### Prerequisite checklist

### Description of changes

Switch from using the context based config accessor towards dependency injection.

Changes the architecture a bit so that config data flow is more explicit.

Left a few todo's, but overall this makes reasoning about config handling a bit easier. Should make unit/integration testing easier. Improves clarity when debugging.

The generic stuff is still around, but we can think whether the added abstractions are worth it and whether using a simpler way would be better.

Another architecture decision point is on what type of the param shall we inject. Perhaps an interface would be better.

The test suite results are equivalent to 2828af22. The app-helloworld target builds and runs.

### DIFF
--- a/cmd/runu/runu.go
+++ b/cmd/runu/runu.go
@@ -39,29 +39,29 @@ type Runu struct {
 	SystemdCgroup bool   `long:"systemd-cgroup" usage:"enable systemd cgroup support"`
 }
 
-func New() *cobra.Command {
+func New(cfg *config.ConfigManager[config.KraftKit]) *cobra.Command {
 	cmd, err := cmdfactory.New(&Runu{}, cobra.Command{
 		Short: "Run OCI-compatible unikernels",
 		CompletionOptions: cobra.CompletionOptions{
 			DisableDefaultCmd: true,
 			HiddenDefaultCmd:  true,
 		},
-	})
+	}, cfg)
 	if err != nil {
 		panic(err)
 	}
 
-	cmd.AddCommand(state.New())
-	cmd.AddCommand(create.New())
-	cmd.AddCommand(start.New())
-	cmd.AddCommand(kill.New())
-	cmd.AddCommand(delete.New())
-	cmd.AddCommand(ps.New())
+	cmd.AddCommand(state.New(cfg))
+	cmd.AddCommand(create.New(cfg))
+	cmd.AddCommand(start.New(cfg))
+	cmd.AddCommand(kill.New(cfg))
+	cmd.AddCommand(delete.New(cfg))
+	cmd.AddCommand(ps.New(cfg))
 
 	return cmd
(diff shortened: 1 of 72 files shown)


antoineco (approved):
This is going to be controversial, but you have my blessing 100 times. Thanks for tackling this!

cmd/runu/runu.go
```diff
@@ -84,36 +84,47 @@ func (opts *Runu) PersistentPre(cmd *cobra.Command, _ []string) error {
...
+	//                    command tree to initialize a config manager.
+	//                    Then, we'll use that to obtain the final hydrated command tree.
+	//                    This will do for now, but spend some time thinking about how to make it nicer.
+	cmdInit := New(nil)
+	_, args, err := cmdInit.Find(os.Args[1:])
+	if err != nil {
+		log.G(ctx).Error(err)
+	}
+
+	cfgMgr, err := cli.ConfigManagerFromArgs(args)
```
antoineco, line 110: In fact we don't need a config manager in runu. It's purely driven by CLI flags, as per the runc CLI spec. Feel free to get rid of it.

craciunoiuc (commented):
Looks about right from my side, after a short scan. The `G()` stuff is certainly confusing for new people

cmd/runu/runu.go
jake-ciolek: Thanks. I've tried to remove that dependency just now, but runu's Create command has a dependency on the config. It's being used by genMachineArgs(). The cmdfactory magic relies on param signatures. For it to work, the params of the commands need to have the specific signature (containing config). We can streamline it further down the line but it will require further work. Could we live with this coupling for now?

antoineco: Argh, right I forgot about that.
Alternatively, could we simply initialize an empty config and pass that wherever KraftKit absolutely requires one? `runu` really isn't meant to inherit KraftKit's config (the fact that it does is a hack right now), so if we strip that dependency to the minimum you might be able to do without a config manager.

jake-ciolek:
Shall I try rebasing this? A lot of effort went into this rewrite. For the future of the project, it'd be good to improve the architecture a bit.

craciunoiuc:
Hey, sure thing

I did not invest much time into it because it was still `Draft` and I did not monitor the progress of it

I can try and do a pass after rebasing today-tomorrow, but can't promise anything :christmas_tree: :santa: :mrs_claus:

craciunoiuc:
There is no big feature that might break the refactor afterwards afaik, so it's certainly doable to get in

craciunoiuc:
any luck rebasing this? By the number of conflicting files it looks scary, but things should move smoothly afterwards

PR #873, opened 2023-10-11
feat!: Introduce OCI multi-image index manifests

nderjung:
### Prerequisite checklist

  - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
  - [x] Tested your changes against relevant architectures and platforms;
  - [x] Ran `make fmt` on your commit series before opening this PR;
  - [ ] Updated relevant documentation.

### Description of changes

> **Warning**
>  This PR contains breaking changes to the OCI package manager. Upon release, all packages built on the OCI package manager will need to be re-packaged!

This large refactor reworks the OCI package manager to assume the use of the OCI image specification's index such that we can represent multiple manifests within a single canonical referencable name.  This means that we can build multiple unikernels with the same name.  This makes sense for multi-target projects which build the same application but target different hardware/platform vendors.

The structure of the index can be found [here](https://github.com/opencontainers/image-spec/blob/main/image-index.md) and in this implementation we itemized each manifest in the index in the same way. When we use the `Catalog` method now, we are able to query a remote registry based on the selected OS (platform) and architecture as well as any embedded KConfig options (if they are embedded).

When packaging multiple targets of a single application, the resulting package will be listed for all unique targets:

```
TYPE  NAME        VERSION  FORMAT  DIGEST   PLAT
app   helloworld  latest   oci     e8be0a9  qemu/x86_64
app   helloworld  latest   oci     a6802f4  qemu/arm64
```

Use cases include selecting appropriately based on the host's architecture as well as feature selection via embedded KConfig options.

Notable additional improvements which are incorporated in this pull request include:
- Better offline mode in each package manager: all internal package managers were optimized to reduce instantiation time which makes a noticeable impact when using `kraft` in offline mode;
- Better outputs in `kraft pkg ls` which now has dynamic column data information;
- Better prompts in `kraft pkg` for selecting and packaging individual targets from an application project;
- `kraft pkg prune` has been renamed to `kraft pkg rm` and now works for all underlying implementations (per-manifest type for libraries and components, and directory and containerd for OCI packages);
- Better handling of insecure authenticated OCI registries; and,
- Natively selecting the host platform and architecture in `kraft pkg pull` when no additional flags are provided.

GitHub-Closes: #704
GitHub-Supersedes: #719
GitHub-Depends-On: #874
GitHub-Depends-On: #878

PR #878, opened 2023-10-13
fix(pkg): Add `--force` flag to sourcing and prevent check on unsourcing

nderjung:
### Prerequisite checklist

### Description of changes

This commit fixes an issue whereby the user will have the compatibility of a source checked before the source is saved to config.  To prevent this check, a new `kraft pkg source --force` flag is added which prevents this check.

Inversely, when a source is being removed from `kraft pkg unsource`, its compatibility does not need to be checked because it is being removed, not only is this less computation but if the source was manually added by hand and it was invalid, it would be impossible to remove.

This commit also updates relevant tests which are impacted by this change.  Specifically, tests that use a obviously fake source for testing purposes, the `--force` command is now utilized to ignore what would otherwise error.

### DIFF
--- a/cmd/kraft/pkg/source/source.go
+++ b/cmd/kraft/pkg/source/source.go
@@ -16,7 +16,9 @@ import (
 	"kraftkit.sh/packmanager"
 )
 
-type Source struct{}
+type Source struct {
+	Force bool `short:"F" long:"force" usage:"Do not run a compatibility test before sourcing."`
+}
 
 func New() *cobra.Command {
 	cmd, err := cmdfactory.New(&Source{}, cobra.Command{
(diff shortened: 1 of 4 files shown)


craciunoiuc (changes requested):
Young Skywalker,

You forgot to add e2e tests to check this this new flag. Source/Unsource we test so no excuses :smiling_imp: 

Edit: The tests should run `--force` with a good entry, and a bad entry

nderjung:
You're right Sensai. I just created [a new PR](https://github.com/unikraft/kraftkit/pull/879) which helps me debug this because i was not sure where it was needed :) 

I'll rebase this once that PR is merged so that i can see where i need to add it.

craciunoiuc:
That PR has been vanquished

nderjung:
There's a bit of a catch-22 with adding new tests here. There are several fixes included in #873 which correctly handle non-compliant sources under the manifest manager. I'll open an issue to address adding test cases for the `--force` flag which are dependent on that PR being upstreamed.

craciunoiuc (approved):
All good here. Thanks!

PR #884, opened 2023-10-16
fix: Prevent needless CLI parsing

nderjung:
### Prerequisite checklist

### Description of changes

Initially, `cmd.ParseFlags` in the `AttributeFlags` method was used to associate the provided list of command-line arguments with the respectively registered flag.  However, `AttributeFlags` has since been modified to make this association by directly calling `flag.Set`. Furthermore, calling `cmd.ParseFlags` consistently had the unintended side-affect of populating initialized slice structures with duplicate values.

To mitigate against double parsing, which results in duplicate values in, for example, slice types, remove this additional parsing of the flags.

We call `cmd.ParseFlags` only once after the config manager has been attributed as these values are required by subsequent context systems (e.g. logging).  To prevent double parsing on the root command, set `DisableFlagParsing` to true such that cobra's `ExecuteContext` method does not end up calling this again on the root command.

This also has the benefit of greatly simplify the code by removing [the previously imported `execute` and `executeC` to prevent internal parsing](https://github.com/unikraft/kraftkit/commit/1e6c1a4b38b6fa98e823b3bbaabfabf6d3faff2a) and fixing command completion and listing inherited flags.

### DIFF
--- a/cmdfactory/builder.go
+++ b/cmdfactory/builder.go
@@ -7,8 +7,6 @@ package cmdfactory
 
 import (
 	"context"
-	"errors"
-	"flag"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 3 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #892, opened 2023-10-19
fix(oci): Use auth when provided

craciunoiuc:
### Prerequisite checklist


nderjung:
~~Maybe it makes sense to still do `defaultAuths` and then overwrite with whatever is provided by the query?~~

If the user is setting auths, we should respect that.

nderjung (approved):

PR #894, opened 2023-10-20
fix(pkg): Handle single-target runtimes repackaging

nderjung:
### Prerequisite checklist

### Description of changes

This commit fixes an issue where a user would run `kraft pkg` with using a `Kraftfile` that uses the `runtime` element and the `runtime` element returns from the `Catalog` a single target.  Previously, the host platform and architecture were used when they were unset at the commandline and set to the host's platform and architecture.  Instead, query the catalog without these set and check if only a single entry is returned.  If it does, the user only wants to re-package this.  If we get more options then use the host architecture and platform and re-query the catalog.


craciunoiuc (approved):
All good here. Thanks!

PR #926, opened 2023-10-25
refactor(cli): Expose CLI actions as reusable methods

nderjung:
### Prerequisite checklist

### Description of changes

This PR enables re-using individual actions from `kraft` subcommands as reusable method actions.  Specifically, actions which are procedural and not informational are exposed (this means all list-type actions are excluded).  To make this change possible, the prototype for `cmdfactory.Run` was updated to use `ctx context.Context` as a positional argument as opposed to `cmd *cobra.Command`.  It turns out in all cases that `cmd` was used to access the command's context.  As a result, simply pass this context instead of the command.  This has the beneficial side-effect of allowing the `Run` method of subcommands to be used independently outside of the `cmdfactory`/subcommand structure as long as the relevant context and options populated.

For example, it is now possible to run a Unikraft build internally by calling:

```go
import "kraftkit.sh/internal/cli/kraft/build"

if err := build.Build(ctx, &build.BuildOptions{...}); err != nil {
  // ...
}
```

### DIFF
--- a/cmdfactory/builder.go
+++ b/cmdfactory/builder.go
@@ -40,7 +40,7 @@ type PreRunnable interface {
 }
 
 type Runnable interface {
-	Run(cmd *cobra.Command, args []string) error
+	Run(ctx context.Context, args []string) error
 }
 
 type fieldInfo struct {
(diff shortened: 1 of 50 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #947, opened 2023-10-27
feat(workflows): Enable code spellchecking

craciunoiuc:
### Prerequisite checklist

### Description of changes

This one is for you @nderjung 

If we want this to work properly we need to add some custom dictionaries to make sure that words like `unikraft` and `kraftkit`, etc. are caught

We can add single exceptions, but I think it warrants a dictionary. Examples:
<https://github.com/check-spelling/check-spelling/wiki/Feature:-Area-dictionaries>

We should not merge until then

### DIFF
--- a/.github/workflows/spellcheck.yaml
+++ b/.github/workflows/spellcheck.yaml
@@ -1,0 +1,29 @@
+name: spellcheck
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches: [staging]
+
+jobs:
+  spell-check:
+    runs-on: ubuntu-latest
+    name: Check Spelling
+    steps:
+      - name: Create filter file
+        run: |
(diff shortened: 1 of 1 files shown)


nderjung:
Thanks 😇  But is this not already enabled via [golangci-lint](https://github.com/unikraft/kraftkit/blob/staging/.golangci.yaml#L11) ?

craciunoiuc:
I think that is for variables only? Comments are for sure ommited.

Anyways, this is just an idea right now, they don't really say how this should be structured so I just tried some things

craciunoiuc:
Converting to draft cause it doesn't really work, also is not that urgent

craciunoiuc:
Closing, as this is not really needed and it doesn't really work in our case

Eye detector works well enough

PR #949, opened 2023-10-30
fix(tui): Do not add parents when requesting children 

craciunoiuc:
### Prerequisite checklist

### Description of changes

~This enforces the non-parallel way of adding items. It does not decrease the speed of the process tree.~
This was fixed and nothing sequential is done anymore.

Without this what happens is:
* All processes are added and started as requested
* More processes are started and executed until the process tree stops as all X processes have completed. More specifically, the parent is added every time afterwards.
* There is a chance that more processes are started and added between the moment all X processes have completed and the tree is stopped.

~For 4 processes there is around a 50% chance that at least an additional process is started, or, even worse, because the order is not guaranteed, the duplicates take precedence.~
Processes are started until the whole tree is finished so unless you have a quick internet connection, this will fail.

This means that there is no way to ensure order outside of the `processTree` (basically considering it a black box).

An example with the bug in action:
image

I currently do not have any ideas why this did not manifest itself until now.

Later Update:

The bug happened because the function that was supposed to return children also returned the parents. This resulted in infinite cycles that had a hard stop (the process tree finishing).

The fix removed the parent addition and also just inits the parents in `Init`, as the children are picked up by `Update` anyways.

### DIFF
--- a/tui/processtree/processtree.go
+++ b/tui/processtree/processtree.go
@@ -191,9 +191,8 @@ func (pt *ProcessTree) Init() tea.Cmd {
 		pt.timer.Init(),
 	}
 
-	// Start all child processes
-	children := pt.getNextReadyChildren(pt.tree)
-	for _, pti := range children {
+	// Start all child processes plus this one
+	for _, pti := range pt.tree {
 		pti := pti
 		pti.timeout = pt.timeout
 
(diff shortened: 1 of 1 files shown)


craciunoiuc:
No longer reproduces on latest staging.

Reasoning is still good so it should be kept in mind.

PR #976, opened 2023-11-04
fix(unikraft): Add additional checks for targets

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/unikraft/target/transform.go
+++ b/unikraft/target/transform.go
@@ -28,6 +28,10 @@ func TransformFromSchema(ctx context.Context, data interface{}) (interface{}, er
 	case string:
 		split := strings.SplitN(value, "/", 2)
 
+		if len(split) != 2 {
+			return nil, fmt.Errorf("invalid target format: %s", value)
+		}
+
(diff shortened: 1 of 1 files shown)


unikraft/app/loader.go
```diff
@@ -100,6 +100,12 @@ func NewApplicationFromInterface(ctx context.Context, iface map[string]interface
 		return nil, err
 	}
 
+	for _, target := range app.targets {
+		if target.Architecture().Name() == "" || target.Platform().Name() == "" {
+			return nil, fmt.Errorf("architecture and platform must be specified for all targets")
+		}
+	}
+
```
nderjung: We could do this earlier at the end of [`target.TransformFromSchema`](https://github.com/unikraft/kraftkit/blob/staging/unikraft/target/transform.go#L118) method, saving us having to run this loop. Then you can also two if error checks, one for plat and the other for arch.

craciunoiuc: Fixed and also added a check for the case where you would do `- string` with no `/`, which would result in a sigsegv

nderjung (approved):

PR #1025, opened 2023-11-22
feat(workflows): Do Mac notarization after releases

craciunoiuc:
### Prerequisite checklist

### Description of changes

This workflow will start after a release is published.

Things to have in mind when reviewing:
* `codesign` might not run a pipeline in its current form (might require keychain unlock)
* ~~tools might not appear as installed even though they are because of isolation~~ (probably not the case)
* the GH_PAT token might not have artifact pushing rights
* ~~`sed` formatting is untested, will have to look closely over that~~

In case this fails a script is available to do this automatically with a single run so not really urgent.

### DIFF
--- a/.github/workflows/notarization.yaml
+++ b/.github/workflows/notarization.yaml
@@ -1,0 +1,143 @@
+name: mac-notarization
+
+on:
+  release:
+    types: [released]
+
+jobs:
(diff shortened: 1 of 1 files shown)


nderjung (approved):
Thanks!

PR #1027, opened 2023-11-23
fix(kconfig): Set the value of `No` as `n`

nderjung:
### Prerequisite checklist

### Description of changes

This is the real value which is understood by both this KConfig parser and the parser used by Unikraft.  This value is an artifact from the original source which is not fully compatible.


craciunoiuc (approved):
All good here. Thanks!

PR #1032, opened 2023-11-23
feat: Introduce `.kraftignore`

### LINKED ISSUE #135: Introduce `.kraftignore`
This issue tracks ignoring files during either the build process (where files are recognised) or packaging process (where a filesystem is created), a `.kraftignore` file should exclude any glob-patterns from being used.

mdsahiloss:
### Prerequisite checklist

### Description of changes

This PR fixes https://github.com/unikraft/kraftkit/issues/135

Specifying contents (file or directory paths) in `.kraftignore` file in the working directory forces `Kraftkit` to ignore matching children files or directories sharing from the sharing directory as `rootfs`.

For now, `.kraftignore` only work when `rootfs` is specified for a directory path.

**Rules for `.kraftignore`:**
* Line starts with a `#` will be ignored by `Kraftkit`.
* Every blank line will be ignored by `Kraftkit`.
* Directory or File path that contains `glob-patterns` (characters like `*`, `!`, `[` and `?`) will be ignored by `Kraftkit` and return a warning message.
* File or Directory name can be enclosed by single or double quotes (`'filename'`, `"dir_name"`).
* File or Directory names which contain white spaces must be enclosed by single or double quotes. e.g.
   ```yaml
   "sample file"
   'sample new file'
   ```
* One needs to specify the entire relative path (relative path to the sharing directory e.g. if `rootfs=share/` and `share` directory contains `sample.txt`and `sample/`, So relative path in `.kraftignore` file can be defined as `/sample.txt`, `sample.txt` and `./sample.txt` for `share/sample.txt` to ignore) of the file or directory that have to be ignored.
* Various directory and file paths can be specified in a `.kraftignore` file using space separator or new line e.g.

  ```yaml
  random1/random01/random01.txt /random.txt
  
  random1/random01
  ```
* All the specified paths in `.kraftignore` will be relative to the root of sharing directory, means if `share/` directory contains `sample.txt` and `sample/`, and set as `rootfs` to kraftkit then to ignore `/share/sample.txt` only `sample.txt` needs to be defined in `kraftignore`.
**Note: All the paths can start with `./` or `/`** (Since prefix `/` and `./` with paths simply ignored by Kraftkit).

### DIFF
--- a/initrd/directory.go
+++ b/initrd/directory.go
@@ -17,6 +17,8 @@ import (
 	"kraftkit.sh/log"
 )
 
+var ErrRootfsOutputAlreadyExists = fmt.Errorf("rootfs output already exists")
+
 type directory struct {
 	opts  InitrdOptions
 	path  string
@@ -66,7 +68,7 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
 	}
 
 	if _, err := os.Stat(initrd.opts.output); err == nil {
-		return "", fmt.Errorf("output already exists")
+		return initrd.opts.output, ErrRootfsOutputAlreadyExists
 	}
 
 	f, err := os.OpenFile(initrd.opts.output, os.O_RDWR|os.O_CREATE, 0o644)
--- a/initrd/directory.go
+++ b/initrd/directory.go
@@ -65,6 +65,10 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
 		}
 	}
 
+	if _, err := os.Stat(initrd.opts.output); err == nil {
+		return "", fmt.Errorf("output already exists")
+	}
+
 	f, err := os.OpenFile(initrd.opts.output, os.O_RDWR|os.O_CREATE, 0o644)
 	if err != nil {
 		return "", fmt.Errorf("could not open initramfs file: %w", err)
@@ -74,6 +78,11 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
 
 	writer := cpio.NewWriter(f)
 	defer writer.Close()
+
+	ignoringItems, err := GetKraftIgnoreItems(ctx, initrd.path)
+	if err != nil {
+		return "", err
+	}
 
 	if err := filepath.WalkDir(initrd.path, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
@@ -85,6 +94,21 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
 			return nil // Do not archive empty paths
 		}
 		internal = "." + filepath.ToSlash(internal)
+
+		if len(ignoringItems) > 0 && path != initrd.path {
+			switch isExistInKraftignoreFile(internal, d, ignoringItems) {
+			case SkipDir:
+				log.G(ctx).
+					WithField("directory", internal).
+					Trace("ignoring from archiving")
+				return filepath.SkipDir
+			case Exist:
+				log.G(ctx).
+					WithField("file", internal).
+					Trace("ignoring from archiving")
+				return nil
+			}
+		}
 
 		info, err := d.Info()
 		if err != nil {
--- a/initrd/kraftignore.go
+++ b/initrd/kraftignore.go
@@ -1,0 +1,157 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2024, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+package initrd
+
+import (
+	"bufio"
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"kraftkit.sh/log"
+)
+
+// kraftignore filename
+const KraftignoreFileName = ".kraftignore"
+
+type IgnoringFileType string
+
+const (
+	Exist    = IgnoringFileType("Exist")
+	NotExist = IgnoringFileType("NotExist")
+	SkipDir  = IgnoringFileType("SkipDir")
+)
+
+// GetKraftIgnoreItems returns file and directory names specified in .kraftignore
+func GetKraftIgnoreItems(ctx context.Context, dir string) ([]string, error) {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return []string{}, err
+	}
+	kraftignorePath := filepath.Join(cwd, KraftignoreFileName)
+
+	if _, err := os.Stat(kraftignorePath); errors.Is(err, os.ErrNotExist) {
+		return []string{}, nil
+	} else if err != nil {
+		return []string{}, err
+	}
+
+	kraftignoreFile, err := os.Open(kraftignorePath)
+	if err != nil {
+		return []string{}, err
+	}
+
+	defer func() {
+		kraftIgnoreErr := kraftignoreFile.Close()
+		if kraftIgnoreErr != nil {
+			if err != nil {
+				err = fmt.Errorf("%w: %w", err, kraftIgnoreErr)
+			} else {
+				err = kraftIgnoreErr
+			}
+		}
+	}()
+
+	kraftignoreScanner := bufio.NewScanner(kraftignoreFile)
+	kraftignoreScanner.Split(bufio.ScanLines)
+	var kraftignoreFileLines, ignoringItems []string
+
+	for kraftignoreScanner.Scan() {
+		kraftignoreFileLines = append(kraftignoreFileLines, kraftignoreScanner.Text())
+	}
+
+	for lineNum, line := range kraftignoreFileLines {
+		line = strings.TrimSpace(line)
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		items := findLineItems(line)
+		for _, item := range items {
+			if item == "" || item == "#" {
+				continue
+			}
+
+			if hasGlobPatterns(item) {
+				log.G(ctx).
+					WithField("file", kraftignorePath).
+					Warn("contains a glob pattern ", item,
+						" at line ", lineNum,
+						" which is not supported by Kraftkit")
+				continue
+			}
+
+			if _, err := os.Stat(filepath.Join(dir, item)); os.IsNotExist(err) {
+				log.G(ctx).
+					WithField("file", kraftignorePath).
+					Warn("contains ", item,
+						" at line ", lineNum,
+						" which does not exist in the provided rootfs directory")
+				continue
+			}
+
+			ignoringItems = append(ignoringItems, item)
+		}
+	}
+
+	return ignoringItems, err
+}
+
+// isExistInKraftignoreFile checks if the path exist in .kraftignore
+func isExistInKraftignoreFile(internal string, pathInfo fs.DirEntry, kraftignoreItems []string) IgnoringFileType {
+	for _, ignoringItem := range kraftignoreItems {
+		if internal == ignoringItem {
+			if pathInfo.IsDir() {
+				return SkipDir
+			}
+			return Exist
+		}
+	}
+	return NotExist
+}
+
+// hasGlobPatterns checks if the item contains glob pattern
+func hasGlobPatterns(item string) bool {
+	return strings.ContainsAny(item, "*?![{")
+}
+
+// findLineItems finds items in a line of .kraftignore
+func findLineItems(line string) []string {
+	items := strings.Split(line, " ")
+	for index := 0; index < len(items); index++ {
+		charToFind := ""
+		if strings.HasPrefix(items[index], `"`) && !strings.HasSuffix(items[index], `"`) {
+			charToFind = `"`
+		} else if strings.HasPrefix(items[index], `'`) && !strings.HasSuffix(items[index], `'`) {
+			charToFind = `'`
+		}
+
+		if len(charToFind) > 0 {
+			i := index + 1
+			for ; i < len(items) && !strings.HasSuffix(items[i], charToFind); i++ {
+				items[index] += " " + items[i]
+				items = append(items[:i], items[i+1:]...)
+				i--
+			}
+			items[index] += " " + items[i]
+			items = append(items[:i], items[i+1:]...)
+		}
+		items[index] = strings.Trim(items[index], `"`)
+		items[index] = strings.Trim(items[index], `'`)
+		items[index] = strings.TrimSpace(items[index])
+		items[index] = strings.TrimPrefix(items[index], "../")
+		if !strings.HasPrefix(items[index], "./") {
+			if !strings.HasPrefix(items[index], "/") {
+				items[index] = "/" + items[index]
+			}
+			items[index] = "." + items[index]
+		}
+		items[index] = strings.TrimSuffix(items[index], string(filepath.Separator))
+	}
+	return items
+}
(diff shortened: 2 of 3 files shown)


initrd/kraftignore.go
```diff
@@ -0,0 +1,103 @@
...
+				continue
+			}
+
+			if strings.HasPrefix(item, "./") {
+				item = strings.TrimPrefix(item, ".")
+			} else if !strings.HasPrefix(item, "/") {
+				item = "/" + item
+			}
+
+			item = strings.TrimSuffix(item, "/")
```
craciunoiuc: Can you use `filepath.Separator` instead of `/` 🙏

initrd/directory.go
```diff
@@ -86,6 +95,21 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
...
 		}
 
+		if len(ignoringItems) > 0 && path != initrd.path {
+			switch isExistInKraftignoreFile(internal, d, ignoringItems) {
+			case "SkipDir":
+				log.G(ctx).
+					WithField("directory", internal).
+					Trace("ignoring from archiving")
+				return filepath.SkipDir
+			case "Exist":
```
craciunoiuc: This could be a list of consts instead of strings

initrd/directory.go
```diff
@@ -65,6 +65,10 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
 		}
 	}
 
+	if _, err := os.Stat(initrd.opts.output); err == nil {
+		os.Remove(initrd.opts.output)
```
craciunoiuc: I think this also returns an error, can you check if nil?

initrd/kraftignore.go
```diff
@@ -0,0 +1,103 @@
...
+		return []string{}, err
+	}
+
+	kraftignoreFile, err := os.Open(initrd.opts.kraftignorePath)
+	if err != nil {
+		return []string{}, err
+	}
+
+	defer func() {
+		err = kraftignoreFile.Close()
```
craciunoiuc: This will overwrite any error below. You could try just appending to the existing error if an error already exists

initrd/kraftignore.go
```diff
@@ -0,0 +1,103 @@
...
+			item = strings.TrimSpace(item)
+			if item == "" {
+				continue
+			}
+
+			if hasGlobPatterns(item) {
+				log.G(ctx).
+					WithField("file", initrd.opts.kraftignorePath).
+					Warn("contains a glob pattern ", item,
+						" which is not supported by Kraftkit")
```
craciunoiuc: Also specify that the line is skipped

initrd/kraftignore.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			return "Exist"
+		}
+	}
+	return ""
+}
+
+func hasGlobPatterns(item string) bool {
+	if strings.Contains(item, "*") || strings.Contains(item, "?") ||
+		strings.Contains(item, "!") || strings.Contains(item, "[") {
```
craciunoiuc: Optional: Check if a library out there can do this, maybe there are other symbols

craciunoiuc (changes requested):
Some initial comments from my side, did not try out the code yet

initrd/kraftignore.go
mdsahiloss: Done 👍🏽

initrd/directory.go
mdsahiloss: Nice catch 👍🏽

initrd/directory.go
mdsahiloss: Done 👍🏽

initrd/kraftignore.go
mdsahiloss: Done 👍🏽

initrd/kraftignore.go
mdsahiloss: Done 👍🏽

initrd/kraftignore.go
mdsahiloss: Hey! I couldn't find anything relavant, So updated this function to cover all the `globPatterns` as mentioned [here](https://code.visualstudio.com/docs/editor/glob-patterns#_glob-pattern-syntax).

craciunoiuc:
I think we also need to document this somewhere otherwise it will be lost

@nderjung do you have a use case in mind? maybe an app which would currently benefit from `.kraftignore`?

craciunoiuc:
Hey @MdSahil-oss found 1 small bug but otherwise it seems to work:

```
test.txt <-- works (ignored)

# test1.txt <-- works (not ignored)

nginx/test3.txt    #       test2.txt  <-- works (both ignored, basically # was ignored)

test.txt <-- works (nothing changed)

# test.txt <-- works (nothing changed)

./nginx <-- works (ignored)

./nginx/* <-- works (prints warning)

.///////////////////nginx <-- fails (multiple slashes are usually ignored)
./nginx/../test1.txt <-- fails (I think .. is not understood)
././././././././././././test1.txt <-- fails (I think . is not understood)
```

craciunoiuc:
One more thing: does it support using `""` or `''` characters? I was thinking about the case of `"./nginx/strange test.txt"`

craciunoiuc (changes requested):
The code looks okay, waiting on your response for the comments above

mdsahiloss:
> Hey @MdSahil-oss found 1 small bug but otherwise it seems to work:
> 
> ```
> test.txt <-- works (ignored)
> 
> # test1.txt <-- works (not ignored)
> 
> nginx/test3.txt    #       test2.txt  <-- works (both ignored, basically # was ignored)
> 
> test.txt <-- works (nothing changed)
> 
> # test.txt <-- works (nothing changed)
> 
> ./nginx <-- works (ignored)
> 
> ./nginx/* <-- works (prints warning)
> 
> .///////////////////nginx <-- fails (multiple slashes are usually ignored)
> ./nginx/../test1.txt <-- fails (I think .. is not understood)
> ././././././././././././test1.txt <-- fails (I think . is not understood)
> ```

Nice catch @craciunoiuc, I will work on your suggestions

mdsahiloss:
> One more thing: does it support using `""` or `''` characters? I was thinking about the case of `"./nginx/strange test.txt"`

Don't think so, I will update this PR for these characters.

initrd/kraftignore.go
```diff
@@ -0,0 +1,129 @@
...
+	kraftignoreFile, err := os.Open(initrd.opts.kraftignorePath)
+	if err != nil {
+		return []string{}, err
+	}
+
+	defer func() {
+		kraftIgnoreErr := kraftignoreFile.Close()
+		if kraftIgnoreErr != nil {
+			if err != nil {
+				err = fmt.Errorf("%v\n%v", err, kraftIgnoreErr)
```
craciunoiuc:
```suggestion
				err = fmt.Errorf("%w: %w", err, kraftIgnoreErr)
```

initrd/kraftignore.go
```diff
@@ -0,0 +1,129 @@
...
+	return NotExist
+}
+
+func hasGlobPatterns(item string) bool {
+	if strings.Contains(item, "*") || strings.Contains(item, "?") ||
+		strings.Contains(item, "!") || strings.Contains(item, "[") ||
+		strings.Contains(item, "{") {
+		return true
+	}
+	return false
```
craciunoiuc:
```suggestion
  return strings.ContainsAny(item, "*?![{") 
```

initrd/kraftignore.go
```diff
@@ -0,0 +1,129 @@
...
+	for kraftignoreScanner.Scan() {
+		kraftignoreFileLines = append(kraftignoreFileLines, kraftignoreScanner.Text())
+	}
+
+	for lineNum, line := range kraftignoreFileLines {
+		line = strings.TrimSpace(line)
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		items := strings.Split(line, " ")
```
craciunoiuc: Why do you split on space but also trim it afterwards, are you sure this works with files that have spaces in their name?

initrd/kraftignore.go
```diff
@@ -0,0 +1,129 @@
...
+			if hasGlobPatterns(item) {
+				log.G(ctx).
+					WithField("file", initrd.opts.kraftignorePath).
+					Warn("contains a glob pattern ", item,
+						" at line ", lineNum,
+						" which is not supported by Kraftkit")
+				continue
+			}
+
+			item = strings.TrimPrefix(item, "..")
```
craciunoiuc: Same here, I don't think we should trim this, but rather what I would do would be:
1. Find the full path of the initrd directory
1. Append the file path to the initrd directory
1. Check if it evaluates to a valid file, if yes, trim `1.` from it and use that given path, if not, ignore

This goes for the whole file parsing logic

craciunoiuc (changes requested):
Hey @MdSahil-oss, some more comments from my side

initrd/kraftignore.go
mdsahiloss: Done 👍🏽

initrd/kraftignore.go
mdsahiloss: 😱 What! `strings` lib has `containsAny` func.

BTW Done 👍🏽

initrd/kraftignore.go
mdsahiloss: I didn't implement this for files that have spaces in their name, But now I've updated the PR to work for that as well

initrd/kraftignore.go
mdsahiloss: Updated the logic to work in your recommended way, Please review again.

But still removing `../` if item (a path mentioned in `.kraftignore`) contains `../` as prefix because path inside `rootfs` which starts `../xyz` refer to the file/dir outside of `rootfs` dir.

I hope this make sense to you otherwise ping me again ;)

initrd/kraftignore.go
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2022, Unikraft GmbH and The KraftKit Authors.
```
nderjung: You can put the year as `2024` now :)

initrd/options.go
```diff
@@ -4,10 +4,15 @@
...
+)
+
 type InitrdOptions struct {
-	output   string
-	cacheDir string
-	arch     string
+	output          string
+	cacheDir        string
+	arch            string
+	kraftignorePath string
```
nderjung: I think this option, `kraftignorePath`, can be removed and we can _always_ enable `.kraftignore` for relevant initrd providers. If the file exists the relevant procedure is accounted for.

initrd/options.go
```diff
@@ -41,3 +46,11 @@ func WithArchitecture(arch string) InitrdOption {
...
 	}
 }
+
+// WithKraftignorePath sets the path for .kraftignore.
+func WithKraftignorePath(dir string) InitrdOption {
+	return func(opts *InitrdOptions) error {
+		opts.kraftignorePath = filepath.Join(dir, KraftignoreFileName)
+		return nil
+	}
+}
```
nderjung: As a result of it being implicitly available, we can remove this option method.

initrd/directory.go
```diff
@@ -65,6 +65,12 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
 		}
 	}
 
+	if _, err := os.Stat(initrd.opts.output); err == nil {
+		if err = os.Remove(initrd.opts.output); err != nil {
+			return "", err
+		}
+	}
```
nderjung: This can be potentially dangerous operation since it is hidden away. Instead, simply error out if the output already exists instead of removing it:

```go
if _, err := os.Stat(initrd.opts.output); err == nil {
  return "", fmt.Errorf("output already exists")
}
```

initrd/kraftignore.go
```diff
@@ -0,0 +1,159 @@
...
+type IgnoringFileType string
+
+const (
+	Exist    = IgnoringFileType("Exist")
+	NotExist = IgnoringFileType("NotExist")
+	SkipDir  = IgnoringFileType("SkipDir")
+)
+
+// returns file and directory names specified in .kraftignore
+func getKraftignoreItems(ctx context.Context, initrd *directory) ([]string, error) {
```
nderjung: Let's expose this as a public method that accepts an input directory. This allows the method to be used for other applications outside of the `*directory` implementation.

```diff
- // returns file and directory names specified in .kraftignore
- func getKraftignoreItems(ctx context.Context, initrd *directory) ([]string, error) {
+ // GetKraftIgnoreItems returns file and directory names specified in .kraftignore
+ func GetKraftIgnoreItems(ctx context.Context, dir string) ([]string, error) {
```
You then simply need to replace `initrd.opts.kraftignorePath` with `dir`.

nderjung (commented):
Hi @MdSahil-oss! Thanks so much for this contribution, I think it's a valuable addition to the usability of `kraft`! I've left you some feedback as in-line comments below. Overall, the implementation looks good.

initrd/options.go
mdsahiloss: Done 👍🏽

mdsahiloss: Done 👍🏽

initrd/options.go
mdsahiloss: Done 👍🏽

initrd/kraftignore.go
mdsahiloss: Done 👍🏽

initrd/kraftignore.go
mdsahiloss: Done 👍🏽

initrd/directory.go
mdsahiloss: Done 👍🏽

mdsahiloss: Also, It causes test failure, So I updated test to pass this change.

initrd/directory.go
```diff
@@ -17,6 +17,8 @@ import (
 	"kraftkit.sh/log"
 )
 
+const RootfsOutputAlreadyExist = "rootfs output already exists"
```
nderjung: Let's declare this as an error, this is better standard practice:
```diff
- const RootfsOutputAlreadyExist = "rootfs output already exists"
+ var ErrRootfsOutputAlreadyExists = fmt.Errorf("rootfs output already exists")
```

initrd/kraftignore.go
```diff
@@ -0,0 +1,157 @@
...
+// kraftignore filename
+const KraftignoreFileName = ".kraftignore"
+
+type IgnoringFileType string
+
+const (
+	Exist    = IgnoringFileType("Exist")
+	NotExist = IgnoringFileType("NotExist")
+	SkipDir  = IgnoringFileType("SkipDir")
+)
```
nderjung, line 29: Same here, we can declare these as errors:
```go
var (
  ErrFileExists    = fmt.Errorf("file exists")
  ErrFileNotExists = fmt.Errorf("filt does not exist")
  ErrIgnoreDir     = fmt.Errorf("ignored directory")
)
````

Comment (nderjung) [reply initrd/kraftignore.go:29]: You can then do comparisons via:
```go
if errors.Is(err, ErrFileExists)
```

mdsahiloss: Surely, It can be done. But `kraftiignore.go` not uses these vars as errors, These vars are being used as notifiers for `filepath.WalkDir()` function in `directory.go`.

Please have an eye on `filepath.WalkDir()` function then lemme know if you really want me to change const vars as error vars :)

initrd/directory.go
mdsahiloss: Done 👍🏽

craciunoiuc:
what is the state on this @MdSahil-oss @nderjung 

I think it was almost ready to go out?

PR #1072, opened 2023-12-07
fix(internal): Handle unknown architectures

craciunoiuc:
### Prerequisite checklist

### Description of changes

Running a kernel with a wrong architecture resulted in a panic which was not suggestive.
Fix this by warning the user that the arch is unknown and properly init structures to not cause a panic.

### DIFF
--- a/internal/cli/kraft/run/runner_kernel.go
+++ b/internal/cli/kraft/run/runner_kernel.go
@@ -12,6 +12,7 @@ import (
 
 	machineapi "kraftkit.sh/api/machine/v1alpha1"
 	"kraftkit.sh/unikraft"
+	"kraftkit.sh/unikraft/arch"
 )
 
 // runnerKernel is a simple runner used for instantiating a prebuilt Unikraft
@@ -65,11 +66,11 @@ func (runner *runnerKernel) Prepare(ctx context.Context, opts *RunOptions, machi
 
 		switch fe.Machine {
 		case elf.EM_X86_64, elf.EM_386:
-			machine.Spec.Architecture = "x86_64"
+			machine.Spec.Architecture = arch.ArchitectureX86_64.String()
 		case elf.EM_ARM:
-			machine.Spec.Architecture = "arm"
+			machine.Spec.Architecture = arch.ArchitectureArm.String()
 		case elf.EM_AARCH64:
-			machine.Spec.Architecture = "arm64"
+			machine.Spec.Architecture = arch.ArchitectureArm64.String()
 		default:
 			return fmt.Errorf("unsupported kernel architecture: %v", fe.Machine.String())
 		}
--- a/internal/cli/kraft/run/utils.go
+++ b/internal/cli/kraft/run/utils.go
@@ -261,8 +261,7 @@ func (opts *RunOptions) prepareRootfs(ctx context.Context, machine *machineapi.M
 		fmt.Sprintf(initrd.DefaultInitramfsArchFileName, machine.Spec.Architecture),
 	)
 
-	fi, err := os.Stat(machine.Status.InitrdPath)
-	if err != nil {
+	if _, err := os.Stat(machine.Status.InitrdPath); err != nil {
 		ramfs, err := initrd.New(ctx,
 			opts.Rootfs,
 			initrd.WithOutput(machine.Status.InitrdPath),
@@ -277,18 +276,20 @@ func (opts *RunOptions) prepareRootfs(ctx context.Context, machine *machineapi.M
 			return fmt.Errorf("could not prepare initramfs: %w", err)
 		}
 
-		if _, err = ramfs.Build(ctx); err != nil {
+		if machine.Status.InitrdPath, err = ramfs.Build(ctx); err != nil {
 			return err
 		}
 	}
 
-	// Warn if the initrd path is greater than allocated memory
-	memRequest := machine.Spec.Resources.Requests[corev1.ResourceMemory]
-	if memRequest.Value() < fi.Size() {
-		log.G(ctx).Warnf("requested memory (%s) is less than initramfs (%s)",
-			humanize.Bytes(uint64(memRequest.Value())),
-			humanize.Bytes(uint64(fi.Size())),
-		)
+	if fi, err := os.Stat(machine.Status.InitrdPath); err == nil {
+		// Warn if the initrd path is greater than allocated memory
+		memRequest := machine.Spec.Resources.Requests[corev1.ResourceMemory]
+		if memRequest.Value() < fi.Size() {
+			log.G(ctx).Warnf("requested memory (%s) is less than initramfs (%s)",
+				humanize.Bytes(uint64(memRequest.Value())),
+				humanize.Bytes(uint64(fi.Size())),
+			)
+		}
 	}
 
 	return nil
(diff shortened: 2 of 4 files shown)


internal/cli/kraft/run/runner_kernel.go
```diff
@@ -75,6 +77,14 @@ func (runner *runnerKernel) Prepare(ctx context.Context, opts *RunOptions, machi
 		}
 	} else {
 		machine.Spec.Architecture = opts.Architecture
+
+		if opts.Architecture != "x86_64" &&
+			opts.Architecture != "arm" &&
+			opts.Architecture != "arm64" {
```
nderjung: Let's put this in a method returning a slice (similar to [this](https://github.com/unikraft/kraftkit/blob/staging/machine/platform/platform.go#L44)) inside of https://github.com/unikraft/kraftkit/blob/staging/unikraft/arch/ since this file contains known info about supported Unikraft architectures.
And use:
```go
import (
  "slices"
  ukarch "kraftkit.sh/unikraft/arch"
)
// [...]
if !slices.Contains[[]string](ukarch.KnownArchs(), opts.Architecture) {
```
See [`slices.Contains`](https://pkg.go.dev/slices#Contains)

internal/cli/kraft/run/runner_kernel.go
```diff
@@ -75,6 +77,14 @@ func (runner *runnerKernel) Prepare(ctx context.Context, opts *RunOptions, machi
...
+			log.G(ctx).WithFields(logrus.Fields{
+				"arch": opts.Architecture,
+			}).Warn("Architecture not known, assuming input is correct")
```
nderjung:
```diff
- }).Warn("Architecture not known, assuming input is correct")
+ }).Warn("unknown or incompatible")
```
Which will turn into:
```
 W  unknown or incompatible arch=boop42
```

internal/cli/kraft/run/utils.go
```diff
@@ -277,7 +277,14 @@ func (opts *RunOptions) prepareRootfs(ctx context.Context, machine *machineapi.M
...
 		}
 
-		if _, err = ramfs.Build(ctx); err != nil {
+		var newPath string
+		if newPath, err = ramfs.Build(ctx); err != nil {
+			return err
+		}
+
+		machine.Status.InitrdPath = newPath
+		fi, err = os.Stat(machine.Status.InitrdPath)
```
nderjung: Why `Stat` the InitrdPath? If it is empty, the underlying machine API (and store) will already detect this and act appropriately?

Additionally, for compactness:
```go
machine.Status.InitrdPath, err = ramfs.Build(ctx)
if err != nil {
  return err
}
``

Comment (craciunoiuc) [reply internal/cli/kraft/run/utils.go]: Because, otherwise, kraftkit panics below when it tries to use the fileinfo.

We either move the whole code, or init the fileinfo

Comment (nderjung) [reply internal/cli/kraft/run/utils.go]: Ah I see, we should wrap the memory size check with the stat:
```go
if fi, err := os.Stat(machine.Status.InitrdPath); err == nil {
   // Warn if the initrd path is greater than allocated memory
	memRequest := machine.Spec.Resources.Requests[corev1.ResourceMemory]
	if memRequest.Value() < fi.Size() {
		log.G(ctx).Warnf("requested memory (%s) is less than initramfs (%s)",
			humanize.Bytes(uint64(memRequest.Value())),
			humanize.Bytes(uint64(fi.Size())),
		)
	}
}
```

unikraft/arch/architecture.go
```diff
@@ -15,6 +15,69 @@ import (
...
+type ArchitectureName string
+
+const (
+	ArchitectureUnknown = ArchitectureName("unknown")
+	ArchitectureX86_64  = ArchitectureName("x86_64")
+	ArchitectureAmd64   = ArchitectureX86_64
+	ArchitectureArm64   = ArchitectureName("arm64")
+	ArchitectureAarch64 = ArchitectureArm64
+	ArchitectureArm     = ArchitectureName("arm")
+	ArchitectureAarch32 = ArchitectureArm
```
nderjung: Let's keep it simply instead of having aliases:
```go
	ArchitectureUnknown = ArchitectureName("unknown")
	ArchitectureX86_64  = ArchitectureName("x86_64")
	ArchitectureArm64   = ArchitectureName("arm64")
	ArchitectureArm     = ArchitectureName("arm")
```
This is what's in the core

unikraft/arch/architecture.go
```diff
@@ -15,6 +15,69 @@ import (
...
+
+// ArchitecturesByName returns the list of known architectures and their name alises.
+func ArchitecturesByName() map[string]ArchitectureName {
+	return map[string]ArchitectureName{
+		"x86_64":  ArchitectureX86_64,
+		"amd64":   ArchitectureAmd64,
+		"arm64":   ArchitectureArm64,
+		"aarch64": ArchitectureAarch64,
+		"arm":     ArchitectureArm,
+		"aarch32": ArchitectureAarch32,
```
nderjung: Map these back to the simplified set of constants

internal/cli/kraft/run/runner_kernel.go
```diff
@@ -65,16 +68,22 @@ func (runner *runnerKernel) Prepare(ctx context.Context, opts *RunOptions, machi
 
 		switch fe.Machine {
 		case elf.EM_X86_64, elf.EM_386:
-			machine.Spec.Architecture = "x86_64"
+			machine.Spec.Architecture = string(arch.ArchitectureX86_64)
```
nderjung: Since you implement `fmt.Stringer`, you can:
```
arch.ArchitectureX86_64.String()
```

internal/cli/kraft/run/runner_kernel.go
```diff
@@ -65,16 +68,22 @@ func (runner *runnerKernel) Prepare(ctx context.Context, opts *RunOptions, machi
...
 			return fmt.Errorf("unsupported kernel architecture: %v", fe.Machine.String())
 		}
 	} else {
 		machine.Spec.Architecture = opts.Architecture
+
+		if _, found := arch.ArchitecturesByName()[opts.Architecture]; !found {
+			log.G(ctx).WithFields(logrus.Fields{
+				"arch": opts.Architecture,
+			}).Warn("unknown or incompatible")
+		}
```
nderjung: Just checked out this PR to look at everything in context. Can we move this into `run.go` since it applies to all runners? It should be moved just after the `opts.discoverNetworkController` call.

unikraft/arch/architecture.go
```diff
@@ -15,6 +15,63 @@ import (
...
+func (ht ArchitectureName) String() string {
+	return string(ht)
+}
+
+// ArchitectureByName returns the architecture for a given name.
+// If the name is not known, it returns it unchanged.
+func ArchitectureByName(name string) ArchitectureName {
+	architectures := ArchitecturesByName()
+	if _, ok := architectures[name]; !ok {
+		return ArchitectureName(name)
```
nderjung: Shouldn't this be `unknown`?

craciunoiuc: That's what I thought when I copied it from the Platform code

nderjung: Ok. Platform code should later be updated :)

nderjung (approved):

PR #1086, opened 2023-12-11
fix(buildenvs): Enable network user mode in QEMU build (via slirp)

nderjung:
### Prerequisite checklist


nderjung:
@craciunoiuc I see you solved the build issues by downgrading QEMU from 7.2 to 7.1. Why don't we just go up to qemu 8.X? Since we a). support it in kraftkit itself and b). are using this qemu build itself for CI purposes and reproducibility purposes. I remember you saying there was a small perf drop because of the downgrade.

craciunoiuc:
@nderjung I also tried the latest `8.X` qemu in my attempts.

The problem in this PR is not solved by that, meaning that it won't have slirp compiled in or compilable-in, which results in two successive solutions:
1. Revert that specific commit in `qemu` that allows for slirp compiling
or
2. Compile non-statically and add all dependencies manually (will require extensive Dockerfile re-write)

If we are blocking on this, I say merge this PR and then work to make 2. happen. 1. needs too much maintaining and it's not worth it.

craciunoiuc:
> Compile non-statically and add all dependencies manually (will require extensive Dockerfile re-write)

Turns out, it was not so extensive after-all. This should now work and provide us with the latest qemu which can do what we want

craciunoiuc (approved):
All good here me. Thanks!

PR #1112, opened 2023-12-20
feat(net): Prevent removing networks that are in use

LucaSeri:
Check whether there are any running machines that are using the network before removing it. Additionally introduce a force flag that can override this.

### Prerequisite checklist

### Description of changes
This PR adds a check that prevents a user from removing a network in use, unless forcing it.

### DIFF
--- a/internal/cli/kraft/net/remove/remove.go
+++ b/internal/cli/kraft/net/remove/remove.go
@@ -11,14 +11,18 @@ import (
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	machineapi "kraftkit.sh/api/machine/v1alpha1"
 	networkapi "kraftkit.sh/api/network/v1alpha1"
 	"kraftkit.sh/cmdfactory"
 	"kraftkit.sh/iostreams"
+	"kraftkit.sh/log"
 	"kraftkit.sh/machine/network"
+	mplatform "kraftkit.sh/machine/platform"
 )
 
 type RemoveOptions struct {
 	Driver string `noattribute:"true"`
+	Force  bool   `long:"force" short:"f" usage:"Force removal of the network" default:"false"`
 }
 
 // Remove a local machine network.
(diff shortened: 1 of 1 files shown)


craciunoiuc (commented):
looks good on an initial view

once out of draft I'll also try it out

internal/cli/kraft/net/remove/remove.go
```diff
@@ -65,6 +68,29 @@ func (opts *RemoveOptions) Run(ctx context.Context, args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	for _, machine := range machines.Items {
+		if machine.Status.State != machineapi.MachineStateRunning {
+			continue
+		}
+		for _, network := range machine.Spec.Networks {
+			if network.IfName == args[0] {
```
jake-ciolek, line 91: It'd be good to ensure the length of args somewhere before this.

internal/cli/kraft/net/remove/remove.go
```diff
@@ -65,6 +68,29 @@ func (opts *RemoveOptions) Run(ctx context.Context, args []string) error {
...
+				if !opts.Force {
+					return fmt.Errorf("network %s is in use by machine %s. Use --force to remove it anyway", args[0], machine.Name)
```
jake-ciolek, line 93: see above

internal/cli/kraft/net/remove/remove.go
LucaSeri: Nice catch, thanks for this! I added an additional check.

internal/cli/kraft/net/remove/remove.go
```diff
@@ -65,6 +72,42 @@ func (opts *RemoveOptions) Run(ctx context.Context, args []string) error {
...
+				} else {
+					fmt.Fprintf(iostreams.G(ctx).ErrOut, "warning: network %s is in use by machine %s. Removing anyway\n", args[0], machine.Name)
```
nderjung: Let's use the native warning log-level:
```go
log.G(ctx).Warnf("network '%s' is in use by machine '%s'", args[0], machine.Name)
```

LucaSeri: Fixed, thanks!

nderjung (approved):
Thanks!

PR #1160, opened 2024-01-17
fix(run): Always perform a check of the host's platform

nderjung:
### Prerequisite checklist

### Description of changes

Previously, this step was skipped when the user manually sets the platform via the `--plat` argument.  However, in some circumstances, the user can supply the same platform as the detected platform.  In this case, the mechanism which performs the check is skipped.  That same check also detects whether the host platform exposes virtualization extensions (and therefore whether emulation mode is necessary).  This would in turn lead to a non-obvious error message that would require the user to manually request emulation mode.

Instead, by always performing this check we can detect emulation mode. Additionally, this information can enrich peripheral error messages during the setting of the platform.

### DIFF
--- a/internal/cli/kraft/run/run.go
+++ b/internal/cli/kraft/run/run.go
@@ -190,20 +190,22 @@ func (opts *RunOptions) discoverMachineController(ctx context.Context) error {
 
 	opts.platform = mplatform.PlatformUnknown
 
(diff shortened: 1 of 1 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #1167, opened 2024-01-18
fix(run): Do not attach initrd when project embeds an initrd

nderjung:
### Prerequisite checklist

### Description of changes

This PR fixes an issue where a project's `Kraftfile` specifies an embedded filesystem and would otherwise attach the same filesystem as an en external initramfs when after the build it would be executed (i.e. `kraft build` followed by `kraft run`).  In the context of a project, this makes no sense since, even with user-provided (UP) which would a). otherwise overwrite all filesystem entries with the same files and b). these files would not differ since they are part of the same project.


razvand (approved):
Reviewed-by: Razvan Deaconescu <razvand@unikraft.io>

PR #1179, opened 2024-01-21
fix(tui): Correctly handle new `ctx` with custom log and iostreams out

nderjung:
### Prerequisite checklist

### Description of changes

This PR fixes an issue where the "stdout" of both the general logger and the iostreams was incorrectly set for all "subprocesses" of `paraprogress` and `processtree`.  Since the `context.Context` object was not unique per process, all subprocesses would output to the last process (becauase of the iteration process which attempted to set the output writer).

To make it so that each context is unique, perform a deepcopy of only the log and iostreams context objects and modify appropriately with new values.  The rest of the context is re-used as-is, allowing subprocesses to continue to access values as necessary.

### DIFF
--- a/tui/paraprogress/paraprogress.go
+++ b/tui/paraprogress/paraprogress.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/LastPossum/kamino"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/muesli/termenv"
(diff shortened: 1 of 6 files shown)


tui/processtree/processtree.go
```diff
@@ -110,6 +111,44 @@ func NewProcessTree(ctx context.Context, opts []ProcessTreeOption, tree ...*Proc
...
+
+		ios, err := kamino.Clone(iostreams.G(ictx),
+			kamino.WithZeroUnexported(),
+		)
+		if err != nil {
+			return err
+		}
+
+		ios.Out = iostreams.NewNoTTYWriter(item, iostreams.G(ctx).Out.Fd())
+		ios.In = iostreams.G(ctx).In
```
craciunoiuc: What abous `ios.Err`?

tui/paraprogress/paraprogress.go
```diff
@@ -64,32 +65,57 @@ func NewParaProgress(ctx context.Context, processes []*Process, opts ...ParaProg
...
+
+			ios, err := kamino.Clone(iostreams.G(pctx),
+				kamino.WithZeroUnexported(),
+			)
+			if err != nil {
+				return nil, err
+			}
+
+			ios.Out = iostreams.NewNoTTYWriter(process, iostreams.G(ctx).Out.Fd())
+			ios.In = iostreams.G(ctx).In
```
craciunoiuc: Same here

tui/processtree/processtree.go
nderjung: Makes sense. This hasn't affected my testing so far and adding it hasn't caused any issues, so I've now forced pushed with an updated.

craciunoiuc (approved):
All good here. Thanks!

PR #1229, opened 2024-02-07
fix(logtail): Correctly find null prefix

LucaSeri:
Because we previously discarded all contents to the last null character, log contents such as:```Useful information\x00``` would cause the information to be discarded. Instead, look for the first non-null character.

### Prerequisite checklist


nderjung (approved):
Thanks!

PR #1230, opened 2024-02-09
fix(gendocs): Remove old link prefixes

craciunoiuc:
### Prerequisite checklist

### Description of changes

@nderjung this might not be universally correct, but keep it in mind when deploying docs. After an initial scan it looks fine.

### DIFF
--- a/tools/gendocs/main.go
+++ b/tools/gendocs/main.go
@@ -83,6 +83,14 @@ func generateMarkdown(cmd *cobra.Command, dir string) error {
 		return err
 	}
 
+	if hasAliases(cmd) {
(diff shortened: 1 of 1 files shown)


nderjung:
Can you please add a CLI arg that provides the user the option to specify the URL prefix? This allows the docs to be placed in different places whilst simultaneously preventing this type of fix in the future.

The flag should be named `-url-prefix` and should be propagated as a GNU Make variable, `DOCS_URL_PREFIX`, for the `gendocs` target.  The default value should be set to `/docs/cli/reference/`.

craciunoiuc:
This is no longer needed and was fixed together with the new docs

PR #1241, opened 2024-02-12
fix(kraft): Handle tailing logs from fast-exiting kernels

nderjung:
### Prerequisite checklist

### Description of changes

This PR handles an issue where fast-exiting run-to-completion-type unikernels (specifically those emited via the QEMU VMM) would result in the inability to render the available logs since the connectivity to the machine's event stream would be inaccessible (due to the exit).

To remedy this, a pre-emptive static check for the state of the machine is performed if access to the stream is not possible.  This static check determines whether the context is cancelled in all execution paths which would have otherwise prevented the logs from the machine to be properly handled.


craciunoiuc (approved):
All good here. Thanks!

PR #1313, opened 2024-02-28
feat(create): Automatically assign subnet

LucaSeri:
### Prerequisite checklist

### Description of changes

This `PR` makes the ``--network`` parameter of ``kraft net create`` optional.
Finding a new subnet is achieved by searching through a predefined pool which is also [the one used in Docker](https://github.com/moby/libnetwork/blob/master/ipamutils/utils.go#L18-L20).

Throughout this pool, we search for an "aligned" subnet that does not intersect any existing subnetworks.

This could be further improved by allowing users to define their own default pool within the kraft config.

### DIFF
--- a/internal/cli/kraft/net/create/create.go
+++ b/internal/cli/kraft/net/create/create.go
@@ -59,19 +59,6 @@ func NewCmd() *cobra.Command {
 func (opts *CreateOptions) Pre(cmd *cobra.Command, _ []string) error {
 	opts.Driver = cmd.Flag("driver").Value.String()
 
-	// TODO(nderjung): A future implementation can list existing networks and
-	// generate new subnet and gateway appropriately.  Simply calculate a new
-	// subnet which is out of bounds from all existing subnets for the given
-	// driver.  Additionally, a gateway can be determined by selecting the first
-	// allocatable IP of a given subnet, i.e. 0.0.0.1.  The subnet and gateway
-	// can therefore be expressed in the form X.Y.Z.1/N.
-	// if opts.Subnet == "" {
-	// 	return fmt.Errorf("cannot create network without subnet")
-	// }
-	if opts.Network == "" {
-		return fmt.Errorf("cannot create network without gateway and subnet in CIDR format")
-	}
-
 	return nil
 }
 
@@ -86,6 +73,20 @@ func (opts *CreateOptions) Run(ctx context.Context, args []string) error {
 	controller, err := strategy.NewNetworkV1alpha1(ctx)
 	if err != nil {
 		return err
+	}
+
+	if opts.Network == "" {
+		existingNetworks, err := controller.List(ctx, &networkapi.NetworkList{})
+		if err != nil {
+			return err
+		}
+
+		freeNetwork, err := network.FindFreeNetwork(network.DefaultNetworkPool, existingNetworks)
+		if err != nil {
+			return err
+		}
+
+		opts.Network = freeNetwork.String()
 	}
 
 	addr, err := netlink.ParseAddr(opts.Network)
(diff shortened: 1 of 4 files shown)


internal/cli/kraft/net/create/create.go
```diff
@@ -109,3 +104,102 @@ func (opts *CreateOptions) Run(ctx context.Context, args []string) error {
...
+
+	defaultNetworkPool := []networkPoolEntry{
+		{"172.17.0.0/16", 16},
+		{"172.18.0.0/16", 16},
+		{"172.19.0.0/16", 16},
+		{"172.20.0.0/14", 16},
+		{"172.24.0.0/14", 16},
+		{"172.28.0.0/14", 16},
+		{"192.168.0.0/16", 20},
+	}
```
nderjung: This should probably be a globally exported variable coming out of `machine/network`

nderjung: Similarly, the `findFreeNetwork` could be a general-purpose method which accepts an input set of networks and an available pool of networks.

Finally, the `net.Interfaces()` listing (input networks) should be specific to the underlying driver.

internal/cli/kraft/net/create/create.go
```diff
@@ -109,3 +104,102 @@ func (opts *CreateOptions) Run(ctx context.Context, args []string) error {
...
+			subnetIndex = subnetIndex + (1 << (32 - poolEntry.Size))
+		}
+	}
+
+	return "", fmt.Errorf("unable to find a free network in the default pool")
+}
+
+func networksIntersect(a, b *net.IPNet) bool {
+	return a.Contains(b.IP) || b.Contains(a.IP)
+}
```
nderjung: Move to `machine/networks/utils.go` and export.

internal/cli/kraft/net/create/create.go
LucaSeri: Thanks for the review @nderjung! Can you please elaborate a bit on your last point:
> Finally, the net.Interfaces() listing (input networks) should be specific to the underlying driver.

Why is this the case? My idea was that listing all the interfaces on the host, we would be able to identify all subnetworks that are already taken, which we should avoid regardless of the driver we are using for the newly created network.

nderjung: Go's `net.Interfaces()` method is a general-purpose method for returning all interfaces from the host (and has OS-specific implementations) and therefore not necessarily ones under control by the driver.  We can a). shortcut having to query the host entirely by referencing the pool by the driver, making this implementation OS-agnostic and therefore driver-dependent and b). guarantee that a generated "free" IP comes from the driver and therefore guaranteed to work under its supervision.

Ultimately, it's driver which should be responsible for announcing the available pool of networks.  One way tackle this could be to have an IPAM pool which is specified as part of the `NetworkSpec`.  This gives the user the ability to customize the available IPs but also the underlying driver implementation to return the available set.

Check out how CNI is doing something like this for some ideas: https://www.cni.dev/plugins/current/ipam/host-local/

Ultimately, this configuration would be either a). passed in by the user in some form (because the underlying driver could be interfaced with and accept a starting configuration) and b). potentially manipulated later by the external driver implementation.  For example, the `NetworkService.Get` method may return a modified pool range because it's invoking an external driver program which is outputting this information; the driver is external and therefore its IPAM range is not managed by KraftKit becauase it is configured externally.

nderjung: In fact this probably means that there is an input `NetworkSpec` IPAM address pool and a `NetworkStatus` with current available pool, the latter of which is used to derive a free IP.

nderjung (approved):
Great! Thanks!

PR #1327, opened 2024-02-29
perf: simplify digest sort

antoineco:
OCI image digests (sha256 hex values) are unlikely to be parseable as int, so the extra conditional and strings conversions are slowing down the comparison.


craciunoiuc (approved):
All good here. Thanks!

PR #1336, opened 2024-03-01
feat(workflows): Add webinstall workflow to push image

craciunoiuc:
### Prerequisite checklist

### Description of changes

This saves some minutes having to do it by hand every release. Restart is still manual for now.

### DIFF
--- a/.github/workflows/webinstall.yaml
+++ b/.github/workflows/webinstall.yaml
@@ -1,0 +1,31 @@
+name: tools/webinstall
+
+on:
+  push:
+    branches: [stable]
+    paths:
+      - 'tools/webinstall/**'
+
+jobs:
+  myself-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{github.actor}}
+          password: ${{secrets.GITHUB_TOKEN}}
+
+      - name: Build & Push OCI Image
+        uses: docker/build-push-action@v5
+        with:
+          push: true
+          tags: ghcr.io/unikraft/kraftkit/webinstall:latest
+          platforms: linux/amd64
+          file: ./tool/webinstall/Dockerfile
+          secrets: |
+            GIT_AUTH_TOKEN=${{ secrets.GITHUB_TOKEN }}


.github/workflows/webinstall.yaml
```diff
@@ -0,0 +1,33 @@
...
+          registry: index.unikraft.io
+          username: ${{ secrets.REG_USERNAME }}
+          password: ${{ secrets.REG_TOKEN }}
+
+      - name: Build & Push OCI Image
+        uses: docker/build-push-action@v5
+        if: ${{ github.event_name == 'push' }}
+        with:
+          push: true
+          tags: index.unikraft.io/kraftkit-internal/webinstall:latest
```
antoineco: Better avoid pushing to Harbor if we can. Maybe this instead? https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

.github/workflows/webinstall.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  myself-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to Registry
+        if: ${{ github.event_name == 'push' }}
```
antoineco: Nit-pick: the workflow is only configured to run on "push".

.github/workflows/webinstall.yaml
craciunoiuc: why? This is something I anyway do manually

antoineco: Because when we upgrade Kubernetes and Harbor is down, no application can be deployed and we have a 
downtime. (chicken and egg issue)

.github/workflows/webinstall.yaml
```diff
@@ -0,0 +1,33 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: index.unikraft.io
+          username: ${{ secrets.REG_USERNAME }}
+          password: ${{ secrets.REG_TOKEN }}
+
+      - name: Build & Push OCI Image
+        uses: docker/build-push-action@v5
+        if: ${{ github.event_name == 'push' }}
```
antoineco: Same as previous comment

craciunoiuc: It's working very badly, this is fixed.

.github/workflows/webinstall.yaml
craciunoiuc: Yes, makes sense, switched to `ghcr`

antoineco (approved):

PR #1337, opened 2024-03-01
fix(processtree): Correct logic to determine when to show logs

nderjung:
The effective if-statement incorrectly hid the logs when `hideError` was set to true.


craciunoiuc (approved):
All good here. Thanks!

PR #1338, opened 2024-03-01
feat(initrd): Dynamically instantiate BuildKit in container

nderjung:
This PR introduces the ability to dynamically start BuildKit using the [testcontainer framework][0] which provides an abstraction over, at least at the time of PR, Docker, Docker Desktop and Podman.

Its ability to interface with the host and start a container based on these providers may able to provide to the vast set of users an ease-of-use improvement which would otherwise have resulted in the Dockerfile initrd provider to spit a warning, requiring additional user interaction.  This has caused many users needless hassle and so this PR aims to solve this by dynamically starting BuildKit and connecting to it.

Since the container may still linger from quick successions of `kraft` invocations, dynamically allocate a port for BuildKit between 4000-5000.

[0]: https://testcontainers.com/

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,7 @@ require (
 	github.com/google/go-containerregistry v0.19.0
 	github.com/google/go-github/v32 v32.1.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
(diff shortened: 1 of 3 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #1348, opened 2024-03-03
fix(run): Improve filter and selection process of potential packages

nderjung:
This PR partially reworks the filtering of possible runtime packages from the given input for the Kraftfile-with-runtime context.  In this rework;

a). the initial query to the `Catalog` is wrapped in a process tree which hides the search logs from the catalog, reduding the visible output in "fancy-log" mode;
b). uses the provided CLI flags for platform and architecture later in the query process such that the option of selecting a package with alternative platform/architecture is possible.

Ultimately this PR improves the experience of running and selecting a possible runtime following a `kraft run` invocation.


craciunoiuc (approved):
All good here. Thanks!

PR #1355, opened 2024-03-03
feat(kraft): Warn if the user has invoked kraft via sudo

nderjung:
This PR is dependent on the docs of installation process being updated to incorporate adjustments to allow for sudoless use of `kraft`.

### DIFF
--- a/internal/cli/kraft/kraft.go
+++ b/internal/cli/kraft/kraft.go
@@ -163,6 +163,20 @@ func Main(args []string) int {
 		ctx = iostreams.WithIOStreams(ctx, copts.IOStreams)
 	}
 
+	if (os.Getenv("SUDO_UID") != "" || os.Getenv("SUDO_GID") != "" || os.Getenv("SUDO_USER") != "") && !config.G[config.KraftKit](ctx).NoWarnSudo {
+		log.G(ctx).Warn("detected invocation via sudo!")
+		log.G(ctx).Warn("")
+		log.G(ctx).Warn("mixing invocations of kraft with sudo can lead to unexpected behavior")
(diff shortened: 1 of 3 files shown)


craciunoiuc:
This should be good to go right?

nderjung:
Tests are failing. Not sure why.

craciunoiuc:
It means it worked :D

Tests for `kraft net` are invoked with sudo, so the function `NewKraftKitPriviledged` (or smth like that) needs updating to have the suppress flag in it.

Because of the disclaimer parsing fails.

nderjung:
But I set `KRAFTKIT_NO_WARN_SUDO=1` in the test, to disable showing the message and it still shows.

craciunoiuc:
Maybe the env is only passed to `sudo` in the current format? Can you change stuff a bit to also allow a `--no-warn-sudo` flag 🙏

internal/cli/kraft/kraft.go
```diff
@@ -163,6 +163,20 @@ func Main(args []string) int {
 		ctx = iostreams.WithIOStreams(ctx, copts.IOStreams)
 	}
 
+	if (os.Getenv("SUDO_UID") != "" || os.Getenv("SUDO_GID") != "" || os.Getenv("SUDO_USER") != "") && os.Getenv("KRAFTKIT_NO_WARN_SUDO") == "" {
```
craciunoiuc: can you switch this to a config kraftkit flag so we can then also use it as a flag `--no-warn-sudo`?

craciunoiuc (approved):
All good here. Thanks!

PR #1250, opened 2024-02-14
feat: Add netdevip fields and allow multiple networks

LucaSeri:
### Prerequisite checklist

  - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
  - [x] Tested your changes against relevant architectures and platforms;
  - [x] Ran `make fmt` on your commit series before opening this PR;
  - [ ] Updated relevant documentation.

### Description of changes

This PR:
- introduces changes to interfaces to accommodate the new `netdev` syntax introduces in [0.16.1](https://unikraft.org/blog/2024-01-18-unikraft-releases-v0.16.1)
- allows specifying multiple networks when running a unikernel
- adds some fixes for `NetworkIterator`

PR #1368, opened 2024-03-05
fix: Remove dangling interfaces

LucaSeri:
### Prerequisite checklist

### Description of changes

Previously, repeatedly running a unikernel:
```bash
kraft run --rm --network some-net unikraft.org/nginx:1.15
```
would cause interfaces to pile up:

```bash
some-net@if0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc fq_codel master some-net state DOWN group default qlen 1000
some-net@if1: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc fq_codel master some-net state DOWN 
some-net@if2: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc fq_codel master some-net state DOWN group default qlen 1000
```

This `PR` reverts a previous commit https://github.com/unikraft/kraftkit/pull/1250/commits/49f248ac4f65a54f576b23098d2f82422886d6d3, by instead allowing a network update to only remove the interfaces that belong to that specific network.

Also, it fixes a bug in start that would cause removing interfaces from unikernels with multiple networks to be wrong.

### DIFF
--- a/internal/cli/kraft/start/start.go
+++ b/internal/cli/kraft/start/start.go
@@ -233,7 +233,7 @@ func Start(ctx context.Context, opts *StartOptions, machineNames ...string) erro
 
 				// Remove the new network interface
 				for i, iface := range found.Spec.Interfaces {
-					if iface.UID == machine.Spec.Networks[0].Interfaces[0].UID {
+					if iface.UID == network.Interfaces[0].UID {
 						ret := make([]networkapi.NetworkInterfaceTemplateSpec, 0)
 						ret = append(ret, found.Spec.Interfaces[:i]...)
 						found.Spec.Interfaces = append(ret, found.Spec.Interfaces[i+1:]...)
--- a/machine/network/bridge/v1alpha1.go
+++ b/machine/network/bridge/v1alpha1.go
@@ -336,6 +336,31 @@ func (service *v1alpha1Network) Update(ctx context.Context, network *networkv1al
 		network.Spec.Interfaces[i] = iface
 	}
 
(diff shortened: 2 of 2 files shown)


craciunoiuc:
image


Did not work for me :(

Created a network and then ran some unikernels (with firecracker, if it matters), and then there were still dangling networks

LucaSeri:
Hi @craciunoiuc, thanks for checking. Did you run with the ``--rm`` flag?

I think this was introduced at the same time as ``kraft start`` - an exited kernel can be restarted, so the interface is kept (it is also visible in ``kraft net inspect``). 
Therefore, there are sort of 3 cases:
* A running kernel keeps its interface up
* An exited kernel keeps its interface down -> I think this is intended?
* A removed kernel should not keep any interfaces -> This is what this PR handles (previously, even when removing the machine it would still keep its interface).

Could you check if that is also the case for you?

craciunoiuc:
Whoops, I did not :D, retrying

Edit: after running ` sudo kraft net create -n 172.47.0.1/24 kraft2`
ran `sudo kraft run --plat fc --arch x86_64 -M 512M --rm --network kraft2 .` interfaces were still there
ran `sudo kraft rm --all`. Interfaces were still there
ran `sudo kraft net rm kraft2`, bridge was gone but they were still there (this did not happen when not removing)

nderjung (approved):
Tried this out and it works for me on Linux (including your commands @craciunoiuc ;-) ). Thanks!

PR #1369, opened 2024-03-05
fix(oci): Correctly handle "fullrefs" without an identifier

nderjung:
This commit is an addendum to b73dca8088 which mistakenly did not include all correct expansions of the full reference with included identifier (or tag).


craciunoiuc (approved):
All good here. Thanks!

PR #1372, opened 2024-03-05
fix(machine): Properly display instance memory

### LINKED ISSUE #1371: kraft ps reports 0M when not using the -M flag for kraft run
### Describe the bug

kraft ps reports 0M when not using the -M flag for kraft run

### Steps to reproduce

run kraft run w/o the -M flag
when finished do kraft ps, look at MEM field, should say 0MiB

### Expected behavior

kraft ps should display whatever the default M value set by kraft is.

x86_64

macOS

### Relevant log output

```shell
➜  http-rust1.75-actix-web4 git:(main) ✗ kraft ps
NAME              KERNEL                          ARGS     CREATED         STATUS   MEM   PORTS                   PLAT
admiring_ndakasi  oci://unikraft.org/base:latest  /server  22 minutes ago  running  0MiB  0.0.0.0:8080->8080/tcp  qemu/x86_64
```

craciunoiuc:
### Prerequisite checklist

### Description of changes

GitHub-Fixes: #1371

### DIFF
--- a/internal/cli/kraft/ps/ps.go
+++ b/internal/cli/kraft/ps/ps.go
@@ -177,7 +177,7 @@ func (opts *PsOptions) PsTable(ctx context.Context) ([]PsEntry, error) {
 			Args:    strings.Join(machine.Spec.ApplicationArgs, " "),
 			Kernel:  machine.Spec.Kernel,
 			State:   machine.Status.State,
-			Mem:     fmt.Sprintf("%dMiB", machine.Spec.Resources.Requests.Memory().Value()/MemoryMiB),
+			Mem:     machine.Spec.Resources.Requests.Memory().String(),
 			Created: humanize.Time(machine.ObjectMeta.CreationTimestamp.Time),
 			Ports:   machine.Spec.Ports.String(),
 			Arch:    machine.Spec.Architecture,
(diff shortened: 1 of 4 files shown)


nderjung:
For future reference: the [`resource.Quantity`](https://pkg.go.dev/k8s.io/apimachinery@v0.29.2/pkg/api/resource#Quantity) structure which is the inner-most structure that contains the value of the memory has unexported fields which are not accessible by the `encoding/gob` package during state serialization.  As a result, upon deserialization, these values are set to zero which translates to the user-level view of `0MiB`.

This solution represents a compromise by on-the-fly re-populating these values from the platform's configuration structure which has this information saved.

One of possible future solutions include:
- Performing JSON serialization (as opposed to binary serialization) of machine state.  This would require all QEMU structures implement relevant JSON (Un)Marshalling methods (possible, but non-zero effort);
- The underlying `resource.Quantity` structure provides upstream binary encoding (unlikely);
- The underlying `resource.Qunatity` structure is replaced for another (third-party or in-house); at this time, no good alternative exists since the `resource` package contains a lot of very useful and Kubernetes-complaint methods and structures exist.  It would ultimately be a shame to replace this (possible, but non-zero effort).

In my view, JSON serialization makes the most sense since it is under our jurisdiction and compatible with external structures.  The compromise may be performance but there is an increase in user-level utility by saving state in JSON, making machine state easier to debug and parse with external tools.

nderjung (approved):
This works as intended, thanks!

PR #1374, opened 2024-03-06
feat(cloud): update to go-sdk 0.5

antoineco:
Updates KraftKit to use the new SDK version 0.5

### DIFF
--- a/internal/cli/kraft/cloud/instance/list/list.go
+++ b/internal/cli/kraft/cloud/instance/list/list.go
@@ -68,9 +68,21 @@ func (opts *ListOptions) Run(ctx context.Context, args []string) error {
 		kraftcloud.WithToken(config.GetKraftCloudTokenAuthConfig(*auth)),
 	)
 
-	instances, err := client.WithMetro(opts.metro).List(ctx)
+	instListResp, err := client.WithMetro(opts.metro).List(ctx)
 	if err != nil {
 		return fmt.Errorf("could not list instances: %w", err)
+	}
+	if len(instListResp) == 0 {
+		return nil
+	}
+
+	uuids := make([]string, 0, len(instListResp))
+	for _, instItem := range instListResp {
+		uuids = append(uuids, instItem.UUID)
+	}
+	instances, err := client.WithMetro(opts.metro).GetByUUIDs(ctx, uuids...)
+	if err != nil {
+		return fmt.Errorf("getting details of %d instance(s): %w", len(instListResp), err)
 	}
 
 	return utils.PrintInstances(ctx, opts.Output, instances...)
--- a/internal/cli/kraft/cloud/scale/initialize/initialize.go
+++ b/internal/cli/kraft/cloud/scale/initialize/initialize.go
@@ -14,8 +14,8 @@ import (
 	"github.com/spf13/cobra"
 
 	kraftcloud "sdk.kraft.cloud"
-	kraftcloudinstances "sdk.kraft.cloud/instances"
-	kraftcloudautoscale "sdk.kraft.cloud/services/autoscale"
+	kcinstances "sdk.kraft.cloud/instances"
+	kcautoscale "sdk.kraft.cloud/services/autoscale"
 
 	"kraftkit.sh/cmdfactory"
 	"kraftkit.sh/config"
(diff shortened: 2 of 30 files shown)


internal/cli/kraft/cloud/instance/list/list.go
```diff
@@ -68,10 +69,19 @@ func (opts *ListOptions) Run(ctx context.Context, args []string) error {
...
 	if err != nil {
 		return fmt.Errorf("could not list instances: %w", err)
 	}
 
-	return utils.PrintInstances(ctx, opts.Output, instances...)
+	var is []kcinstances.GetResponseItem
+	for _, instItem := range instListResp {
+		i, err := client.WithMetro(opts.metro).GetByUUID(ctx, instItem.UUID)
+		if err != nil {
+			return fmt.Errorf("gettings details of instance %s: %w", instItem.UUID, err)
```
craciunoiuc:
```suggestion
			return fmt.Errorf("getting details of instance %s: %w", instItem.UUID, err)
```

No typo is safe

internal/cli/kraft/cloud/scale/initialize/initialize.go
```diff
@@ -28,9 +28,9 @@ type InitOptions struct {
 	Client       kraftcloud.KraftCloud `noattribute:"true"`
 	CooldownTime time.Duration         `long:"cooldown-time" short:"c" usage:"The cooldown time of the config (ms/s/m/h)" default:"1000000000"`
 	Master       string                `long:"master" short:"i" usage:"The UUID or Name of the master instance"`
-	MaxSize      uint                  `long:"max-size" short:"M" usage:"The maximum size of the configuration" default:"10"`
+	MaxSize      int                   `long:"max-size" short:"M" usage:"The maximum size of the configuration" default:"10"`
 	Metro        string                `noattribute:"true"`
-	MinSize      uint                  `long:"min-size" short:"m" usage:"The minimum size of the configuration"`
+	MinSize      int                   `long:"min-size" short:"m" usage:"The minimum size of the configuration"`
```
craciunoiuc, line 33: Not sure about the platform implementation, but wouldn't it be better to have this as `uint` to catch user errors?

Min/Max are both >0 and if we use `uint` cobra can catch (I think) if someone used a negative value.

What do you think? I'm fine either way

internal/cli/kraft/cloud/scale/initialize/initialize.go
```diff
@@ -109,80 +109,91 @@ func (opts *InitOptions) Run(ctx context.Context, args []string) error {
...
 					continue
 				}
-
-				if instance.ServiceGroup.UUID != args[0] {
+				if i.ServiceGroup.UUID != args[0] {
 					continue
 				}
 
-				possible = append(possible, instance)
+				possible = append(possible, stringerInstance{i})
```
craciunoiuc: This got me properly confused, what Go magic is this?

Not even sure what a `stringerInstance` is

internal/cli/kraft/cloud/service/create/create.go
```diff
@@ -58,7 +58,7 @@ func Create(ctx context.Context, opts *CreateOptions, args ...string) (*kraftclo
 		)
 	}
 
-	services := []kraftcloudservices.Service{}
+	services := make([]kcservices.CreateRequestService, 0, 2)
```
craciunoiuc: Is this correct? A service can have more than 2 open ports

craciunoiuc (commented):
Some more comments from my side

internal/cli/kraft/cloud/scale/initialize/initialize.go
antoineco: We can yes, I don't have strong opinions for/against it.

The reason why I changed `uint` to `int` in the SDK is that the API accepts `int` (to be clarified), and I would like the validation to remain as much as possible **on the API side** to avoid drifts in client code.

internal/cli/kraft/cloud/scale/initialize/initialize.go
antoineco: This type is at the bottom of the file.
KraftKit requires types to be `fmt.Stringer` to be able to perform the interactive selection, but the SDK has no reason to expose stringer types (what does it mean for an API response to be `fmt.Stringer`, what does it print? All fields, some fields?).

Since Go interfaces can be implemented close to where they are actually required, we simply use a mini wrapper type which implements `fmt.Stringer` around the API response, and voilà: KraftKit works without touching the SDK.

craciunoiuc: Whoa, really cool

internal/cli/kraft/cloud/service/create/create.go
antoineco: It's a naive pre-allocation for the case where the user did not provide service group details. `append` will automatically grow the array if we add more than 2 elements.

As far as I can tell this is the only case where we can pre-allocate without parsing. However we could also just perform allocations within each `if` case.

antoineco: I split the allocations into the two cases in [da0715e8..98db95e3](https://github.com/unikraft/kraftkit/compare/da0715e8aa67c125930667f42ef9633f0cb295ed..98db95e38fed65aa0fbfbaf7fddd20f5f08d35cd). I agree that this feels clearer.

internal/cli/kraft/cloud/instance/list/list.go
antoineco: Laser eyes! Done in [da0715e8..98db95e3](https://github.com/unikraft/kraftkit/compare/da0715e8aa67c125930667f42ef9633f0cb295ed..98db95e38fed65aa0fbfbaf7fddd20f5f08d35cd)

craciunoiuc (approved):
All good here. Thanks!

PR #1383, opened 2024-03-07
fix(cloud): write to nil Env map

antoineco:
At the time we try to write to `req.Env`, the `map[string]string` is still nil.


nderjung (approved):
Thanks!

PR #1384, opened 2024-03-07
fix(rm): Check all machines regardless of status

LucaSeri:
### Prerequisite checklist

### Description of changes
Machines keep references to their interfaces even in different states. For example, an exited machine maintains its interface DOWN and it is set UP again when using ``kraft start`` on that particular machine. 

Also check for those interfaces when removing network.


nderjung (approved):
Thanks!

Issue #1207, opened 2024-01-30
Add `-o raw` option for KraftCloud commands


marcrittinghaus:
### Feature request summary

For Kraftcloud development and debugging as well as to be able to easily retrieve raw information from the server, the `kraft cloud` commands should support a `-o raw` option to output 1:1 the server's JSON response.

This is needed also because `kraft` modifies and hides information even with the `-o json` option. For example, the `created at` field just reports `X days ago` instead of the exact timestamp. Another example, the service group name of an instance is not reported. 

Furthermore, it makes it easy to access new information as soon as they become available from the server, without the need to wait for `kraft` integration.

### Related architectures

None

### Related platforms

None

antoineco:
For information the work on the Go SDK is happening here: https://github.com/kraftcloud/go-sdk/pull/22

This will ensure that `-o json` returns payloads which are faithful to what is documented at [docs.kraft.cloud](https://docs.kraft.cloud/api/v1/). This is however only one part of the story, as enabling a true raw mode will require bypassing current SDK methods.

antoineco:
The last part of the changes needed for the `-o raw` flag will be implemented as soon as https://github.com/kraftcloud/go-sdk/pull/26 makes it into the SDK.

antoineco:
Second large change to the SDK that supports this feature: https://github.com/kraftcloud/go-sdk/pull/26

PR #1385, opened 2024-03-08
feat(cloud): print API response items with `-o=json`

antoineco:
Related to #1207

Note that this prints arrays of semi-raw _response items_[^1] but not the true _raw_ API responses. There is still work to do in the Go SDK to be able to print _true_ raw responses.

[^1]: "semi" because a) the order of attributes may be shuffled and b) some attributes that were not part of the response may instead be printed as _null_


craciunoiuc (approved):
All good here. Thanks!

PR #1402, opened 2024-03-13
feat(logs): Log multiple machines at once

LucaSeri:
### Prerequisite checklist

### Description of changes

Allows ``kraft logs`` to log multiple machines at once.
The ``logs`` module takes care of synchronization internally, so it can be directly called into by other command (e.g. ``start``).
The ``prefix-name`` and ``prefix`` arguments were removed in favor of the following intuitive behavior:
* When logging a single machine, don't prefix the name
* When logging multiple machines, prefix each line with the name of the machine that produced it (if the user doesn't opt out of this via ``no-prefix``


nderjung (approved):
Great! Thanks! This works as expected.

PR #1425, opened 2024-03-18
fix(internal): Require at least 1Mi for running

### LINKED ISSUE #1382: If user forgets to add "M" to the -M flag, kraft silently uses default memory amount
### Describe the bug

If user forgets to add "M" to the -M flag, kraft silently uses defaults memory amount, which can cause the app to fail without giving the user much indication as to why.

### Steps to reproduce

Run something like:

kraft run --rm -p 1313:1313 --plat qemu --arch x86_64 -M 512 # NOTE: no "M" after 512

on any app that requires more than the default amount of memory (which I think is 64M)

### Expected behavior

kraft should probably give an error, saying that the unit is missing, and stopping, rather than silently continuing, applying (silently again) the memory default, and then potentially failing if the app needs memory, without the user understanding why.

x86_64

macOS

### Relevant log output

```shell
% kraft run --rm -p 1313:1313 --plat qemu --arch x86_64 -M 512
 W  using hardware emulation
 i  using arch=x86_64 plat=qemu
[+] pulling unikraft.org/hugo:latest (qemu/x86_64) •••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••• 100% [1m 46s]
[+] building rootfs... done!                                                                                                                                    x86_64 [3.1s]
 E  machine fatally exited
```

craciunoiuc:
Lower values would get divided by Mi and turned to 0, which defaulted to 64Mi.

GitHub-Fixes: #1382

### Prerequisite checklist

### DIFF
--- a/internal/cli/kraft/run/run.go
+++ b/internal/cli/kraft/run/run.go
@@ -26,7 +26,6 @@ import (
 	mplatform "kraftkit.sh/machine/platform"
 	"kraftkit.sh/packmanager"
 	"kraftkit.sh/tui/selection"
-	"kraftkit.sh/unikraft/arch"
 	ukarch "kraftkit.sh/unikraft/arch"
 )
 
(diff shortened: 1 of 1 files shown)


internal/cli/kraft/run/run.go
```diff
@@ -338,6 +348,7 @@ func (opts *RunOptions) Run(ctx context.Context, args []string) error {
 		if err != nil {
 			return err
 		}
+		fmt.Printf("quantity: %v\n", quantity)
```
nderjung: Lingering

nderjung (approved):
Thanks!

PR #1438, opened 2024-03-22
feat(internal): Enable experimental rootfs compression

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/initrd/directory.go
+++ b/initrd/directory.go
@@ -5,8 +5,10 @@
 package initrd
 
 import (
+	"compress/gzip"
 	"context"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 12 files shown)


initrd/directory.go
```diff
@@ -155,6 +157,39 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
...
 	}
 
+	if initrd.opts.compress {
+		writer.Close()
+		f.Close()
+
+		fr, err := os.OpenFile(initrd.opts.output, os.O_RDONLY, 0o644)
+		if err != nil {
+			return "", fmt.Errorf("could not open initramfs file: %w", err)
+		}
```
antoineco: Instead of doing this you could just [Seek](https://pkg.go.dev/os#File.Seek) back to the beginning of the file with `f.Seek(io.SeekStart, 0)`.

antoineco: (unless you are really worried about `os.O_RDWR` being misused by `gzip.NewWriter`)

antoineco (approved):
Approving with one non-blocking comment.

initrd/directory.go
craciunoiuc: hmm, makes sense, I can do that yes. I was more worried about if the data is fully present there (hence the closes before)

craciunoiuc: did, seems to work fine 👍

PR #1439, opened 2024-03-22
feat: new service domains attribute for multi-dns

antoineco:
### Description of changes

Adds support for the new multi-DNS feature which, more specifically, adds a new `domains` array to responses from the `instances` and `services` services.

### Out of scope

- Changing the command line API to allow people to create instances or services with multiple domains.
- Changing `get/ls` outputs to include multiple domains and/or certificate information.
- Changing the output of `kraft cloud deploy` to include multiple domains and/or certificate information.

We should do these as follow ups. Let's support API changes first, and add features later.


craciunoiuc (approved):
All good here. Thanks!

PR #1454, opened 2024-03-26
fix(run): Don't erase existing volumes when passing volume arg

### LINKED ISSUE #1340: Passing volumes to kraft run via -v overwrites ``rootfs`` and ``volumes`` from ``Kraftfile``
### Describe the bug

When working with an application that has some ``rootfs``/``volumes`` as part of the ``kraftfile``, attempting to pass additional volumes via the cli directly simply overrides the ``kraftfile`` as opposed to being added on top, resulting in very unintuitive behaviour.

### Steps to reproduce

```yaml
spec: v0.6

runtime: base-9p:latest

rootfs: ./Dockerfile

volumes:
  - ./data:/data

cmd: ["/server"]
```
When issuing ``kraft run``, the generated argument for ``qemu`` is as expected:
```bash
 vfs.fstab=[ "initrd0:/:extract:::" "fs2:/data:9pfs:::mkmp" ]
```

When issuing ``kraft run -v ./cli-vol:/cli-vol``, the generated argument for ``qemu`` only contains the cli passed volume:
```bash
vfs.fstab=[ "fs1:/cli-vol:9pfs:::mkmp" ]
```

### Expected behavior

For the volume to be added on top of the ``Kraftfile``

_No response_

_No response_

### Relevant log output

_No response_

LucaSeri:
### Prerequisite checklist

### Description of changes

Take a simple `Kraftfile` example:
```yaml
spec: v0.6

runtime: base-9p:latest

rootfs: ./Dockerfile
 
cmd: ["/server"]
```

Here, we would like to be able to run this application both with the default rootfs:
```
kraft run
```
or maybe we want to sometimes provide a volume on top of it:
```
kraft run --volume ./some-path:/other-path
```

This PR ensures the fact that the ``volume`` cli argument is applied on top of the config from the ``Kraftfile``.

Github-fixes: https://github.com/unikraft/kraftkit/issues/1340

### DIFF
--- a/internal/cli/kraft/run/utils.go
+++ b/internal/cli/kraft/run/utils.go
@@ -189,8 +189,10 @@ func (opts *RunOptions) parseVolumes(ctx context.Context, machine *machineapi.Ma
 
 	var err error
 	controllers := map[string]volumeapi.VolumeService{}
-	machine.Spec.Volumes = []volumeapi.Volume{}
 
+	if machine.Spec.Volumes == nil {
+		machine.Spec.Volumes = make([]volumeapi.Volume, 0)
+	}
 	for _, volLine := range opts.Volumes {
 		var hostPath, mountPath string
 		split := strings.Split(volLine, ":")


nderjung (approved):
Thanks!

PR #1473, opened 2024-03-28
fix(webinstall): Disable completion prompt

whiteadam:
### Prerequisite checklist

### Description of changes

**Problem:** When passing -y to disable confirmation prompts, the user was still prompted to select a language for shell completions.

**Solution:** Added logic to skip `install_completions` function when `-y` is used on installer script.

### DIFF
--- a/tools/webinstall/install.sh
+++ b/tools/webinstall/install.sh
@@ -1713,7 +1713,9 @@ main() {
     _main_kraft_ret="$?"
 
     # Install kraftkit completions
-    install_completions "$_main_arch"
+    if [ "$NEED_TTY" = "y" ]; then
+        install_completions "$_main_arch"
+    fi
 
     return $_main_kraft_ret
 }


tools/webinstall/install.sh
```diff
@@ -1713,7 +1713,9 @@ main() {
     _main_kraft_ret="$?"
 
     # Install kraftkit completions
-    install_completions "$_main_arch"
+    if [ "$NEED_TTY" != "n" ]; then
```
craciunoiuc:
```suggestion
    if [ "$NEED_TTY" = "y" ]; then
```

Let's invert the check to make sure that we respect the default value `y`.

We need to do the check here because there is no default value for the shell type.

craciunoiuc (changes requested):
Looks good, thanks for spotting this. I have one suggestion to match the cases where someone writes `$NEED_TTY="no"` (for example)

nderjung:
Hi @whiteadam, could you please fixup/rebase your commits into a single commit and force-push? Thanks!

whiteadam:
Sorry @nderjung, it's been a while since I did proper public github PRs 😄

craciunoiuc:
oh no, now the commit is mangled:

image

can you leave it like:

```
fix(webinstall): Disable completion prompt

Signed-off-by: Adam White <whiteadam@users.noreply.github.com>
```

whiteadam:
Ah, I am hitting all of the wrong buttons today. I tried using the UI instead of CLI and look where it's getting me.

craciunoiuc (approved):
Persistence is key. All good here. Thanks!

PR #1476, opened 2024-03-30
feat: Add pause command

LucaSeri:
### Prerequisite checklist

### Description of changes

The pause command is similar to ``stop``, except on running ``start`` the unikernel will not be restarted, instead it will just resume from the instruction when it was paused. This commit just adapts ``stop`` to instead use the already existing ``Pause`` functionality inside the machine controller.


nderjung (approved):
Thanks!

PR #1487, opened 2024-04-01
fix(action): Elevate group permissions if unknown UID:GID are used

craciunoiuc:
### Prerequisite checklist

### Description of changes

This moves the check outside the container and makes sure that files are always accesible. To do this, group permissions are elevated in the working directories when unknown UID:GID are used. This works because groups are hardcoded in the container for the 1000, 1001, 1002 GIDs.

To disable this if permissions are to be left untouched set 'grouprwx' to 'false'.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -32,6 +32,10 @@ inputs:
     description: Path to store runtime artifacts
     required: false
     default: /github/workspace/.kraftkit/runtime
+  grouprwx:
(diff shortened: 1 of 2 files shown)


craciunoiuc:
rebased to latest staging will cancel runs

all passed before

nderjung (approved):
Thanks!

PR #1494, opened 2024-04-03
refactor(compose): Wrap with processtree

LucaSeri:
### Prerequisite checklist

### Description of changes

Improve the CLI by wrapping the underlying kraft calls with processtree.

The solution is a bit hacky, since embedding processtrees within one another does not work if they are in ``--log-type fancy``. Instead, only use ``--log-type`` for the top level processtree and artificially set the logging type to ``basic`` for the inner ones.

### DIFF
--- a/iostreams/iostreams.go
+++ b/iostreams/iostreams.go
@@ -188,7 +188,7 @@ func (s *IOStreams) IsStdoutTTY() bool {
 	}
 
 	// support KRAFTKIT_FORCE_TTY
-	if s.term.IsTerminalOutput() {
+	if s.term != nil && s.term.IsTerminalOutput() {
 		return true
 	}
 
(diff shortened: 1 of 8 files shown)


iostreams/iostreams.go
```diff
@@ -188,7 +188,7 @@ func (s *IOStreams) IsStdoutTTY() bool {
 	}
 
 	// support KRAFTKIT_FORCE_TTY
-	if s.term.IsTerminalOutput() {
+	if s.term != nil && s.term.IsTerminalOutput() {
```
LucaSeri, line 191: I didn't dig into why this happens when nesting ``processtrees``, but maybe this is not the way to handle this seeing as this check is not performed anywhere else.

I'm not sure whether this is fine or it means somewhere else needs fixing.

LucaSeri:
Closing this for the moment as we decided this way of wrapping is a pretty ugly hack.

PR #1499, opened 2024-04-03
fix(oci): Clean partial files on error

### LINKED ISSUE #1479: Unexpected EOF
### Describe the bug

When using kraftkit I sometimes get the " unpacking the image: writing file: unexpected EOF" error.

### Steps to reproduce

To achieve this error I used the "kraft run" command in the "catalog/examples/http-python3.12-flask3.0" directory and let it begin pulling the image and interrupting it (CTRL + C) after letting it run for about half the time it takes to fully pull the image. This almost always left me with the above mentioned error, and upon running "kraft run" again each time it would almost immediately give this error again. Deleting the .tar archive from the ./local/share/kraftkit/runtime/oci/digest/sha256 fixed this bug, allowing me to run "kraft run" without issue again. Both the cancellation and deletion can be repeated to get this error again / make it go away.

I think this happens because either the incomplete tar file is not checked (or an attempt to read it is made), but I am not sure of this.
Most times after getting this bug and making it disappear I get the "received event error: context canceled, E  event: context canceled" errors, but am unsure if it is related.

I am using Native Ubuntu 22.04

### Expected behavior

Not getting said error.

_No response_

_No response_

### Relevant log output

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

This does not fully depollute the local cache, but it's a start

GitHub-Fixes: #1479

### DIFF
--- a/oci/handler/directory.go
+++ b/oci/handler/directory.go
@@ -402,6 +402,12 @@ func (handle *DirectoryHandler) PullDigest(ctx context.Context, mediaType, fullr
 			defer manifestWriter.Close()
 
 			if _, err = manifestWriter.Write(manifestRaw); err != nil {
+				if err2 := manifestWriter.Close(); err2 != nil {
+					return fmt.Errorf("%w: could not close manifest writer: %w", err, err2)
+				}
+				if err2 := os.RemoveAll(manifestPath); err2 != nil {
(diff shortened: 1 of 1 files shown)


nderjung (approved):
Thanks!

PR #1506, opened 2024-04-04
fix(internal): Limit interface name size to os limit

### LINKED ISSUE #1456: ``kraft net create`` fails if network name is too long
### Describe the bug

When trying to create a network, there is a name length limit that is generated by the ``kraftkit`` convention of naming interfaces. On linux, for instance, this limit is 15 characters.

A possible solution would be to have some internal mapping from the network name to some random interface name that adheres to the limit.

### Steps to reproduce

Run:
```bash
kraft net create a-somewhat-longer-name

E  bridge creation for a-somewhat-longer-name failed: numerical result out of range
```

### Expected behavior

For the network to be created

_No response_

_No response_

### Relevant log output

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

GitHub-Fixes: #1456

### DIFF
--- a/machine/network/bridge/v1alpha1.go
+++ b/machine/network/bridge/v1alpha1.go
@@ -23,6 +23,13 @@ import (
 	"kraftkit.sh/machine/network/macaddr"
 )
 
+const (
+	// defaultNetworkMaxSize is the maximum size of a network name on Linux.
+	// We subtract 5 to account for the network interface suffix '@if%d%d%d'.
+	// The value 16 is taken from the glibc headers.
+	defaultNetworkNameMaxSize = 15 - 5
+)
+
 type v1alpha1Network struct{}
 
 func NewNetworkServiceV1alpha1(ctx context.Context, opts ...any) (networkv1alpha1.NetworkService, error) {
@@ -33,6 +40,10 @@ func NewNetworkServiceV1alpha1(ctx context.Context, opts ...any) (networkv1alpha
 func (service *v1alpha1Network) Create(ctx context.Context, network *networkv1alpha1.Network) (*networkv1alpha1.Network, error) {
 	if network.Name == "" {
 		return nil, fmt.Errorf("cannot create network without name")
+	}
+
+	if len(network.Name) > defaultNetworkNameMaxSize {
+		return nil, fmt.Errorf("network name is too long, must be less than %d characters", defaultNetworkNameMaxSize)
 	}
 
 	if network.ObjectMeta.UID == "" {
--- a/test/e2e/cli/net_create_test.go
+++ b/test/e2e/cli/net_create_test.go
@@ -83,7 +83,7 @@ var _ = Describe("kraft net create", func() {
 	When("invoked with one positional argument and a network without a subnet", func() {
 		BeforeEach(func() {
 			cmd.Args = append(cmd.Args, "--network", "172.45.1.1")
-			cmd.Args = append(cmd.Args, "test-create-0")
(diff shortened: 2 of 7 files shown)


internal/cli/kraft/net/create/create.go
```diff
@@ -20,6 +20,13 @@ import (
 	"kraftkit.sh/machine/network"
 )
 
+const (
+	// defaultNetworkMaxSize is the maximum size of a network name on Linux.
+	// We subtract 2 to account for the network interface suffix.
+	// The value 16 is taken from the glibc headers.
+	defaultNetworkNameMaxSize = 15 - 2
```
LucaSeri: This is still a somewhat tight limit. We append at least 4 characters, possibly going into double digits or maybe more (`@if%d`). A safe limit could be obtained by looking at the mask size of the network and seeing the number of interfaces that we could have on that network, but this limit might be too loose for the average use case and prevent users from naming their network properly.

Maybe we could have this limit (maybe rather 15 - 4) and warn the users about not being able to fully use the address space, possibly erroring in a better way if we run out of names for interfaces?

craciunoiuc: Ah sorry I thought it was just a number, then I'll decrease to accomodate for `@if%d%d`

craciunoiuc: As for erroring in the code when out of bounds makes sense, though I can't fully point the place where this should happen. Maybe you can do this in a separate PR? 100 interfaces should be enough for now :D

internal/cli/kraft/net/create/create.go
```diff
@@ -56,9 +63,13 @@ func NewCmd() *cobra.Command {
 	return cmd
 }
 
-func (opts *CreateOptions) Pre(cmd *cobra.Command, _ []string) error {
+func (opts *CreateOptions) Pre(cmd *cobra.Command, args []string) error {
 	opts.Driver = cmd.Flag("driver").Value.String()
 
+	if len(args[0]) > defaultNetworkNameMaxSize {
```
LucaSeri: Thanks @craciunoiuc!
One last thing, can you please move this to ``Run`` and also add a check ``len(args) > 0``? Some places including compose use this as an API directly :pray:

LucaSeri (changes requested):

internal/cli/kraft/net/create/create.go
craciunoiuc: sure thingy

LucaSeri (approved):
Thanks! Some tests are failing but logic looks good.

craciunoiuc:
It seems we were not respecting that limit ourselves in the tests, will change now

craciunoiuc:
@LucaSeri limited test if names and works now

internal/cli/kraft/net/create/create.go
```diff
@@ -20,6 +20,13 @@ import (
 	"kraftkit.sh/machine/network"
 )
 
+const (
+	// defaultNetworkMaxSize is the maximum size of a network name on Linux.
+	// We subtract 5 to account for the network interface suffix '@if%d%d%d'.
+	// The value 16 is taken from the glibc headers.
+	defaultNetworkNameMaxSize = 15 - 5
```
nderjung: This should be internalised to the bridge network implementation, since it is specific to Linux.

craciunoiuc: done

nderjung: Perfect, thanks!

nderjung (approved):
Thanks!

PR #1507, opened 2024-04-04
fix(cloud): handle volume import cancellation

antoineco:
Until now, context cancellations (e.g. through `Ctrl-C`) were not propagated to `copyCPIO`. As a result, imports would continue until completion instead of being interrupted, potentially for a very long time if the tranfered data is large.

This PR ensures that context cancellations are propagated to the `tls.Conn` by setting a write deadline accordingly.

--- 

⚠️ Due to a bug in KraftKit, `paraprogress.Start()` returns `nil` upon context cancellation instead of the error returned by the inner function.

For example, when the context is cancelled after introducing the feature from this PR, `io.Copy` returns:
```
write tcp4 172.24.172.99:51516->136.144.49.17:42069: i/o timeout
```
And `copyCPIO` returns:
```
incomplete write (42/100)
```

Yet, `paraprogress.Start()` returns `nil` so `fancymap.PrintFancyMap` is called regardless.

Illustration, with a few `Println()` added to show errors:
```
[\] Importing data (19 MiB) ••••••••••••••••••••••••••••••••••••••   23% [1.0s]
^C
write tcp4 172.24.172.99:51516->136.144.49.17:42069: i/o timeout
incomplete write (42/100)

[●] Import complete
 │
 ├─── volume: f127ea39-0e42-4c1d-b416-805e22872408
 ├─ imported: 19 MiB
 └─ capacity: 25 MiB

 E  could not import volume data
```

### DIFF
--- a/internal/cli/kraft/cloud/volume/import/cpio.go
+++ b/internal/cli/kraft/cloud/volume/import/cpio.go
@@ -14,6 +14,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"time"
 
 	"kraftkit.sh/initrd"
 )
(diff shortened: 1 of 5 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #1510, opened 2024-04-04
feat(internal)!: Switch to multiple subdomains/domains

### LINKED ISSUE #1474: Allow specifying multiple domains and subdomains in kraft cloud
### Feature request summary

Since the API change on Kraft Cloud you are able to specify a mixture of multiple domains and subdomains (see [here](https://docs.kraft.cloud/api/v1/services/#creating-a-new-service-group) and [here](https://docs.kraft.cloud/api/v1/instances/#creating-a-new-instance)). To support this `kraft` should be extended to accept multiple `--domain` and `--subdomain` (and a mixture of both) parameters.

Just like in the previous API with `dns_name`, a trailing dot (`.`) for the `domains[].name` field indicates an FQDN. So to have `www.example.com`, you set `domains[].name: www.example.com.` (trailing dot), which should be the result of `--domain www.example.com`. A missing dot indicates a subdomain. So to have `my-project.<metro>.kraft.host`, you set `domains[].name: my-project`, which should be the result of `--subdomain my-project`.

The certificate does need to be set at this point. It is automatically picked or generated.

### Describe alternatives

_No response_

### Related architectures

None

### Related platforms

None

### Additional context

_No response_

craciunoiuc:
### Prerequisite checklist

### Description of changes

GitHub-Fixes: #1474

Not sure if certificates are also needed, what do you think @antoineco?

### DIFF
--- a/internal/cli/kraft/cloud/deploy/deploy.go
+++ b/internal/cli/kraft/cloud/deploy/deploy.go
@@ -35,11 +35,11 @@ type DeployOptions struct {
 	Client                 kraftcloud.KraftCloud     `noattribute:"true"`
 	Compress               bool                      `local:"true" long:"compress" short:"z" usage:"Compress the initrd package (experimental)"`
 	DeployAs               string                    `local:"true" long:"as" short:"D" usage:"Set the deployment type"`
+	Domain                 []string                  `local:"true" long:"domain" short:"d" usage:"Set the domain names for the service"`
 	DotConfig              string                    `long:"config" short:"c" usage:"Override the path to the KConfig .config file"`
 	Env                    []string                  `local:"true" long:"env" short:"e" usage:"Environmental variables"`
 	Features               []string                  `local:"true" long:"feature" short:"f" usage:"Specify the special features to enable"`
 	ForcePull              bool                      `long:"force-pull" usage:"Force pulling packages before building"`
-	FQDN                   string                    `local:"true" long:"fqdn" short:"d" usage:"Set the fully qualified domain name for the service"`
 	Jobs                   int                       `long:"jobs" short:"j" usage:"Allow N jobs at once"`
 	KernelDbg              bool                      `long:"dbg" usage:"Build the debuggable (symbolic) kernel image instead of the stripped image"`
 	Kraftfile              string                    `local:"true" long:"kraftfile" short:"K" usage:"Set the Kraftfile to use"`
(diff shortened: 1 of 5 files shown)


internal/cli/kraft/cloud/deploy/deploy.go
```diff
@@ -63,7 +63,7 @@ type DeployOptions struct {
 	ScaleToZero            bool                      `local:"true" long:"scale-to-zero" short:"0" usage:"Scale the instance to zero after deployment"`
 	ServiceGroupNameOrUUID string                    `long:"service-group" short:"g" usage:"Attach the new deployment to an existing service group"`
 	Strategy               packmanager.MergeStrategy `noattribute:"true"`
-	SubDomain              string                    `local:"true" long:"subdomain" short:"s" usage:"Set the name to use when provisioning a subdomain"`
+	SubDomain              []string                  `local:"true" long:"subdomain" short:"s" usage:"Set the names to use when provisioning subdomains"`
```
antoineco, line 66:
```suggestion
	SubDomains             []string                  `local:"true" long:"subdomain" short:"s" usage:"Set the names to use when provisioning subdomains"`
```

We should also probably indicate "repeat for multiple values" (or is it comma-separated?) in the usage string.

internal/cli/kraft/cloud/deploy/deploy.go
```diff
@@ -119,10 +119,8 @@ func (opts *DeployOptions) Pre(cmd *cobra.Command, _ []string) error {
 	opts.Strategy = packmanager.MergeStrategy(cmd.Flag("strategy").Value.String())
 
 	domain := cmd.Flag("domain").Value.String()
-	if len(domain) > 0 && len(opts.FQDN) > 0 {
-		return fmt.Errorf("cannot use --domain and --fqdn together")
-	} else if len(domain) > 0 && len(opts.FQDN) == 0 {
```
antoineco, line 124: Does this mean we _can_ use both together now?

internal/cli/kraft/cloud/deploy/deploy.go
craciunoiuc: you repeat the flag for multiple values afaik `-s ... -s ...`

internal/cli/kraft/cloud/deploy/deploy.go
craciunoiuc: That's what I could understand from <https://github.com/unikraft/kraftkit/issues/1474> and the API, not sure if correct.

craciunoiuc:
checks failed because of GH fail, will rerun now

internal/cli/kraft/cloud/deploy/deploy.go
antoineco: Then we should have _some way_ to communicate this to the user via `--help`, because the flag's usage alone doesn't indicate how to pass multiple values, and each CLI tool has their own way of doing that.

antoineco: I see now that KraftKit indicates the type `strings` for such flag:
```
-e, --env strings            Environmental variables
```

We cannot possibly write "repeat for multiple values" in every []string flag, so maybe an example that uses multiple domain values is enough. In this case, feel free to ignore my comment.

antoineco (approved):

PR #1537, opened 2024-04-10
feat: Introduce systemd (WIP)

### LINKED ISSUE #1466: Implement systemd unikernel manager
### Feature request summary

This issue tracks the ability to set and manage Unikraft unikernels as individual systemd entries.  Ultimately this would give users the ability to use unikernels as secure, persistent system services.

At a glance, creating and managing unikernels as systemd processes should be designed to be simple, making it an effective mechanism and alternative approach for system services.  As such, a new flag `--systemd` should be provided to the following subcommands to enable managing and running via this interface:

- `kraft run --systemd [...]`
- `kraft start --systemd [...]`
- `kraft stop --systemd [...]`
- `kraft rm --systemd [...]`
- `kraft ps --systemd [...]`

In order to prevent large disruption to the individual subcommand codebases, (i.e. preventing having a large `if opts.Systemd` flag with a different line-of-execution), a simple wrapper around the machine API can be used to proxy all interface methods into relevant systemd service management actions.

In order to propagate such functionality, a new machine service implementation is required in `kraftkit.sh/machine/platform/systemd` which implements the `MachineService interface`:

https://github.com/unikraft/kraftkit/blob/71509c7b8d211d50c863fb1d34696c9be654b5e3/api/machine/v1alpha1/machine.zip.go#L149-L162

To utilize the machine API via a controller, it should be a simple case of wrapping an existing controller, e.g.:

```go
import (
  machineapi "kraftkit.sh/api/machine/v1alpha1"
  "kraftkit.sh/machine/platform/systemd"
)

var controller machineapi.MachineService

if opts.Systemd {
  controller, err = systemd.NewMachineV1alpha1ServiceSystemdWrapper(ctx, controller)
  if err != nil {
    // [...]
  }
}
```

The new method `NewMachineV1alpha1ServiceSystemdWrapper` will consume all interface methods and directly interface with systemd which can be accomplished with [this existing Go module](https://github.com/kardianos/service).

> [!WARNING]
> Installing systemd process should ultimately create an entry which invokes `kraft` itself and that not all interface methods are a simple 1-to-1 call of one another.  This is because wrapper will pass through the machine API store twice and some verbs do not map directly to the machine API.  Careful consideration should be practice when proxying commands.  For example, the `Create` command should both create the systemd service entry and also initialize the unikernel.  The `Start` method, however, could, only start the systemd service as this will invoke `kraft` which will actually start the unikernel.

### Additional context

@MdSahil-oss wants to work on this.

mdsahiloss:
### Prerequisite checklist

### Description of changes

GitHub-Closes: #1466

### DIFF
--- a/internal/cli/kraft/run/utils.go
+++ b/internal/cli/kraft/run/utils.go
@@ -459,3 +459,66 @@ func (opts *RunOptions) parseEnvs(_ context.Context, machine *machineapi.Machine
 
 	return nil
 }
+
+// GetArgs returns all the arguments of run command as an array of strings.
+func (opts *RunOptions) GetArgs() []string {
+	args := []string{}
+	if opts.Detach {
+		args = append(args, "--detach")
+	}
+	if opts.DisableAccel {
+		args = append(args, "--disable-acceleration")
+	}
+	if opts.PrefixName {
+		args = append(args, "--prefix-name")
+	}
+	if opts.WithKernelDbg {
+		args = append(args, "--symbolic")
+	}
+	if opts.Remove {
+		args = append(args, "--rm")
+	}
+	if len(opts.Name) > 0 {
+		args = append(args, "--name", opts.Name)
+	}
+	if len(opts.Architecture) > 0 {
+		args = append(args, "--arch", opts.Architecture)
+	}
+	if len(opts.RunAs) > 0 {
+		args = append(args, "--as", opts.RunAs)
+	}
+	if len(opts.KernelArgs) > 0 {
+		args = append(args, "--kernel-arg", strings.Join(opts.KernelArgs, " "))
+	}
+	if len(opts.Kraftfile) > 0 {
+		args = append(args, "--kraftfile", opts.Kraftfile)
+	}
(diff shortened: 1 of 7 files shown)


internal/cli/kraft/run/utils.go
```diff
@@ -459,3 +459,66 @@ func (opts *RunOptions) parseEnvs(_ context.Context, machine *machineapi.Machine
...
+
+// GetArgs returns all the arguments of run command as an array of strings.
+func (opts *RunOptions) GetArgs() []string {
+	args := []string{}
+	if opts.Detach {
+		args = append(args, "--detach")
+	}
+	if opts.DisableAccel {
+		args = append(args, "--disable-acceleration")
+	}
```
nderjung, line 471: Check out: https://pkg.go.dev/reflect#StructTag.Get

This will simplify getting the flag name.

mdsahiloss: Got it 👍🏽

craciunoiuc:
Hey @MdSahil-oss 

Do you have an estimate for when you'll restart work on this? :pray:

PR #1556, opened 2024-04-17
feat(bridge): Allow arbitrary network length

LucaSeri:
### Prerequisite checklist

### Description of changes

Linux has a limitation on the length of interface names. Hence, this commit attempts to maintain the old naming scheme on a best effort basis, but in case that is not possible, it instead uses a hash to name the bridge, keeping the mapping internally.

Previously, the ``network.name`` used to be the same as the interface name. This was convenient because it allowed easy checks on whether a network already exists. 

The approach I used here is the following: 
* During creation, if the object from the store already has a ``UUID``, then it means the network with that name was already created (regardless of the actual interface name)
* During creation, if we error out, return ``nil`` instead of ``network`` such that nothing gets saved and we dont pollute the store with networks that do not actually have an interface.

### DIFF
--- a/machine/network/bridge/v1alpha1.go
+++ b/machine/network/bridge/v1alpha1.go
@@ -23,13 +23,6 @@ import (
 	"kraftkit.sh/machine/network/macaddr"
 )
 
-const (
-	// defaultNetworkMaxSize is the maximum size of a network name on Linux.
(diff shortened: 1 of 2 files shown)


nderjung (approved):
Thanks!

PR #1564, opened 2024-04-18
fix(qemu): Show first line after receiving "Booting from ROM"

nderjung:
### Prerequisite checklist

### Description of changes

Return the channel if the phrase from QEMU has been received. Since the line has now been consumed, the next won't be if we simply `continue`, because this will be consumed to be checked.  Without this, the first line from the QEMU log will be lost.  For applications which output a single line, this would otherwise be lost.


craciunoiuc (approved):
All good here. Thanks!

PR #1585, opened 2024-04-21
refactor(webinstall): Reprompt after invalid input instead of exiting

### LINKED ISSUE #1561: Webinstall script: Prompt user to try again on invalid input
### Feature request summary

When installing kraftkit using the Webscript, if the user enters an invalid input the script throws an exception and exits. I think it would be nicer to just prompt the user to try again. For example, when I ran the installation script I made a typo and entered "zhs" instead of "zsh", which caused the script to exit immediately. If nobody has an objection to this I would like to contribute this change. 

### Describe alternatives

_No response_

### Related architectures

None

### Related platforms

None

### Additional context

_No response_

HendrikHuebner:
### Prerequisite checklist

### Description of changes
Closes  #1561

When a user enters something invalid in response to a prompt by the webinstall script, the script now displays a message such as: 

`Invalid input. Please try again or exit with Ctrl+C.`

Additionally, for yes/no questions I have created the prompt_user_yes_no function, which simplifies the script a little bit since it gets rid of some duplicated code.

### DIFF
--- a/tools/webinstall/install.sh
+++ b/tools/webinstall/install.sh
@@ -199,6 +199,27 @@ get_user_response() {
     _RETVAL="$_gur_read_answer"
 }
 
+# prompt_user_yes_no asks the user a question which must be answered with yes or no. Returns a boolean value.
+# $1: question
+# $2: default answer
+# Returns:
+# _RETVAL: true for "yes", false for "no"
+prompt_user_yes_no() {
+  while true;
+  do
+      get_user_response "$1" "$2"
+      _answer="$_RETVAL"
+
+      if printf "%s" "$_answer" | "$GREP" -q -E "$_NO_ANS"; then
+          return 1
+      elif printf "%s" "$_answer" | "$GREP" -q -E "$_YES_ANS"; then
+          return 0
+      else
+          printf "Invalid input. Please try again or exit with Ctrl+C\n"
+      fi
+    done
+}
+
 # do_cmd runs a command and asks the user if they want to retry with root
 # if it fails
 # $@: command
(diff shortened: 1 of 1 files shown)


tools/webinstall/install.sh
```diff
@@ -199,6 +199,27 @@ get_user_response() {
...
+prompt_user_yes_no() {
+  while true;
+  do
+      get_user_response "$1" "$2"
+      _answer="$_RETVAL"
+
+      if printf "%s" "$_answer" | "$GREP" -q -E "$_NO_ANS"; then
+          return 1
+      elif printf "%s" "$_answer" | "$GREP" -q -E "$_YES_ANS"; then
+          return 0
```
craciunoiuc, line 216: I think this does not catch default cases?

see `_NO_ANS_DEFAULT`/`_YES_ANS_DEFAULT`

craciunoiuc (changes requested):
Hey @HendrikHuebner 

Looks really good, left you a comment. I still need to try it out.

Can you also reword your commit to start with `refactor(webinstall):` :pray: 

`refactor(webinstall): Prompt user again on invalid input instead of exiting`

Make sure you leave a single commit afterwards

tools/webinstall/install.sh
HendrikHuebner: I don't think those are actually needed. The default case (an empty string) is already handled with the second argument of get_user_response. The only thing that makes the `_NO_ANS_DEFAULT/_YES_ANS_DEFAULT `regexes different is that they match an empty string as well, so if I'm not mistaken that check was duplicated in the original script as well.

craciunoiuc: They match also empty strings and only newlines

I'll try it like this now and see if it works

craciunoiuc: Tried it out now

Seems to work, nice

craciunoiuc (changes requested):
Seems that the commit is still to long 😅 

Try renaming it to, this one, it's 70 chars:

`refactor(webinstall): Reprompt after invalid input instead of exiting`

We can merge after

Previous one was `76` > `74` chars whoops

craciunoiuc (approved):
All good here. Thanks!

PR #1597, opened 2024-04-24
feat(cloud): Add cloud purge subcommand

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/internal/cli/kraft/cloud/certificate/remove/remove.go
+++ b/internal/cli/kraft/cloud/certificate/remove/remove.go
@@ -21,12 +21,13 @@ import (
 )
 
 type RemoveOptions struct {
-	Output string `long:"output" short:"o" usage:"Set output format. Options: table,yaml,json,list,raw" default:"table"`
-	All    bool   `long:"all" usage:"Remove all certificates"`
-
-	metro         string
-	token         string
-	allowInsecure bool
+	Output        string                `long:"output" short:"o" usage:"Set output format. Options: table,yaml,json,list,raw" default:"table"`
+	All           bool                  `long:"all" usage:"Remove all certificates"`
+	AllowInsecure bool                  `noattribute:"true"`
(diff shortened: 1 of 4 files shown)


internal/cli/kraft/cloud/certificate/remove/remove.go
```diff
@@ -88,7 +87,7 @@ func (opts *RemoveOptions) Pre(cmd *cobra.Command, args []string) error {
 }
 
 func (opts *RemoveOptions) Run(ctx context.Context, args []string) error {
-	auth, err := config.GetKraftCloudAuthConfig(ctx, opts.token)
+	auth, err := config.GetKraftCloudAuthConfig(ctx, opts.Token)
 	if err != nil {
 		return fmt.Errorf("could not retrieve credentials: %w", err)
 	}
```
nderjung: Can you update this code to use this convention?

https://github.com/unikraft/kraftkit/blob/b495be88eb42bc7ffe28f2a33f778dca719127ff/internal/cli/kraft/cloud/instance/create/create.go#L68-L78

I.e. expose `Auth` and `Client` (the general-purpose KC client) as non-attributes? This way we can just pass in existing values from external calls.

(please also apply to img/remove.go)

craciunoiuc:
Needs to be updated to account for new commands

craciunoiuc:
Will leave in a draft state as it is currently broken due to not being able to remove instances/services with scaling attached and no way to remove scaling everywhere.

Thus this needs initial manual intervention which defeats the purpose of the command.

craciunoiuc:
Close for now as this will probably make its way into the new cli

jedevc:
What should this command do? If you want it in the new cli, probably worth opening a linear ticket!

craciunoiuc:
PURGE

In the new cli it should remove all resources from all registered metros

PR #1638, opened 2024-05-09
fix(oci): Don't delete dependencies of index

LucaSeri:
When we package an application it can be the case that we have a dependency on another package (e.g. a kernel). In these cases, the dependency is not actually packed together with our application, instead we just keep a reference to it. The problem arises when we attempt to overwrite such a package. Because all dependencies were previously deleted, this meant that both of the underlying packages would be removed, but only the application would get repackaged, resulting in an app that cannot be run due to missing layers.

Setting this flag to false and not removing dependencies solves this issue, but causes the cache to grow indefinitely, which is less than ideal.

### Prerequisite checklist

### DIFF
--- a/oci/pack.go
+++ b/oci/pack.go
@@ -291,7 +291,7 @@ func NewPackageFromTarget(ctx context.Context, targ target.Target, opts ...packm
 		}
 
 	case packmanager.StrategyOverwrite:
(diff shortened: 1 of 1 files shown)


LucaSeri:
I opened https://github.com/unikraft/kraftkit/issues/1639 to track the implementation of a garbage collection mechanism that will fix the indefinite growth of the cache.

nderjung (approved):
Thanks!

PR #1678, opened 2024-05-26
fix!: Major improvements to `kraft cloud compose`

nderjung:
### Prerequisite checklist

### Description of changes

This PR includes major fixes and improvements which increase the compatibility of using `kraft cloud compose` against existing compose projects.

This series of commits largely include numerous fixes to the creation (building) of images as well as some partial refactors to the `kraft cloud compose up` subcommand for better compatibility.  Additionally, the following new features are available:

- Dynamically building working directories and projects without the need for a `Kraftfile` and only the presence of a `Dockerfile`;
- Introduce a `--runtime` flag allowing users to specify an alternative pre-built Unikraft unikernel runtime to be used when images are dynamically built from a `Dockerfile`;
- Introduce a `--no-build` flag to `kraft cloud compose up` to prevent the build step from occuring before instance instantiation;
- Preliminary support for an `--entrypoint` flag which is used to signal when to use arguments supplied in the existing OCI image and when to overwrite them.


craciunoiuc:
What would be the alternative to this then? Using the internal IP everywhere?

nderjung:
All instances in your account can communicate via internal networking.

initrd/ociimage.go
```diff
@@ -213,7 +213,7 @@ func (initrd *ociimage) Build(ctx context.Context) (string, error) {
 		// No other file types are part of the archive.  As a result we only check
 		// whether the path is the same as the reference path which indicates
 		// whether the entry is a symbolic link.
-		if f.FileType == scfile.TypeSymLink {
+		if f.FileType == scfile.TypeSymLink || f.FileType == scfile.TypeHardLink {
 			header.Mode |= cpio.TypeSymlink
```
craciunoiuc: Are you 100% sure this is correct? You are marking the HardLink as a SymLink

nderjung: Good point; i will double check.

internal/cli/kraft/cloud/instance/create/create.go
```diff
@@ -191,6 +192,13 @@ func Create(ctx context.Context, opts *CreateOptions, args ...string) (*kcclient
 	}
 	if opts.Memory > 0 {
 		req.MemoryMB = ptr(int(opts.Memory))
+	} else {
+		// Set the default memory to the size of the image if not specified plus an
+		// arbitrary 10% buffer.  Only set the value if it is greater than 128 MiB.
+		mem := int(math.Round(float64(image.SizeInBytes / 1024 / 921)))
```
craciunoiuc: This looks wrong, why are you dividing KiB by `921` and not `1024` to get MiB

I Get the 10% buffer, but you can just multiply by `1.1` after

(Go anyway transforms the divisions you a single op for efficiency) (I think)

craciunoiuc: Moreover, remember we are still using bbuddy, which will waste 1/2 of the memory it allocates, so 10% is not enough for most cases

internal/cli/kraft/cloud/compose/up/up.go
```diff
@@ -173,6 +177,45 @@ func Up(ctx context.Context, opts *UpOptions, args ...string) error {
...
+
+			userPkgName = fmt.Sprintf(
+				"%s/%s",
+				userName,
+				strings.ReplaceAll(service.Image, "_", "-"),
+			)
+			officialPkgName = strings.ReplaceAll(service.Image, "_", "-")
+		} else {
+			userPkgName = fmt.Sprintf(
+				"%s/%s-%s:latest",
```
craciunoiuc, line 202: This is not prefixed with `index.unikraft.io`?

internal/cli/kraft/cloud/instance/create/create.go
nderjung: Will double check which allocator we're using. Funny though, this was enough for my testing to work.

I can change it to `(n/1024/1024)*1.1` to make it more easily readable.

nderjung: But what i can do is just increase the memory size to, instead, to the nearest 2^N.

internal/cli/kraft/cloud/compose/up/up.go
nderjung: The API does not like this. Additionally, `index.unikraft.io` (+`/official`) is stripped in the next `create.Create` method anyway.

craciunoiuc: okay

internal/cli/kraft/cloud/instance/create/create.go
craciunoiuc: Actually, just leave the 10% buffer, we can just pray to Simon to get the other allocator in

initrd/ociimage.go
nderjung: This should hopefully now be resolved through the latest commit.

craciunoiuc (approved):
How bad can it go?

PR #1683, opened 2024-05-30
build: Bump all non-dependabot dependencies

craciunoiuc:
### Prerequisite checklist

### Description of changes

This PR is required as `macos-11` and `ubuntu-20.04` will soon be deprecated.

Moreover, as we pin the exact debian version we must make sure we update at least once an year to keep apt lists up to date.

While at it, do patch bumps for qemu and golang.

Qemu 9 is already out but does not offer significant changes for us so we can skip that one for now and wait for 10.

Note: Ubuntu 24.04 is currently in 'beta' (for actions) but is set to release soon.


nderjung (approved):
We are so fresh. Thanks!

PR #1698, opened 2024-06-03
fix(run): Pass workdir when preparing rootfs

LucaSeri:
### Prerequisite checklist

  - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
  - [x] Tested your changes against relevant architectures and platforms;
  - [x] Ran `make fmt` on your commit series before opening this PR;
  - [x] Updated relevant documentation.

### Description of changes

Follow-up for https://github.com/unikraft/kraftkit/pull/1696. Passing the workdir in the ``prepareRootfs`` step was previously missed, causing runs outside the current directory to fail.

PR #1696, opened 2024-06-02
fix(initrd): Correctly propagate workdir to dockerfile

LucaSeri:
<!--

Thank you for opening a new PR to the Unikraft Open Source Project!  We welcome
new changes, features, fixes, and more!  Please fill in this form to indicate
the status of your PR.  Please ensure you have read the contribution guidelines
before opening a new PR as this will cover the PR process:

  https://unikraft.org/docs/contributing/

  Kraftkit follows the same guidelines as the Unikraft Open Source Project.

-->

### Prerequisite checklist

<!--
Please mark items appropriately:
-->

  - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
  - [x] Tested your changes against relevant architectures and platforms;
  - [x] Ran `make fmt` on your commit series before opening this PR;
  - [x] Updated relevant documentation.

### Description of changes
Commit 5cb57adbb0cf81fbc59ce69761c31fbe186030d5 reworked the way the workdir is passed to different initrd implementation. This ended up
in having a duplicate workdir entry in the dockerfile initrd type. Changes take into account that with the new implementation, the path of the dockerfile is relative to the workdir.

This change ultimately solves an issue related to kraftkit not being able to find the dockerfile when building outside the cwd:

```
$ kraft build ../scss/redis/
<!> building rootfs                                                                                                                                                                                                                                               x86_64 [2.8s]
 i  creating ephemeral buildkit container
 i  #1 [internal] load build definition from Dockerfile
 i  #1 transferring dockerfile: 2B done
 i  #1 DONE 0.0s
 E  could not wait for err group: could not solve: failed to solve: failed to read dockerfile: open Dockerfile: no such file or directory
```

<!--
Please provide a detailed description of the changes made in this new PR.
-->

PR #1707, opened 2024-06-07
fix(initrd): Reverse order of detection

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/initrd/detect.go
+++ b/initrd/detect.go
@@ -12,13 +12,13 @@ import (
 // New attempts to return the builder for a supplied path which
 // will allow the provided ...
 func New(ctx context.Context, path string, opts ...InitrdOption) (Initrd, error) {
-	if builder, err := NewFromOCIImage(ctx, path, opts...); err == nil {
-		return builder, nil
-	} else if builder, err := NewFromDockerfile(ctx, path, opts...); err == nil {
+	if builder, err := NewFromFile(ctx, path, opts...); err == nil {
 		return builder, nil
 	} else if builder, err := NewFromDirectory(ctx, path, opts...); err == nil {
 		return builder, nil
-	} else if builder, err := NewFromFile(ctx, path, opts...); err == nil {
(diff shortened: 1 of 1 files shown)


initrd/dockerfile.go
```diff
@@ -126,6 +126,10 @@ func NewFromDockerfile(ctx context.Context, path string, opts ...InitrdOption) (
 		return nil, fmt.Errorf("file is not a Dockerfile")
 	}
 
+	if _, err := os.Stat(path); err != nil {
```
LucaSeri: This will actually break building outside the ``cwd``.
E.g.: ``kraft build ./example``

This raises an issue about what "path" means in the context of this initrd implementation. The root problem is actually how the ramfs is constructed [here](https://github.com/unikraft/kraftkit/blob/staging/internal/cli/kraft/run/utils.go#L367). Path is not relative to the cwd, but instead it is relative to the working directory, which is only seeded after the opts are being applied.

A fix would be to move the check after the opts and concatenate the path to the workdir, or maybe the whole thing should be redesigned a bit such that path is relative to the cwd?

LucaSeri (changes requested):
Hi @craciunoiuc, thanks for this!

Please see my comment, which is related to the fixes in https://github.com/unikraft/kraftkit/pull/1696 and https://github.com/unikraft/kraftkit/pull/1698.

initrd/dockerfile.go
craciunoiuc: Hmm, no idea, in the error that triggered me to fix this, the path had the given workdir also

Also, this check is already present in the file/directory anyway so i thought to match

Still the second commit should go in as is because the order is currently not correct

nderjung: Remove the stat and keep the re-order; update commit message about re-ordering.

LucaSeri (approved):
All good, thanks!

Reviewed-by: Luca Seritan luca.seritan@gmail.com

nderjung (approved):
Thanks!

PR #1772, opened 2024-07-12
fix(cloud): Several fixes to the cloud commands

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/internal/cli/kraft/cloud/instance/create/create.go
+++ b/internal/cli/kraft/cloud/instance/create/create.go
@@ -52,17 +52,17 @@ type CreateOptions struct {
 	Replicas            uint                          `local:"true" long:"replicas" short:"R" usage:"Number of replicas of the instance" default:"0"`
 	Rollout             RolloutStrategy               `noattribute:"true"`
 	RolloutQualifier    RolloutQualifier              `noattribute:"true"`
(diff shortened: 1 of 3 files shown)


internal/cli/kraft/cloud/instance/create/create.go
```diff
@@ -743,6 +743,10 @@ func (opts *CreateOptions) Pre(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("could not populate metro and token: %w", err)
 	}
 
+	if opts.ScaleToZeroCooldown < time.Millisecond && opts.ScaleToZeroCooldown != 0 {
+		return fmt.Errorf("scale-to-zero-cooldown needs to be at least 1ms: %s", opts.ScaleToZeroCooldown)
```
marcrittinghaus, line 747: That is no ideal, because there is a configurable lowest value. So if people try 100us, they get the error to use 1ms. Then they use 1ms and get an error that Xms is the minimum.

craciunoiuc: Yes, but I think the lowest it can be is still 1ms, so it's good to at least catch that

If the actual limit is higher they'll just get another error telling them the actual limit

marcrittinghaus (approved):

PR #1812, opened 2024-07-26
fix(cloud): Updating the help information

tatodorov:
### Prerequisite checklist

### Description of changes
- Updated help for scale-to-zero with the correct default value
   According to the help string, the default value is `on`, while in reality, it is `off`.
- Added short `-a` flag to remove all services
   The `kraft cloud inst rm` accepts long `--all` and short `-a` flags, but `kraft cloud svc rm` accepted only long `--all` flag.
- Updated the example for `kraft cloud svc create` and the malformed port error message
   The example used flag `-p` which doesn't exist and also the syntax was incorrect.
   Also, the malformed port message was incorrect.
- Updated the flag for attaching a volume to instance to singular `--volume`
   That will make it aligned with the examples below and the actual behavior as we can actually pass only one volume name to the flag.
- Added short flag `-a` to `kraft cloud inst stop` and `kraft cloud inst start`
- Added information about supported and default values for flags `--plat` and `--arch` of `kraft pkg`
- Added validation for the values set to `--plat` and `--arch` when used with `kraft pkg`
- Added validation for the values set to `--plat` and `--arch` when used with `kraft build`

### DIFF
--- a/internal/cli/kraft/build/build.go
+++ b/internal/cli/kraft/build/build.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 
@@ -27,6 +28,7 @@ import (
 	"kraftkit.sh/log"
 	"kraftkit.sh/packmanager"
 	"kraftkit.sh/unikraft/app"
+	"kraftkit.sh/unikraft/arch"
 	"kraftkit.sh/unikraft/target"
 )
 
@@ -34,7 +36,7 @@ var ErrContextNotBuildable = fmt.Errorf("could not determine what or how to buil
 
 type BuildOptions struct {
 	All          bool            `long:"all" usage:"Build all targets"`
-	Architecture string          `long:"arch" short:"m" usage:"Filter the creation of the build by architecture of known targets"`
+	Architecture string          `long:"arch" short:"m" usage:"Filter the creation of the build by architecture of known targets (x86_64/arm64/arm)"`
 	DotConfig    string          `long:"config" short:"c" usage:"Override the path to the KConfig .config file"`
 	Env          []string        `long:"env" short:"e" usage:"Set environment variables to be built in the unikernel"`
 	ForcePull    bool            `long:"force-pull" usage:"Force pulling packages before building"`
@@ -47,7 +49,7 @@ type BuildOptions struct {
 	NoFetch      bool            `long:"no-fetch" usage:"Do not run Unikraft's fetch step before building"`
 	NoRootfs     bool            `long:"no-rootfs" usage:"Do not build the root file system (initramfs)"`
 	NoUpdate     bool            `long:"no-update" usage:"Do not update package index before running the build"`
-	Platform     string          `long:"plat" short:"p" usage:"Filter the creation of the build by platform of known targets"`
+	Platform     string          `long:"plat" short:"p" usage:"Filter the creation of the build by platform of known targets (fc/qemu/xen)"`
 	PrintStats   bool            `long:"print-stats" usage:"Print build statistics"`
 	Project      app.Application `noattribute:"true"`
 	Rootfs       string          `long:"rootfs" usage:"Specify a path to use as root file system (can be volume or initramfs)"`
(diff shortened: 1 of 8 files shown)


internal/cli/kraft/pkg/pkg.go
```diff
@@ -35,7 +35,7 @@ import (
 )
 
 type PkgOptions struct {
-	Architecture string                    `local:"true" long:"arch" short:"m" usage:"Filter the creation of the package by architecture of known targets"`
```
craciunoiuc, line 38: The architecture is actually inferred from the host system. There is a detection check somewhere and that one is used.

I would not set a default here as we can't know for sure, but the text message is good.

internal/cli/kraft/pkg/pkg.go
```diff
@@ -49,7 +49,7 @@ type PkgOptions struct {
 	NoKConfig    bool                      `local:"true" long:"no-kconfig" usage:"Do not include target .config as metadata"`
 	NoPull       bool                      `local:"true" long:"no-pull" usage:"Do not pull package dependencies before packaging"`
 	Output       string                    `local:"true" long:"output" short:"o" usage:"Save the package at the following output"`
-	Platform     string                    `local:"true" long:"plat" short:"p" usage:"Filter the creation of the package by platform of known targets"`
+	Platform     string                    `local:"true" long:"plat" short:"p" usage:"Filter the creation of the package by platform of known targets (qemu/fc/kraftcloud)" default:"qemu"`
```
craciunoiuc: Same here, there is no clear default, so I would leave that out of the PR. The help text update is good though

internal/cli/kraft/pkg/pkg.go
```diff
@@ -62,6 +63,34 @@ type PkgOptions struct {
...
+	"fc":         true,
+	"kraftcloud": true,
+}
+
+// supportedArchitectures lists the supported architectures
+var supportedArchitectures = map[string]bool{
+	"x86_64": true,
+	"arm64":  true,
+	"arm":    true,
+}
```
craciunoiuc: I think you can find those somewhere, no need to create them here

internal/cli/kraft/build/build.go
```diff
@@ -47,7 +47,7 @@ type BuildOptions struct {
 	NoFetch      bool            `long:"no-fetch" usage:"Do not run Unikraft's fetch step before building"`
 	NoRootfs     bool            `long:"no-rootfs" usage:"Do not build the root file system (initramfs)"`
 	NoUpdate     bool            `long:"no-update" usage:"Do not update package index before running the build"`
-	Platform     string          `long:"plat" short:"p" usage:"Filter the creation of the build by platform of known targets"`
+	Platform     string          `long:"plat" short:"p" usage:"Filter the creation of the build by platform of known targets (qemu/fc)"`
```
craciunoiuc: I'd also add `xen` here to the list

internal/cli/kraft/build/build.go
```diff
@@ -59,6 +59,33 @@ type BuildOptions struct {
...
+	"qemu": true,
+	"fc":   true,
+}
+
+// supportedArchitectures lists the supported architectures
+var supportedArchitectures = map[string]bool{
+	"x86_64": true,
+	"arm64":  true,
+	"arm":    true,
+}
```
craciunoiuc: Same here, no need to add them, they are already somewhere, I'll link.

craciunoiuc (changes requested):
Hey @tatodorov,

Nice fixes you got here, I left you some comments, but other than that it looks pretty good, thanks a lot!

internal/cli/kraft/pkg/pkg.go
craciunoiuc: file is `machine/platform/platform.go` you can use the stuff in there

also for the architectures it is `unikraft/arch/architecture.go`

internal/cli/kraft/build/build.go
craciunoiuc: file is `machine/platform/platform.go` you can use the stuff in there

also for the architectures it is `unikraft/arch/architecture.go`

internal/cli/kraft/pkg/pkg.go
tatodorov: Thanks!
Already corrected.

internal/cli/kraft/pkg/pkg.go
tatodorov: Thanks!
Already corrected.

internal/cli/kraft/pkg/pkg.go
tatodorov: Thanks!
Already using them from the corresponding packages.

craciunoiuc (approved):
All good here. Thanks!

PR #1899, opened 2024-09-23
feat(internal): Pass 8 CSPRNG random integer words

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/internal/cli/kraft/run/runner_kraftfile_runtime.go
+++ b/internal/cli/kraft/run/runner_kraftfile_runtime.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/klauspost/cpuid"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	machineapi "kraftkit.sh/api/machine/v1alpha1"
@@ -24,6 +25,7 @@ import (
 	"kraftkit.sh/tui/selection"
 	"kraftkit.sh/unikraft/app"
 	ukarch "kraftkit.sh/unikraft/arch"
+	"kraftkit.sh/unikraft/export/v0/ukrandom"
 	"kraftkit.sh/unikraft/target"
 )
 
@@ -353,6 +355,17 @@ func (runner *runnerKraftfileRuntime) Prepare(ctx context.Context, opts *RunOpti
 	} else if len(runtime.Command()) > 0 {
 		machine.Spec.ApplicationArgs = runtime.Command()
 	}
+
+	var kernelArgs []string
+	if !runtime.KConfig().AllNoOrUnset(
+		"CONFIG_LIBUKRANDOM",
+		"CONFIG_LIBUKRANDOM_CMDLINE_INIT",
+		"CONFIG_LIBUKRANDOM_LCPU",
+	) && !cpuid.CPU.Rdrand() {
+		kernelArgs = append(kernelArgs, ukrandom.ParamRandomSeed.WithValue(ukrandom.NewRandomSeed()).String())
+	}
+
+	machine.Spec.KernelArgs = kernelArgs
 
 	// If automounting is enabled, and an initramfs is provided, set it as a
 	// volume if a initram has been provided.
--- a/internal/cli/kraft/run/runner_kraftfile_runtime.go
+++ b/internal/cli/kraft/run/runner_kraftfile_runtime.go
@@ -10,7 +10,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/klauspost/cpuid"
+	"github.com/klauspost/cpuid/v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	machineapi "kraftkit.sh/api/machine/v1alpha1"
@@ -357,12 +357,18 @@ func (runner *runnerKraftfileRuntime) Prepare(ctx context.Context, opts *RunOpti
 	}
 
 	var kernelArgs []string
-	if !runtime.KConfig().AllNoOrUnset(
+	hasUkRandom := !runtime.KConfig().AllNoOrUnset(
 		"CONFIG_LIBUKRANDOM",
-		"CONFIG_LIBUKRANDOM_CMDLINE_INIT",
-		"CONFIG_LIBUKRANDOM_LCPU",
-	) && !cpuid.CPU.Rdrand() {
+	)
+	hasCmdlineSupport := !runtime.KConfig().AllNoOrUnset(
+		"CONFIG_LIBUKRANDOM_CMDLINE_SEED",
+	)
+	hasNoCpuRandomnessSupport := runtime.KConfig().AllNoOrUnset("CONFIG_LIBUKRANDOM_LCPU") ||
+		!(cpuid.CPU.Has(cpuid.RDRAND) || cpuid.CPU.Has(cpuid.RNDR))
+	if hasUkRandom && hasNoCpuRandomnessSupport && hasCmdlineSupport {
 		kernelArgs = append(kernelArgs, ukrandom.ParamRandomSeed.WithValue(ukrandom.NewRandomSeed()).String())
+	} else if hasUkRandom && hasNoCpuRandomnessSupport && !hasCmdlineSupport {
+		log.G(ctx).Warn("RDRAND is not supported by the host CPU to be able to run Unikraft v0.17.0 and greater with CPU-generated randomness")
 	}
 
 	machine.Spec.KernelArgs = kernelArgs
--- a/internal/cli/kraft/run/runner_kraftfile_unikraft.go
+++ b/internal/cli/kraft/run/runner_kraftfile_unikraft.go
@@ -11,12 +11,14 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/klauspost/cpuid"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	machineapi "kraftkit.sh/api/machine/v1alpha1"
 	volumeapi "kraftkit.sh/api/volume/v1alpha1"
 	"kraftkit.sh/config"
 	"kraftkit.sh/unikraft/app"
+	"kraftkit.sh/unikraft/export/v0/ukrandom"
 	"kraftkit.sh/unikraft/target"
 )
 
@@ -147,6 +149,12 @@ func (runner *runnerKraftfileUnikraft) Prepare(ctx context.Context, opts *RunOpt
 		"CONFIG_LIBVFSCORE_AUTOMOUNT_CI_EINITRD",
 	)
 
+	noRandom := t.KConfig().AllNoOrUnset(
+		"CONFIG_LIBUKRANDOM",
+		"CONFIG_LIBUKRANDOM_CMDLINE_INIT",
+		"CONFIG_LIBUKRANDOM_LCPU",
+	) && !cpuid.CPU.Rdrand()
+
 	if runner.project.Rootfs() != "" && opts.Rootfs == "" && noEmbedded {
 		opts.Rootfs = runner.project.Rootfs()
 	}
@@ -178,6 +186,10 @@ func (runner *runnerKraftfileUnikraft) Prepare(ctx context.Context, opts *RunOpt
 			continue
 		}
 		appArgs = append(appArgs, arg)
+	}
+
+	if !noRandom {
+		kernelArgs = append(kernelArgs, ukrandom.ParamRandomSeed.WithValue(ukrandom.NewRandomSeed()).String())
 	}
 
 	machine.Spec.KernelArgs = kernelArgs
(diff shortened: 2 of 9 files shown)


unikraft/export/v0/ukrandom/params.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"crypto/rand"
+	"fmt"
+	"strings"
+
+	"kraftkit.sh/unikraft/export/v0/ukargparse"
+)
+
+var ParamRandomBytes = ukargparse.NewParamStrSlice("random", "seed", nil)
```
nderjung: Let's call this var, to reflect its name in the cmdline, just `ParamRandomSeed`.

Subsequent changes are then necessary to the constructor.

nderjung (approved):
All good on my side implementation-wise. Thanks!

Regarding the interaction with Unikraft itself, I'd like to cc @razvand and @michpappas before merging to help verify everything in conjunction with https://github.com/unikraft/unikraft/pull/1496

craciunoiuc:
Updated to check for RDRAND/RDSEED

If they are both present then ignore the config option and let Unikraft take its own bytes from the device/generator

unikraft/export/v0/ukrandom/params.go
```diff
@@ -0,0 +1,54 @@
...
+	return random
+}
+
+// String implements fmt.Stringer and returns a valid set of random bytes.
+func (rng RandomSeed) String() string {
+	var sb strings.Builder
+
+	sb.WriteString("[ ")
+	for i := 0; i < 8; i++ {
+		sb.WriteString(fmt.Sprintf("%04x ", rng[i]))
```
michpappas: You need to prepend each number with `0x` otherwise they won't be interpreted by `uklibparam` neither as valid hex, nor as decimals.

internal/cli/kraft/run/runner_kraftfile_runtime.go
```diff
@@ -358,6 +360,16 @@ func (runner *runnerKraftfileRuntime) Prepare(ctx context.Context, opts *RunOpti
 		machine.Spec.ApplicationArgs = runtime.Command()
 	}
 
+	var kernelArgs []string
+	if !runtime.KConfig().AllNoOrUnset(
+		"CONFIG_LIBUKRANDOM",
+		"CONFIG_LIBUKRANDOM_CMDLINE_INIT",
+	) && !(cpuid.CPU.Rdrand() && cpuid.CPU.Rdseed()) {
```
michpappas: There is a theoretical case where a CPU has `RDSEED` but not `RDRAND`, but I haven't come across such a thing in practice, so I'm not sure if it's worth adding a check. Saying that ,the opposite is possible: `RDRAND` was introduced before `RDSEED`, so there are CPUs out there that have the former but not the latter. The implementation in Unikraft has the minimum requirement of  `RDRAND` when the driver is enabled (seeding functions fall back to `RDRAND`), so in theory you may skip the check for `RDSEED` altogether.

Another thing related to portability: I had a quick look at https://github.com/klauspost/cpuid and it looks like it lacks detection for `FEAT_RNG` on Arm (along with many others). You may consider opening a PR upstream to add these so that they're there by the time we need them.

michpappas (changes requested):
@craciunoiuc just a few comments. Also if you update this, it would be great if you posted a few of the generated cmdlines.

internal/cli/kraft/run/runner_kraftfile_runtime.go
craciunoiuc: noted, will do

craciunoiuc:
```
qemu-system-x86_64 -append random.seed=[ 0xef59fa23 0x1eba1280 0xb0972ca2 0xc734f02f 0x2079dc04 0x158e0490 0x656d5095 0xa407c115 ] -- /usr/bin/nginx  -cpu qemu64,+pdpe1gb,+rdrand,+rdseed,-vmx,-svm -daemonize -device virtio-net-pci,mac=02:b0:b0:84:3e:01,netdev=hostnet0 -device pvpanic -display none -kernel /home/cez/.local/share/kraftkit/runtime/99876f61-99a0-4072-a2f8-5b6483635e20/unikraft/bin/kernel -machine pc -m size=128M -monitor unix:/home/cez/.local/share/kraftkit/runtime/99876f61-99a0-4072-a2f8-5b6483635e20/qemu_mon.sock,server,nowait -name 99876f61-99a0-4072-a2f8-5b6483635e20 -netdev user,id=hostnet0,hostfwd=tcp::8081-:80 -nographic -no-reboot -S -parallel none -pidfile /home/cez/.local/share/kraftkit/runtime/99876f61-99a0-4072-a2f8-5b6483635e20/machine.pid -qmp unix:/home/cez/.local/share/kraftkit/runtime/99876f61-99a0-4072-a2f8-5b6483635e20/qemu_control.sock,server,nowait -qmp unix:/home/cez/.local/share/kraftkit/runtime/99876f61-99a0-4072-a2f8-5b6483635e20/qemu_events.sock,server,nowait -rtc base=utc -serial file:/home/cez/.local/share/kraftkit/runtime/99876f61-99a0-4072-a2f8-5b6483635e20/machine.log -smp cpus=1,threads=1,sockets=1 -vga none
```

this is how it looks for one run, note that quotes don't show up in the logs

michpappas (approved):
Looks good, thanks a lot!

Reviewed-by: Michalis Pappas <michalis@unikraft.io>

nderjung (approved):
Thanks everyone! Looks good from my side. I'll merge this when https://github.com/unikraft/unikraft/pull/1496 is merged into staging.

nderjung:
Linking for completeness: https://github.com/klauspost/cpuid/pull/152

internal/cli/kraft/run/runner_kraftfile_runtime.go
```diff
@@ -354,6 +356,17 @@ func (runner *runnerKraftfileRuntime) Prepare(ctx context.Context, opts *RunOpti
 		machine.Spec.ApplicationArgs = runtime.Command()
 	}
 
+	var kernelArgs []string
+	if (!runtime.KConfig().AllNoOrUnset("CONFIG_LIBUKRANDOM_LCPU") || !cpuid.CPU.Has(cpuid.RDRAND)) &&
```
michpappas: The 1st condition here looks incorrect. It should check if `CONFIG_LIBUKRANDOM_LCPU` is *not* enabled.

internal/cli/kraft/run/runner_kraftfile_unikraft.go
```diff
@@ -147,6 +149,12 @@ func (runner *runnerKraftfileUnikraft) Prepare(ctx context.Context, opts *RunOpt
 		"CONFIG_LIBVFSCORE_AUTOMOUNT_CI_EINITRD",
 	)
 
+	noRandom := t.KConfig().AllNoOrUnset("CONFIG_LIBUKRANDOM_LCPU") && cpuid.CPU.Has(cpuid.RDRAND) ||
```
michpappas: This one looks like different logic altoghether 🤔

internal/cli/kraft/run/runner_package.go
```diff
@@ -338,6 +340,17 @@ func (runner *runnerPackage) Prepare(ctx context.Context, opts *RunOptions, mach
 		})
 	}
 
+	var kernelArgs []string
+	if (!targ.KConfig().AllNoOrUnset("CONFIG_LIBUKRANDOM_LCPU") || !cpuid.CPU.Has(cpuid.RDRAND)) &&
```
michpappas: This one too should check if `CONFIG_LIBUKRANDOM_LCPU` is *not* enabled.

machine/firecracker/v1alpha1.go
```diff
@@ -87,8 +87,8 @@ func (service *machineV1alpha1Service) Create(ctx context.Context, machine *mach
 		return machine, fmt.Errorf("cannot create firecracker instance with emulation")
 	}
 
-	if !cpuid.CPU.Rdrand() || !cpuid.CPU.Rdseed() {
-		log.G(ctx).Warn("RDRAND and RDSEED are not supported by the host CPU to be able to run Unikraft v0.17.0 and greater with hardware randomization")
+	if !cpuid.CPU.Has(cpuid.RDRAND) {
+		log.G(ctx).Warn("RDRAND is not supported by the host CPU to be able to run Unikraft v0.17.0 and greater with hardware randomization")
```
michpappas: Since you're updating this line, the term normally used is "randomness", not "randomization". Or if you want to be more precise, "CPU-generated randomness".

machine/qemu/v1alpha1.go
```diff
@@ -438,8 +437,8 @@ func (service *machineV1alpha1Service) Create(ctx context.Context, machine *mach
 				}),
 			)
 		} else {
-			if !cpuid.CPU.Rdrand() || !cpuid.CPU.Rdseed() {
-				log.G(ctx).Warn("RDRAND and RDSEED are not supported by the host CPU, try rerunning with emulation '-W' to be able to run Unikraft v0.17.0 and greater with hardware randomization")
+			if !cpuid.CPU.Has(cpuid.RDRAND) {
+				log.G(ctx).Warn("RDRAND is not supported by the host CPU, try rerunning with emulation '-W' to be able to run Unikraft v0.17.0 and greater with hardware randomization")
```
michpappas: Same comment about the message, as above.

michpappas (changes requested):
@craciunoiuc besides the comments below, I think we should have the conditions on for generating the cmdline rewritten in a more clear / readable way, especially given the security-related nature of that code 🙏🏼 

Also since your changes for `FEAT_RNG` were merged upstream, please consider if you should also update the CPUID check for Arm too.

internal/cli/kraft/run/runner_kraftfile_unikraft.go
craciunoiuc: did some logic toke reduce the number of negations, will simplfy in another way

craciunoiuc:
Checks are automatically done when checking for the RDRAND features. Imagine it as a an interface with different implementations depending on the OS running the library. So all is good.

internal/cli/kraft/run/runner_kraftfile_runtime.go
```diff
@@ -354,6 +356,18 @@ func (runner *runnerKraftfileRuntime) Prepare(ctx context.Context, opts *RunOpti
 		machine.Spec.ApplicationArgs = runtime.Command()
 	}
 
+	var kernelArgs []string
+	hasCmdlineSupport := !runtime.KConfig().AllNoOrUnset(
+		"CONFIG_LIBUKRANDOM",
```
michpappas, line 361: These are merely "I would had done it differently" / in  `unikraft` we do it differently,  so totally up to you to take into account or ignore. I only care about functional correctness, and the current implementation looks good in that aspect :)

1. I would had written this a bit differently, having a separate variable for the presence of `CONFIG_LIBUKRANDOM`, and then do the check for that at the final condition:
```go
if libUKRandom && hasNoCpuRandomnessSupport && hasCmdlineSupport {
```
 The reason is that there there is no stronger correlation between `CONFIG_LIBUKRANDOM` and `CONFIG_LIBUKRANDOM_CMDLINE_INIT` than `CONFIG_LIBUKRANDOM_CMDLINE_LCPU`.
 
2. I would had also moved this logic into a function to avoid copying (and keeping up to date) into three different places.

3. Looking at the individual commits, it seems that some changes are introduced in one commit just to be updated (read corrected) in another, which people may frown up.

machine/firecracker/v1alpha1.go
```diff
@@ -87,8 +87,8 @@ func (service *machineV1alpha1Service) Create(ctx context.Context, machine *mach
 		return machine, fmt.Errorf("cannot create firecracker instance with emulation")
 	}
 
-	if !cpuid.CPU.Rdrand() || !cpuid.CPU.Rdseed() {
-		log.G(ctx).Warn("RDRAND and RDSEED are not supported by the host CPU to be able to run Unikraft v0.17.0 and greater with hardware randomization")
+	if !cpuid.CPU.Has(cpuid.RDRAND) {
```
michpappas: Should these be additionally conditional to `!hasNoCpuRandomnessSupport`?

internal/cli/kraft/run/runner_kraftfile_runtime.go
craciunoiuc: 1. I can do it, sure
2. If I do it I need to do it in other places too, so I'll skip this one 😆 
3. Yeah, I was a bit split on this also, but I think the bump should be done separately as it has significant changes

machine/firecracker/v1alpha1.go
craciunoiuc: Makes sense

craciunoiuc: Actually, symbols are not accessible in there, I will move the warning closer to that

michpappas (approved):
Looks good, thanks!

Reviewed-by: Michalis Pappas <michalis@unikraft.io>

razvand (approved):

PR #2007, opened 2024-12-10
feat!: Introduce support for EROFS initrds

craciunoiuc:
### Prerequisite checklist

### Description of changes

Remaining ideas to discuss once @nderjung finds some time:
- ~~worth creating/finding an erofs library to handle packaging? (my opinion: for now, no, if use case becomes wide-spread and unwieldly, then start doing it)~~
- ~~worth creating subdirectories for initrd implementations? Right now we have `CPIO`, and `EROFS`, but we might have more in the future, and might need those fancy interfaces + packages + structures (my opinion: yes, but for this very early use case, we can keep it like this)~~
- worth writing a `CPIO->EROFS` converter? Right now it throws an error and informs people they should unpack with `cpio`/`bsdcpio` first and then run on the directory (my opinion: no, not needed; only people who would turn CPIO files to EROFS files most probably already have these tools, meaning they are devs). **Still open to discussion but it can be done in a later iteration.**

To compile and test, you will require at least go `1.25.0`


craciunoiuc:
Action fail is expected, it adds new interface elements that are not available in the action, but can't be bumped either

nderjung:
> worth creating/finding an erofs library to handle packaging? (my opinion: for now, no, if use case becomes wide-spread and unwieldly, then start doing it)

Yes. Right now this invokes `mkfs.erofs` which won't work on other systems, i.e. macOS.  For internal testing and validation this implementation is fine but we should iterate (see points below).

> worth creating subdirectories for initrd implementations? Right now we have CPIO, and EROFS, but we might have more in the future, and might need those fancy interfaces + packages + structures (my opinion: yes, but for this very early use case, we can keep it like this)

Yes. It's better to have an interface which we can use to to output different formats (which is true given our roadmap) in the future.  Tarball, squashfs, etc.

> worth writing a CPIO->EROFS converter? Right now it throws an error and informs people they should unpack with cpio/bsdcpio first and then run on the directory (my opinion: no, not needed; only people who would turn CPIO files to EROFS files most probably already have these tools, meaning they are devs)

No.  Each initrd "source" should emit the resulting filesystem as an intermediate filesystem representation.  This intermediate filesystem representation should be passable into a converter which serializes appropriately based on its relevant headers/format.  This will similarly be true for a de-serialization process which would allow the user to convert a given output format back into the intermediate filesystem representaton and thus into another format.

craciunoiuc:
For 1. I don't think it's that bad. I could find the C code that is built into the `mkfs.erofs` binary, so theoretically we could "translate" it, but maybe we can pass this to a GSoC student or similar :D

upsaurav12:
Hi UniKraft Team,
I am Saurav Upadhyay , I would like to show my interest in contributing to "Converting the eroFS library to Golang and testing it" project. I always wanted to contribute on project requiring Golang/C  found this project. 
I am learning Golang and actively contributing to CNCF project for writing unit tests in golang. I know basic knowledge of C but converting C to Golang code might be challenging for me but I do my best. 

I would like to know your opinion and suggestions about this project's priorities
Thank you, 
Saurav Upadhyay.

gdochadipa:
Hi UniKraft Team,
My name is Gede Ocha, I would like to contributing on issue "Converting the eroFS library to Golang and testing it". But I have a question. On main EROFS lib, they implement config by modifying the Linux kernel directly. In this new implement are will implement config file also? if yes, what approach setting will set for config file ?

I would like to know your opinion and suggestions about this project's priorities
Thank you

PsychoPunkSage:
Hi `@Unikraft Team`,

Over the past two months, I’ve been working on a Go implementation of EROFS, now available as an MVP here: [ErgoFS](https://github.com/PsychoPunkSage/ErgoFS). This project was inspired by one of the ideas proposed for GSoC (although it was ultimately not selected as an official GSoC project). Nonetheless, I invested significant time and effort into building it, and I wanted to share it here so that the work doesn't go unnoticed.

### ✅ Current Status of ErgoFS:

* **Core Functionality**:

  * Superblock creation is now functional and matches the official EROFS behavior.
  * Key data structures like Inode, Superblock, BlockHeader, and Xattr have been implemented.
  * Support for compression (initially LZ4) is in place, including modular compression builders and macros.

* **Recent Additions**:

  * Fragment deduplication logic using CRC32C and syscall-based comparisons.
  * Compression hints via regex-based matching for dynamic configuration.
  * Chunked file reads in `ErofsIoRead` to safely handle large data.
  * CRC32C checksum calculation via `ErofsGetCrc32c`.

* **Architecture Enhancements**:

  * Introduced intermediate structs like `SuperBlockOnDisk` for clean separation of runtime and disk formats.
  * Converted C macros to idiomatic Go functions to preserve logic while maintaining language safety.
  * Switched from `google/uuid` to a custom UUID aligned with EROFS for better consistency.

* **Work in Progress**:

  * Inode logic: handling creation, disk reads, and compressed formats.
  * Integration of compression context (`ZErofsCompressIctx`) for upcoming pipeline support.
  * Modularization and project restructuring for maintainability and contribution-readiness.

I plan to continue refining this implementation if the time permits. Any feedback, suggestions, or contributions from the community are more than welcome—especially if anyone is interested in evolving this toward a C-agnostic, production-ready EROFS implementation in Go.

Thanks for your time and consideration!

— [@PsychoPunkSage](https://github.com/PsychoPunkSage)

erofs/constants.go
```diff
@@ -0,0 +1,93 @@
+// SPDX-License-Identifier: MPL-2.0
+/*
+ * Copyright (C) 2024 Damian Peckett <damian@pecke.tt>.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
```
nderjung, line 8: In this first commit, can you please also add a Git SHA to the URL so that the link is porcelained?

erofs/writer.go
```diff
@@ -143,8 +150,34 @@ func (w *writer) firstPass() (metaSize, dataSize int64, err error) {
...
-			ino.Ino = nid
-			ino.Size = uint32(size)
+			if ino.Mode&S_IFMT == S_IFDIR {
+				ino.Ino = nid
+				ino.Size = uint32(size)
+			} else {
+				info, err := os.Lstat(filepath.Join(w.rootPath, path))
+				if err != nil {
+					return metaSize, dataSize, fmt.Errorf("failed to stat file %q: %w", path, err)
+				}
```
nderjung:
```diff
-                               info, err := os.Lstat(filepath.Join(w.rootPath, path))
+                               f, ok := w.src.(fs.StatFS)
+                               if !ok {
+                                       return metaSize, dataSize, fmt.Errorf("source filesystem does not support stat")
+                               }
+
+                               info, err := f.Stat(path)
```

erofs/writer.go
```diff
@@ -154,8 +187,33 @@ func (w *writer) firstPass() (metaSize, dataSize int64, err error) {
 			w.inodes[path] = ino
 
 		case InodeExtended:
-			ino.Ino = nid
-			ino.Size = uint64(size)
+			if ino.Mode&S_IFMT == S_IFDIR {
+				ino.Ino = nid
+				ino.Size = uint64(size)
+			} else {
+				info, err := os.Lstat(filepath.Join(w.rootPath, path))
```
nderjung:
```diff
-                               info, err := os.Lstat(filepath.Join(w.rootPath, path))
+                               f, ok := w.src.(fs.StatFS)
+                               if !ok {
+                                       return metaSize, dataSize, fmt.Errorf("source filesystem does not support stat")
+                               }
+
+                               info, err := f.Stat(path)
```

erofs/writer.go
```diff
@@ -35,16 +34,18 @@ const (
 
 // Create creates an EROFS filesystem image from the source filesystem and writes
 // it to the destination writer.
-func Create(dst io.WriterAt, src fs.FS) error {
+func Create(dst io.WriterAt, src string) error {
```
nderjung: This change to the public method prototype changes the contract with the imported implementation, resulting in a breaking change.  I think we should not mix between string and interface methods as prototypes, it's either both strings or both interfaces.  Since it's better to not break the compatibility with this package, we should see if we can fix the issue without breaking this contract.

I think the issue that you're trying to solve in this commit is that the `ReadlinkFS` provider is no longer an interface method within the Go runtime (I'm not sure if it was previously deprecated).

[The `os.DirFS` documentation](https://pkg.go.dev/os#DirFS) states that it "implements [io/fs.StatFS](https://pkg.go.dev/io/fs#StatFS), [io/fs.ReadFileFS](https://pkg.go.dev/io/fs#ReadFileFS) and [io/fs.ReadDirFS](https://pkg.go.dev/io/fs#ReadDirFS)".  This means that we can safely detect it by casting to these types around the codebase in order to retrieve what want (see later comments).

erofs/writer.go
```diff
@@ -34,11 +34,12 @@ const (
 
 // Create creates an EROFS filesystem image from the source filesystem and writes
 // it to the destination writer.
-func Create(dst io.WriterAt, src string) error {
+func Create(dst io.WriterAt, src string, allRoot bool) error {
```
nderjung: Similarly, the `allRoot` option breaks the prototype.  Instead, let's introduce optional (variadic) method options in the form:
```go
func Create(dst io.WriterAt, src fs.FS, opts ...ErofsCreateOption) error
```

Then introduce the option `WithAllRoot(true)` option which applies the internal `allRoot` flag.

erofs/writer.go
```diff
@@ -303,6 +361,14 @@ func (w *writer) populateInodes() error {
...
 
 			nlink = len(entries) + 2
+		} else {
+			nlink = getNLinks(fi)
+			ino := getIno(fi)
+			if _, ok := fsToErofsLinkMap[ino]; !ok {
+				fsToErofsLinkMap[ino] = inodeCount{
+					Count: 0,
+				}
+			}
```
nderjung, line 390: This addition must also be added to `fs.IsDir()` codepath, directories also have inodes.

Makefile
```diff
@@ -269,7 +269,9 @@ test: test-unit test-framework test-e2e test-cloud-e2e ## Run all tests.
 test-unit: GOTEST_EXCLUDE := third_party/ test/ hack/ buildenvs/ dist/ docs/ tools/
 test-unit: GOTEST_PKGS := $(foreach pkg,$(filter-out $(GOTEST_EXCLUDE),$(wildcard */)),$(pkg)...)
 test-unit: ## Run unit tests.
-	$(GINKGO) -v -p -randomize-all -tags "containers_image_storage_stub,containers_image_openpgp,osusergo,netgo" $(GOTEST_PKGS)
+	tar xpzf $(WORKDIR)/erofs/testdata/testfs.tar.gz -C $(WORKDIR)/erofs/testdata
+	$(GINKGO) -v -p -randomize-all -tags "containers_image_storage_stub,containers_image_openpgp,osusergo,netgo" $(GOTEST_PKGS) || true
+	rm -rf $(WORKDIR)/erofs/testdata/testfs
```
nderjung: I'm not a fan of ephemeral changes to the source directory as they can be hard to detect.

Instead, let's migrate all the test-stuff directly into the `initrd` package -- as it then is homogeneous across all filesystem implementations.  The files are already laid out there as a tarball and Dockerfile which we can leverage.

In a subsequent comment I'll suggest formatting.

Makefile
```diff
@@ -76,7 +76,7 @@ YTT_VERSION        ?= v0.51.0
 YTT                ?= $(GO) run carvel.dev/ytt/cmd/ytt@$(YTT_VERSION)
 GORELEASER_VERSION ?= v2.7.0
 GORELEASER         ?= $(GO) run github.com/goreleaser/goreleaser/v2@$(GORELEASER_VERSION)
-GINKGO_VERSION     ?= v2.22.2
+GINKGO_VERSION     ?= v2.23.3
```
nderjung: Update Git commit title with version change.

initrd/options.go
```diff
@@ -4,29 +4,69 @@
...
 }
 
-// WithOutput sets the location of the output location of the resulting CPIO
-// archive file.
-func WithOutput(output string) InitrdOption {
+// WithCompression sets the compression of the resulting CPIO archive file.
+func WithCompression(compress bool) InitrdOption {
 	return func(opts *InitrdOptions) error {
-		opts.output = output
+		opts.compress = compress
```
nderjung: It looks like you re-ordered the file, this makes git blame sad, can you please keep the function order?

initrd/file.go
```diff
@@ -60,7 +60,45 @@ func (initrd *file) Name() string {
 }
 
 // Build implements Initrd.
-func (initrd *file) Build(_ context.Context) (string, error) {
+func (initrd *file) Build(ctx context.Context) (string, error) {
+	if initrd.opts.fsType == FsTypeErofs {
```
nderjung: Let's use a `switch` here.

initrd/file.go
```diff
@@ -60,7 +60,45 @@ func (initrd *file) Name() string {
...
+		}
+
+		if isCpio {
+			return "", fmt.Errorf("CPIO-to-EROFS conversion currently not supported. Use 'bsdcpio' or 'cpio' to unpack it first to a directory")
+		}
+
+		if !isErofs {
+			return "", fmt.Errorf("file %s is not a valid EROFS archive", initrd.path)
+		}
+	} else if initrd.opts.fsType == FsTypeCpio {
```
nderjung: `switch` case

initrd/utils.go
```diff
@@ -55,3 +61,147 @@ func compressFiles(output string, writer *cpio.Writer, reader *os.File) error {
 
 	return nil
 }
+
+func convertToErofs(output string, source string, isTar, allRoot bool) error {
+	// Extract the archive to a temporary directory
```
nderjung: It looks like the large `if isTar` code path is reversing the earlier change you were doing with `Create` and changing this into a `string` input.  Given the previous suggestion this should no longer be necessary and the `if isTar` should be simpler.

However, I am in favour of totally removing this function and replacing in-line `erofs.Create` where this method is called (+ preamble).

initrd/utils.go
```diff
@@ -55,3 +61,147 @@ func compressFiles(output string, writer *cpio.Writer, reader *os.File) error {
...
+		tarReader = tar.NewReader(gzReader)
+	} else {
+		tarReader = tar.NewReader(file)
+	}
+	_, err = tarReader.Next()
+	if err == nil || err == tar.ErrHeader {
+		isTar = true
+	}
+	return
+}
```
nderjung: Redundant method `isTarGzFile`, this already exists here:

https://github.com/unikraft/kraftkit/blob/staging/archive/unarchive.go#L46

internal/cli/kraft/pkg/pkg.go
```diff
@@ -35,27 +36,29 @@ import (
...
+	Architecture   string                    `local:"true" long:"arch" short:"m" usage:"Filter the creation of the package by architecture of known targets (x86_64/arm64/arm)"`
+	Args           []string                  `local:"true" long:"args" short:"a" usage:"Pass arguments that will be part of the running kernel's command line"`
+	Compress       bool                      `local:"true" long:"compress" short:"c" usage:"Compress the initrd package (experimental)"`
+	Dbg            bool                      `local:"true" long:"dbg" usage:"Package the debuggable (symbolic) kernel image instead of the stripped image"`
+	Env            []string                  `local:"true" long:"env" short:"e" usage:"Set environment variables to be packed into the package" split:"false"`
+	Force          bool                      `local:"true" long:"force-format" usage:"Force the use of a packaging handler format"`
+	Format         string                    `local:"true" long:"as" short:"M" usage:"Force the packaging despite possible conflicts" default:"oci"`
+	KeepFileOwners bool                      `local:"true" long:"keep-file-owners" usage:"Keep file owners in the rootfs (default root:root)"`
```
nderjung: Maybe:
```
Keep file owners (user:group) in the rootfs (false sets 'root:root')
```
?

oci/pack.go
```diff
@@ -748,6 +748,10 @@ func (ocipack *ociPackage) Unpack(ctx context.Context, dir string) error {
 			initrdPath,
 			initrd.WithArchitecture(image.Architecture),
 			initrd.WithWorkdir(dir),
+
+			// We don't know the filesystem type of the initrd, so we set it to
+			// unknown. We let the initrd subsystem detect it.
+			initrd.WithType(initrd.FsTypeUnknown),
```
nderjung: This can be removed, it's not true that "unknown" is checked, it simply defaults to CPIO (which is correct), but this is not necessary.

initrd/directory.go
```diff
@@ -77,6 +77,10 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
 		return "", fmt.Errorf("could not create output directory: %w", err)
 	}
 
+	if initrd.opts.fsType == FsTypeErofs {
+		return initrd.opts.output, convertToErofs(initrd.opts.output, initrd.path, false, !initrd.opts.keepOwners)
+	}
+
```
nderjung: This needs some thinking as it feels a bit of a hack to add this before a large chunk of code which does this for a different file system type. It's better to consolidate the two different file system types into interfaces or something and homogeneously `switch` between them depending on input format.

erofs/writer.go
craciunoiuc: No need for now, as this is done only for creating hardlinks, and on Unix directories do not support hardlinks

erofs/writer.go
craciunoiuc: As discussed privately, in order to fix this we need to use `gotip` until August, as support for this is yet to be added in go.

initrd/utils.go
craciunoiuc: Added this by mistake

initrd/options.go
craciunoiuc: I ordered* as it was not ordered before

I can just restore the old order

nderjung: If you're ordering something, please make this its own commit, then later introduce new content in a separate commit.

Makefile
craciunoiuc: Fixed in subsequent commit

Makefile
```diff
@@ -269,7 +269,7 @@ test: test-unit test-framework test-e2e test-cloud-e2e ## Run all tests.
 test-unit: GOTEST_EXCLUDE := third_party/ test/ hack/ buildenvs/ dist/ docs/ tools/
 test-unit: GOTEST_PKGS := $(foreach pkg,$(filter-out $(GOTEST_EXCLUDE),$(wildcard */)),$(pkg)...)
 test-unit: ## Run unit tests.
-	$(GINKGO) -v -p -randomize-all -tags "containers_image_storage_stub,containers_image_openpgp,osusergo,netgo" $(GOTEST_PKGS)
+	$(GINKGO) -v -p -randomize-all -tags "containers_image_storage_stub,containers_image_openpgp,osusergo,netgo" $(GOTEST_PKGS) || true
```
nurof3n: please revert change before merging

cpio/cpio_options.go
```diff
@@ -0,0 +1,20 @@
...
+package cpio
+
+type CpioCreateOptions struct {
+	allRoot bool
+}
+
+type CpioCreateOption func(*CpioCreateOptions) error
+
+// WithAllRoot indicates that all files permissions should be set to root:root
+// instead of the original file permissions.
```
nurof3n, line 14:
```
// WithAllRoot toggles whether all files permissions should be set to root:root
// instead of the original file permissions.
```

cpio/create.go
```diff
@@ -0,0 +1,554 @@
...
+			return fmt.Errorf("could not create CPIO archive from directory: %w", err)
+		}
+	case IsCpioFile(source):
+		if err := c.CreateFSFromCpio(ctx, writer, source); err != nil {
+			return fmt.Errorf("could not create CPIO archive from CPIO file: %w", err)
+		}
+	// TODO: Decide if we want to stub this or just hide completely
+	// Stubbing would require duplicating the function already existing in erofs/utils.go
+	// We can't import as it would cause a circular dependency
+	// case IsErofsFile(source):
```
nurof3n: idea: it might be better to structure `cpio` and `erofs` inside a `fs` package, and have common utils thrown in there instead

cpio/create.go
```diff
@@ -0,0 +1,554 @@
...
+			}
+		}
+	}
+
+	// Close the tarball and re-open it to read it again.
+	if err := close(); err != nil {
+		return err
+	}
+
+	_, err = tarArchive.Seek(0, io.SeekStart)
```
nurof3n, line 315: checks are already done when creating the reader

cpio/create.go
```diff
@@ -0,0 +1,554 @@
...
+		}
+		if err != nil {
+			return fmt.Errorf("could not read tar header on first pass: %w", err)
+		}
+
+		if tarHeader.Typeflag == tar.TypeLink {
+			if _, ok := fileCount[tarHeader.Linkname]; !ok {
+				fileCount[tarHeader.Linkname] = inodeCount{
+					Count: 1,
+					Inode: rand.Int32(),
```
nurof3n: might be a better idea to start from a large number (maybe avoiding conflicts with first inodes, idk) and then go sequentially from there

cpio/create.go
```diff
@@ -0,0 +1,554 @@
...
+				fileCount[tarHeader.Linkname] = inodeCount{
+					Count: fileCount[tarHeader.Linkname].Count + 1,
+					Inode: fileCount[tarHeader.Linkname].Inode,
+				}
+			}
+		} else if tarHeader.Typeflag == tar.TypeReg {
+			if _, ok := fileCount[tarHeader.Name]; !ok {
+				fileCount[tarHeader.Name] = inodeCount{
+					Count: 1,
+					Inode: rand.Int32(),
```
nurof3n: ditto

cpio/utils.go
nurof3n, line 1: maybe move these functions to an `fs` package

erofs/create.go
```diff
@@ -0,0 +1,113 @@
...
+	source, fInfoMap, err := unpackOCIImageToDirectory(ctx, source)
+	if err != nil {
+		return fmt.Errorf("could not unpack OCI file: %w", err)
+	}
+	defer os.RemoveAll(source)
+
+	opts = append(opts, withFileInfoMap(fInfoMap))
+
+	if err := c.CreateFSFromDirectory(ctx, writer, source, opts...); err != nil {
+		return fmt.Errorf("could not create CPIO archive from directory: %w", err)
```
nurof3n: error is already wrapped with same message in `CreateFS` (and here it mentions `CPIO` by mistake)

erofs/create.go
```diff
@@ -0,0 +1,113 @@
...
+	source, fInfoMap, err := unpackTarFileToDirectory(ctx, source)
+	if err != nil {
+		return fmt.Errorf("could not unpack tar file: %w", err)
+	}
+	defer os.RemoveAll(source)
+
+	opts = append(opts, withFileInfoMap(fInfoMap))
+
+	if err := c.CreateFSFromDirectory(ctx, writer, source, opts...); err != nil {
+		return fmt.Errorf("could not create CPIO archive from directory: %w", err)
```
nurof3n: error is already wrapped with same message in `CreateFS` (and here it mentions `CPIO` by mistake)

erofs/create.go
```diff
@@ -0,0 +1,113 @@
...
+// CreateFSFromErofs creates an EroFS filesystem from an existing EroFS file.
+func (c *createOptions) CreateFSFromErofs(ctx context.Context, writer *os.File, source string, opts ...ErofsCreateOption) error {
+	// Open and copy all contents from 'source' to the writer
+	f, err := os.Open(source)
+	if err != nil {
+		return fmt.Errorf("could not open CPIO source file: %w", err)
+	}
+	defer f.Close()
+
+	log.G(ctx).Info("Creating EroFS archive")
```
nurof3n: I think this should not start with capital `C`

erofs/create.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return nil
+}
+
+// CreateFSFromErofs creates an EroFS filesystem from an existing EroFS file.
+func (c *createOptions) CreateFSFromErofs(ctx context.Context, writer *os.File, source string, opts ...ErofsCreateOption) error {
+	// Open and copy all contents from 'source' to the writer
+	f, err := os.Open(source)
+	if err != nil {
+		return fmt.Errorf("could not open CPIO source file: %w", err)
```
nurof3n: `could not open EroFS source file`

erofs/create.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	defer f.Close()
+
+	log.G(ctx).Info("Creating EroFS archive")
+
+	_, err = io.Copy(writer, f)
+	if err != nil {
+		return fmt.Errorf("could not copy CPIO data: %w", err)
```
nurof3n: could not copy EroFS data

erofs/erofs_options.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// Map of file information for each file in the Erofs archive
+	// This is used to set the uid, gid, and mode for each file.
+	fInfoMap map[string]fInfo
+}
+
+type ErofsCreateOption func(*ErofsCreateOptions) error
+
+// WithAllRoot indicates that all files permissions should be set to root:root
+// instead of the original file permissions.
```
nurof3n, line 24:
```
// WithAllRoot toggles whether all files permissions should be set to root:root
// instead of the original file permissions.
```

erofs/owner_unix.go
nurof3n, line 1: maybe move these functions to an `fs` package

erofs/owner_windows.go
nurof3n, line 1: maybe move these functions to an `fs` package

erofs/erofs.go
```diff
@@ -0,0 +1,321 @@
...
+
+	return &fileInfo{
+		image: de.image,
+		name:  de.name,
+		inode: ino,
+	}, nil
+}
+
+// ReadLink returns the destination of the named symbolic link.
+// Experimental implementation of: https://github.com/golang/go/issues/49580
```
nurof3n: remove comment since this has been moved upstream

erofs/erofs.go
```diff
@@ -0,0 +1,321 @@
...
+		return "", err
+	}
+
+	ino := de.getInode()
+
+	return ino.Readlink()
+}
+
+// StatLink returns a FileInfo describing the file without following any symbolic links.
+// Experimental implementation of: https://github.com/golang/go/issues/49580
```
nurof3n: remove comment since this has been moved upstream

erofs/utils.go
nurof3n, line 1: maybe move these functions to an `fs` package

initrd/directory.go
```diff
@@ -77,117 +76,28 @@ func (initrd *directory) Build(ctx context.Context) (string, error) {
...
+		return initrd.opts.output, erofs.CreateFS(ctx, initrd.opts.output, initrd.path,
+			erofs.WithAllRoot(!initrd.opts.keepOwners),
+		)
+	case FsTypeCpio:
+		err := cpio.CreateFS(ctx, initrd.opts.output, initrd.path,
+			cpio.WithAllRoot(!initrd.opts.keepOwners),
+		)
 		if err != nil {
-			return fmt.Errorf("received error before parsing path: %w", err)
+			return "", fmt.Errorf("could not create CPIO archive: %w", err)
```
nurof3n, line 91: might be redundant/inconsistent with erofs error message

initrd/dockerfile.go
```diff
@@ -568,201 +561,28 @@ func (initrd *dockerfile) Build(ctx context.Context) (string, error) {
...
-		}
-
-		internal := fmt.Sprintf("./%s", filepath.Clean(tarHeader.Name))
-
-		cpioHeader := &cpio.Header{
-			Name:    internal,
-			Mode:    cpio.FileMode(tarHeader.FileInfo().Mode().Perm()),
-			ModTime: tarHeader.FileInfo().ModTime(),
-			Size:    tarHeader.FileInfo().Size(),
+			return "", fmt.Errorf("could not create CPIO archive: %w", err)
```
nurof3n, line 576: ditto

initrd/file.go
```diff
@@ -60,8 +63,28 @@ func (initrd *file) Name() string {
...
+	case FsTypeErofs:
+		return initrd.opts.output, erofs.CreateFS(ctx, initrd.opts.output, initrd.path,
+			erofs.WithAllRoot(!initrd.opts.keepOwners),
+		)
+	case FsTypeCpio:
+		err := cpio.CreateFS(ctx, initrd.opts.output, initrd.path,
+			cpio.WithAllRoot(!initrd.opts.keepOwners),
+		)
+		if err != nil {
+			return "", fmt.Errorf("could not create CPIO archive: %w", err)
```
nurof3n: ditto

initrd/ociimage.go
```diff
@@ -171,171 +167,32 @@ func (initrd *ociimage) Build(ctx context.Context) (string, error) {
...
+		return initrd.opts.output, erofs.CreateFS(ctx, initrd.opts.output, ociTarballFile,
+			erofs.WithAllRoot(!initrd.opts.keepOwners),
+		)
+	case FsTypeCpio:
+		err := cpio.CreateFS(ctx, initrd.opts.output, ociTarballFile,
+			cpio.WithAllRoot(!initrd.opts.keepOwners),
+		)
 		if err != nil {
-			return err
+			return "", fmt.Errorf("could not create CPIO archive: %w", err)
```
nurof3n, line 187: ditto

initrd/options.go
```diff
@@ -66,13 +72,27 @@ func WithCacheDir(dir string) InitrdOption {
...
+// owners of the files in the initramfs.
+func WithKeepOwners(keep bool) InitrdOption {
 	return func(opts *InitrdOptions) error {
-		opts.arch = arch
+		opts.keepOwners = keep
+		return nil
+	}
+}
+
+// WithOutput sets the location of the resulting CPIO archive file.
```
nurof3n: remove specific `CPIO`

initrd/tarball.go
```diff
@@ -71,227 +67,28 @@ func (initrd *tarball) Build(ctx context.Context) (string, error) {
...
+		return initrd.opts.output, erofs.CreateFS(ctx, initrd.opts.output, initrd.path,
+			erofs.WithAllRoot(!initrd.opts.keepOwners),
+		)
+	case FsTypeCpio:
+		err := cpio.CreateFS(ctx, initrd.opts.output, initrd.path,
+			cpio.WithAllRoot(!initrd.opts.keepOwners),
+		)
 		if err != nil {
-			return "", fmt.Errorf("could not seek to start of tarball: %w", err)
+			return "", fmt.Errorf("could not create CPIO archive: %w", err)
```
nurof3n, line 82: ditto

initrd/utils.go
```diff
@@ -9,24 +9,17 @@ import (
...
+	reader, err := os.Open(path)
 	if err != nil {
-		return fmt.Errorf("could not seek to start of file: %w", err)
+		return fmt.Errorf("could not open initramfs file: %w", err)
 	}
 
 	fw, err := os.OpenFile(output+".gz", os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o644)
 	if err != nil {
-		return fmt.Errorf("could not open initramfs file: %w", err)
+		return fmt.Errorf("could not gzip output file: %w", err)
```
nurof3n: `could not open gzip output file`

internal/cli/kraft/cloud/deploy/deployer_kraftfile_unikraft.go
```diff
@@ -62,6 +62,7 @@ func (deployer *deployerKraftfileUnikraft) Deploy(ctx context.Context, opts *Dep
 		NoUpdate:     opts.NoUpdate,
 		Platform:     "kraftcloud",
 		Rootfs:       opts.Rootfs,
+		RootfsType:   opts.RootfsType,
```
nurof3n: add `KeepFileOwners` option

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
```diff
@@ -382,8 +382,19 @@ func (p *packagerKraftfileRuntime) Pack(ctx context.Context, opts *PkgOptions, a
 		log.G(ctx).Warn("no kernel detected: packaging without - this may produce unexpected results")
 	}
 
+	// Create a temporary directory we can use to store the artifacts from
+	// pulling and extracting the identified package.
+	tempDir, err := os.MkdirTemp("", "kraft-pkg-")
```
nurof3n, line 387: this does not seem to be used - investigate places wher tmp directory is created for no reason

internal/cli/kraft/run/run.go
```diff
@@ -30,30 +31,31 @@ import (
...
+	Prefix        string        `long:"prefix" usage:"Prefix each log line with the given string"`
+	PrefixName    bool          `long:"prefix-name" usage:"Prefix each log line with the machine name"`
+	Remove        bool          `long:"rm" usage:"Automatically remove the unikernel when it shutsdown"`
+	Rootfs        string        `long:"rootfs" usage:"Specify a path to use as root file system (can be volume or initramfs)"`
+	RootfsType    initrd.FsType `noattribute:"true"`
+	RunAs         string        `long:"as" usage:"Force a specific runner"`
+	Runtime       string        `long:"runtime" short:"r" usage:"Set an alternative unikernel runtime"`
+	Target        string        `long:"target" short:"t" usage:"Explicitly use the defined project target"`
+	Volumes       []string      `long:"volume" short:"v" usage:"Bind a volume to the instance"`
+	WithKernelDbg bool          `long:"symbolic" usage:"Use the debuggable (symbolic) unikernel"`
```
nurof3n: add `KeepFileOwners` option

internal/cli/kraft/run/utils.go
```diff
@@ -374,6 +374,7 @@ func (opts *RunOptions) prepareRootfs(ctx context.Context, machine *machineapi.M
 		)),
 		initrd.WithArchitecture(machine.Spec.Architecture),
 		initrd.WithWorkdir(opts.workdir),
+		initrd.WithType(initrd.FsTypeCpio),
```
nurof3n: use `opts.RootfsType`

internal/cli/kraft/cloud/deploy/deployer_kraftfile_unikraft.go
```diff
@@ -62,6 +62,7 @@ func (deployer *deployerKraftfileUnikraft) Deploy(ctx context.Context, opts *Dep
...
 		SaveBuildLog: opts.SaveBuildLog,
```
nurof3n: consider failing early in here because of no erofs support

tools/github-action/build.go
```diff
@@ -39,6 +39,7 @@ func (opts *GithubAction) build(ctx context.Context) error {
 				"rootfs-cache",
 			)),
 			initrd.WithArchitecture(opts.target.Architecture().String()),
+			initrd.WithType(initrd.FsTypeCpio),
```
nurof3n: use `opts.RootfsType`? idk

erofs/writer.go
```diff
@@ -0,0 +1,696 @@
...
+
+	sb := SuperBlock{
+		Magic:         SuperBlockMagicV1,
+		BlockSizeBits: BlockSizeBits,
+		Inodes:        uint64(len(w.inodes)),
+		Blocks:        uint32(1 + (metaSize+dataSize)/BlockSize),
+		MetaBlockAddr: uint32(metaBlockAddr),
+		UUID:          uuid,
+		BuildTime:     uint64(time.Now().Unix()),
+		BuildTimeNsec: uint32(time.Now().Nanosecond()),
```
nurof3n: call `time.Now()` only once

erofs/owner_windows.go
craciunoiuc: These are imported files I would leave as is for now we can refactor later

erofs/owner_unix.go
craciunoiuc: These are imported files I would leave as is for now we can refactor later

internal/cli/kraft/cloud/deploy/deployer_kraftfile_unikraft.go
craciunoiuc: Will add check to `Deployable` function to avoid using Unikraft if erofs

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
craciunoiuc: This was part of all other deployers so I added it here also

I say we keep in mind and remove in a subsequent PR

internal/cli/kraft/run/utils.go
craciunoiuc: Makes sense

(note: for now it will/should always be cpio)

tools/github-action/build.go
craciunoiuc: Should be fine, though there is no configured way to set it, I need to add that also

cpio/create.go
craciunoiuc: Actually this looks insane but I need to reopen the reader because the tar/targz readers read the first X bytes of the stream, meaning I don't know where to properly seek back

erofs/create.go
craciunoiuc: duplication is fine we do it all the time 🫠

initrd/directory.go
craciunoiuc: Not redundant as we can also do the compress step, which is missing from erofs

initrd/file.go
craciunoiuc: Here it makes sense

nurof3n (approved):
LGTM!

Reviewed-by: Alex-Andrei Cioc <andrei.cioc@unikraft.io>

nderjung (approved):
Thanks everyone! :) 

PR #2026, opened 2024-12-19
fix(github-action): Do not build kernel if it does not exist

craciunoiuc:
turns out the action used an outdated packaging system such that packaging anything non-unikraft-based (basically anything that uses `runtime:` ) would fail because of no kernel image

### Prerequisite checklist

### DIFF
--- a/tools/github-action/pack.go
+++ b/tools/github-action/pack.go
@@ -6,26 +6,182 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/mattn/go-shellwords"
+	"kraftkit.sh/config"
+	"kraftkit.sh/initrd"
+	"kraftkit.sh/log"
 	"kraftkit.sh/pack"
 	"kraftkit.sh/packmanager"
+	"kraftkit.sh/tui/selection"
 	"kraftkit.sh/unikraft"
+	"kraftkit.sh/unikraft/app"
+	"kraftkit.sh/unikraft/target"
 )
 
-// pack
-func (opts *GithubAction) packAndPush(ctx context.Context) error {
-	output := opts.Output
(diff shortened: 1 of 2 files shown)


tools/github-action/pack.go
```diff
@@ -6,12 +6,14 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
-	"github.com/mattn/go-shellwords"
+	"kraftkit.sh/config"
+	"kraftkit.sh/internal/cli/kraft/pkg"
```
nderjung: Does importing `kraftkit.sh/internal` work?

tools/github-action/pack.go
```diff
@@ -26,54 +28,30 @@ func (opts *GithubAction) packAndPush(ctx context.Context) error {
...
-		if len(opts.project.Command()) > 0 {
-			opts.Args = opts.project.Command()
-		} else if len(opts.target.Command()) > 0 {
-			opts.Args = opts.target.Command()
-		}
-	}
-
-	cmdShellArgs, err := shellwords.Parse(strings.Join(opts.Args, " "))
+	config.G[config.KraftKit](ctx).Log.Type = "basic"
+	config.G[config.KraftKit](ctx).Log.Level = "info"
```
nderjung: We should not manipulate the log level as this is user programmable.

The log-type should automatically downgrade to `basic` (from `fancy`) in a non-TTY environment. If it is not downgrading, then there is an issue elsewhere.

craciunoiuc: yeah it did not

The environment is a tty though unfortunately afaik :(

tools/github-action/pack.go
craciunoiuc: it does if you use `replace` which I did, will fix

nderjung (approved):
Thanks!

PR #2038, opened 2025-01-08
feat: Introduce tarball manifest provider

nderjung:
### Prerequisite checklist

### Description of changes

This PR series introduces a new "manifest provider" which handles accessing source code, in this case from tarballs.  It makes accessing sources in the following format of a `Kraftfile` possible:

```yaml
spec: v0.6

name: http-c

unikraft:
  source: ./unikraft.tar.gz

libraries:
  lwip:
    source: ./lib-lwip.tar.gz

targets:
- qemu/x86_64
```

### DIFF
--- a/archive/unarchive.go
+++ b/archive/unarchive.go
@@ -6,6 +6,7 @@ package archive
 
 import (
 	"archive/tar"
+	"bytes"
 	"compress/gzip"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


unikraft/component/component.go
```diff
@@ -115,7 +115,7 @@ func urlHasVersion(u *url.URL) string {
 func parseStringProp(entry string) map[string]interface{} {
 	component := make(map[string]interface{})
 
-	if f, err := os.Stat(entry); err == nil && f.IsDir() {
+	if _, err := os.Stat(entry); err == nil {
```
craciunoiuc: I'm fine with this, but at the same time if we want to keep the safety checks we can do

```
f.Mode().IsRegular() || f.Mode().IsDir()
```

Meaning we do not accept sockets/devices/other linux stuff

archive/unarchive.go
```diff
@@ -41,6 +41,45 @@ func UntarGz(src, dst string, opts ...UnarchiveOption) error {
 	return Untar(gzipReader, dst, opts...)
 }
 
+// IsTarGz checks if a file is a valid tarball which has been gzip compressed.
+func IsTarGz(filepath string) (bool, error) {
```
craciunoiuc, line 46: I find this function a bit too "low-level" for its context

How does this compare to this one https://github.com/unikraft/kraftkit/pull/2007/files#diff-df47a0f9231e2e69bf586feadaee4d86b31e7205b892d1bd19535de65cf9cb81R121

(it was generated by copilot if I remember correctly, so it must be right :D )

archive/unarchive.go
```diff
@@ -91,6 +92,49 @@ func Untar(src io.Reader, dst string, opts ...UnarchiveOption) error {
...
+			if len(topLevelDirs) > 1 {
+				break
+			}
+		}
+
+		if len(topLevelDirs) == 1 {
+			uc.stripComponents = 1
+		}
+	}
+
```
craciunoiuc, line 137: Question: isn't it faster to just unpack as is and then `mv` everything one level down? The operation itself does not extra copying compared to here (I can see at least 1 copy)

craciunoiuc (changes requested):
Left a few conceptual comments, overall looks fine

archive/unarchive.go
nderjung: The `initrd` one can be removed and simply reference this one, since this is now in the `archive` pkg

archive/unarchive.go
nderjung: It's at least more memory efficient to do it this way since there's no intermediate buffer, but i'm not sure whether it's necessarily faster since you anyway have to iterate over at least the first-level child entries and move those to the new location.

craciunoiuc (approved):
All good here. Thanks!

PR #2046, opened 2025-01-10
refactor!: Various improvements and fixes to the manifest PM

nderjung:
### Prerequisite checklist

### Description of changes

This PR series contains various improvements and bug fixes to the Manifest package manager, overall making it particularly more friendly to use as a Go SDK.

### DIFF
--- a/manifest/directory.go
+++ b/manifest/directory.go
@@ -76,7 +76,7 @@ func (dp DirectoryProvider) Manifests() ([]*Manifest, error) {
 		Origin:   dp.dir,
 		Channels: []ManifestChannel{
 			{
-				Name:     "default",
+				Name:     dp.mopts.defaultChannelName,
 				Default:  true,
 				Resource: dp.dir,
 			},
--- a/manifest/directory.go
+++ b/manifest/directory.go
@@ -86,7 +86,7 @@ func (dp DirectoryProvider) Manifests() ([]*Manifest, error) {
 	return []*Manifest{manifest}, nil
 }
 
-func (dp DirectoryProvider) PullManifest(ctx context.Context, manifest *Manifest, opts ...pack.PullOption) error {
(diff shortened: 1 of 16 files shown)


manifest/directory.go
```diff
@@ -96,9 +96,45 @@ func (dp DirectoryProvider) PullManifest(ctx context.Context, manifest *Manifest
...
+	local, err := unikraft.PlaceComponent(
+		popts.Workdir(),
+		manifest.Type,
+		manifest.Name,
+	)
+	if err != nil {
+		return fmt.Errorf("could not place component package: %s", err)
+	}
+
+	f, err := os.Lstat(local)
```
craciunoiuc: Is the use of `Lstat` intentional? It only has different results on Windows:

Extracted from gostd on Linux:
```
// Stat returns a [FileInfo] describing the named file.
// If there is an error, it will be of type [*PathError].
func Stat(name string) (FileInfo, error) {
	testlog.Stat(name)
	return statNolog(name)
}

// Lstat returns a [FileInfo] describing the named file.
// If the file is a symbolic link, the returned FileInfo
// describes the symbolic link. Lstat makes no attempt to follow the link.
// If there is an error, it will be of type [*PathError].
//
// On Windows, if the file is a reparse point that is a surrogate for another
// named entity (such as a symbolic link or mounted folder), the returned
// FileInfo describes the reparse point, and makes no attempt to resolve it.
func Lstat(name string) (FileInfo, error) {
	testlog.Stat(name)
	return lstatNolog(name)
}
```

craciunoiuc (approved):
All good here. Thanks!

manifest/directory.go
nderjung: It's not intentional, it was there before but may appear in a diff because I moved code around. I can change to `Stat` instead.

PR #2070, opened 2025-01-14
fix(internal): Load env vars from file when specified

craciunoiuc:
By default it uses '.env' if found. If '--env-file' is specified then use that instead of it.

### Prerequisite checklist

### DIFF
--- a/internal/cli/kraft/cloud/compose/build/build.go
+++ b/internal/cli/kraft/cloud/compose/build/build.go
@@ -104,10 +104,16 @@ func Build(ctx context.Context, opts *BuildOptions, args ...string) error {
 			return err
 		}
 
+		var envFiles []string
+		if opts.EnvFile != "" {
+			envFiles = append(envFiles, opts.EnvFile)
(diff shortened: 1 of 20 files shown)


internal/cli/kraft/cloud/compose/build/build.go
```diff
@@ -337,6 +343,14 @@ func (opts *BuildOptions) Pre(cmd *cobra.Command, args []string) error {
...
 	}
 
+	if cmd.Flag("env-file").Changed {
+		opts.EnvFile = cmd.Flag("env-file").Value.String()
+	}
+
+	if cmd.Flag("env-file").Changed {
+		opts.EnvFile = cmd.Flag("env-file").Value.String()
+	}
+
```
nderjung: Duplicate code? Also why use `.Changed`? This runs through `cmdfactory`'s struct tag parser should load in the string value anyway?

nderjung (approved):
Gracias senior.

PR #2092, opened 2025-01-23
refactor(run): Better plat/arch selection

nderjung:
### Prerequisite checklist

### Description of changes

This PR refactors and generally improves the selection process for targets and/or packages during the invocation of `kraft run`. This begins with breaking up `discoverMachineController` into two seperate methods for detecting the platform and the architecture. Broadly, this, along with migrating the call to these methods to later stages, allows for the various runners to make a wider search before narrowing into the host's plat/arch combo.  This is done in order to 1). provide the user the option to select from the various plat/arch combinations if the host does not match and 2). in the case where the user is working with multiple builds in their project, to always all them to select these when no flags are present.  Previously, the embedded functionality would always just default to the host's plat/arch but this might not always be appropriate given what the user is trying to do.

### DIFF
--- a/internal/cli/kraft/run/run.go
+++ b/internal/cli/kraft/run/run.go
@@ -140,6 +140,7 @@ func NewCmd() *cobra.Command {
 func (opts *RunOptions) Pre(cmd *cobra.Command, _ []string) error {
 	ctx := cmd.Context()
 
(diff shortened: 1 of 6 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #2150, opened 2025-02-12
fix(cache): Add read lock around map

nderjung:
### Prerequisite checklist

### DIFF
--- a/oci/cache/image.go
+++ b/oci/cache/image.go
@@ -30,9 +30,12 @@ func RemoteImage(ref name.Reference, options ...remote.Option) (v1.Image, error)
 		goto lookup
 	}
 
+	imageCacheMu.Lock()
 	if image, ok := imageCache[name]; ok {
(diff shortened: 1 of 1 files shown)


oci/cache/image.go
```diff
@@ -30,9 +30,11 @@ func RemoteImage(ref name.Reference, options ...remote.Option) (v1.Image, error)
 		goto lookup
 	}
 
+	imageCacheMu.Lock()
 	if image, ok := imageCache[name]; ok {
 		return image, nil
 	}
+	imageCacheMu.Unlock()
```
craciunoiuc, line 38: This is wrong right? What if the `if` returns and the mutex is never unlocked it would make everything go boom

Either unlock in the return also, or use a `sync.Map`

craciunoiuc (changes requested):

craciunoiuc (approved):
All good here. Thanks!

PR #2174, opened 2025-02-18
feat: Add sentry and anonymous telemetry support

nderjung:
### Prerequisite checklist

### Description of changes

In order to support the increasing usage of the tool and better facilitate fixes, introduce Sentry back trace handler into `kraft`.  This is an anonymous usage and opt-in feature which will help the KraftKit maintainers catch panics and provide fixes.

The default value of the new configuration option which enables this functionality is **false/"off"** and can be switched via:
    
- env var: `KRAFTKIT_COLLECT_ANONYMOUS_TELEMETRY`
- config file: `collect_anonymous_telemetry`
- CLI flag: `--collect-anonymous-telemetry`
    
We're introducing the option in order to improve the CLI experience for endusers and tackle panics and handle bug reports in a more systematic and prompt manner.
    
The local build experience totally disables telemetry and it's only included as part of official releases.  This is done because 1). contributors are making iterative updates may encounter higher error-rates during developmenet and 2). local-first development may include changes which are not "official" and therefore not debuggable from a telemetry point-of-view.

### DIFF
--- a/internal/cli/kraft/system/set/set.go
+++ b/internal/cli/kraft/system/set/set.go
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package set
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/spf13/cobra"
+
+	"kraftkit.sh/cmdfactory"
+	"kraftkit.sh/config"
+	"kraftkit.sh/log"
+)
+
+type Set struct{}
+
+func NewCmd() *cobra.Command {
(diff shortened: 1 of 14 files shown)


internal/cli/kraft/system/set/set.go
```diff
@@ -0,0 +1,49 @@
...
+	"kraftkit.sh/config"
+)
+
+type Set struct{}
+
+func NewCmd() *cobra.Command {
+	cmd, err := cmdfactory.New(&Set{}, cobra.Command{
+		Short: "Set a KraftKit configuration option",
+		Use:   "set KEY=VALUE",
+		Args:  cobra.ExactArgs(1),
```
craciunoiuc: Could this be updated maybe to support multiple options at a time?

I'm thinking of someone doing an initial setup (👀 ansible) and they want to configure kraftkit with some defaults

craciunoiuc (changes requested):

internal/cli/kraft/system/set/set.go
nderjung: Done

craciunoiuc (approved):
All good here. Thanks!

PR #2321, opened 2025-05-26
feat(oci): Add util for accessing OCI layer contents

robertgzr:
This allows for working with layers as raw tarball streams, without
first unpacking to disk and before storing them in containerd.

Signed-off-by: Robert Günzler <robert@unikraft.cloud>

### Prerequisite checklist

### Description of changes

This PR adds a small utility function to access Unikraft image contents from a
tar stream. Before this was only possible if the image was already unpacked on
disk or in containerd's image store.

### DIFF
--- a/oci/utils/tarball.go
+++ b/oci/utils/tarball.go
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file expect in compliance with the License.
+package utils
+
+import (
+	"archive/tar"
+	"fmt"
+	"io"
+	"io/fs"
+	"path/filepath"
+)
+
+// OpenFromTarOCILayer returns a reader for the given file from the provided
+// tarball reader.
+func OpenFromTarOCILayer(r io.Reader, path string) (io.Reader, error) {
+	tarPath, err := filepath.Rel("/", path)
+	if err != nil {
+		return nil, fmt.Errorf("could not trim leading separator from path %q: %w", path, err)
+	}
+
+	tr := tar.NewReader(r)
+	for {
+		h, err := tr.Next()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, fmt.Errorf("could not advance to the next entry in the OCI image layer: %w", err)
+		}
+
+		if h.Name != tarPath {
+			continue
+		}
+
+		if t := h.Typeflag; t != tar.TypeReg {
+			return nil, fmt.Errorf("path is not a regular file")
+		}
+
+		return tr, nil
+	}
+
+	return nil, &fs.PathError{Op: "open", Path: path, Err: fs.ErrNotExist}
+}


oci/utils/tarball.go
```diff
@@ -0,0 +1,41 @@
+package utils
```
craciunoiuc, line 5:
```suggestion
// SPDX-License-Identifier: BSD-3-Clause
// Copyright (c) 2025, Unikraft GmbH and The KraftKit Authors.
// Licensed under the BSD-3-Clause License (the "License").
// You may not use this file expect in compliance with the License.
package utils
```

oci/utils/tarball.go
```diff
@@ -0,0 +1,41 @@
...
+	"io/fs"
+	"path/filepath"
+)
+
+// OpenFromTarOCILayer returns a reader for the given file from the provided
+// tarball reader.
+func OpenFromTarOCILayer(r io.Reader, path string) (io.Reader, error) {
+	tarPath, err := filepath.Rel("/", path)
+	if err != nil {
+		return nil, fmt.Errorf("trimming leading separator from path %q: %w", path, err)
```
craciunoiuc: We usually format these as `could not ...`, so let's just do the same:

```suggestion
		return nil, fmt.Errorf("could not trim leading separator from path %q: %w", path, err)
```

oci/utils/tarball.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	tr := tar.NewReader(r)
+	for {
+		h, err := tr.Next()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, fmt.Errorf("advancing to the next entry in the OCI image layer: %w", err)
```
craciunoiuc:
```suggestion
			return nil, fmt.Errorf("could not advance to the next entry in the OCI image layer: %w", err)
```

oci/utils/tarball.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+	"io"
+	"io/fs"
+	"path/filepath"
+)
+
+// OpenFromTarOCILayer returns a reader for the given file from the provided
+// tarball reader.
+func OpenFromTarOCILayer(r io.Reader, path string) (io.Reader, error) {
+	tarPath, err := filepath.Rel("/", path)
```
craciunoiuc, line 18: I would replace "/" with `filepath.Separator` or similar to keep at least a basic level of windows compatibility 😅

oci/utils/tarball.go
```diff
@@ -0,0 +1,41 @@
...
+			}
+			return nil, fmt.Errorf("advancing to the next entry in the OCI image layer: %w", err)
+		}
+
+		if h.Name != tarPath {
+			continue
+		}
+
+		if t := h.Typeflag; t != tar.TypeReg {
+			return nil, fmt.Errorf("file has wrong type %d", t)
```
craciunoiuc: I think the file code/number might actually be more confusing than helpful, but it's up to you this one

craciunoiuc (changes requested):
A set of comments from my side, only the header one is mandatory

oci/utils/tarball.go
robertgzr: i will change it to:

```suggestion
			return nil, fmt.Errorf("path is not a regular file")

```

oci/utils/tarball.go
robertgzr: not sure if that makes sense? the purpose of this is to turn the passed path into one relative to `/` because that is how the tarheader name value encodes files nested in subdirectories.

on windows the root path can be at least two things if i have that right: `\\(some host)` or `(some letter):\\`, so using `filepath.Seperator` which would be `\` doesn't do the right thing.

since we are reading archives created by unix tar the paths returned by the reader would always contain unix seperators, i think. so it might make more sense to actually ensure that is the case

craciunoiuc: Okay makes sense, I thought this was the path to the archive actually

All good then

craciunoiuc (approved):
All good here. Thanks!

PR #2340, opened 2025-06-16
feat(initrd): Mount project-scope directory for buildkit caching

craciunoiuc:
### Prerequisite checklist


jedevc:
What's the intention behind this? It means we won't have a global cache for buildkit (each project will have its own).

It would allow builds in parallel, but ideally integrating with docker + build + some other stuff should make this a lot simpler without this change.

craciunoiuc:
This came previously because there was a bug which made ephemeral containers not cache their artifacts which meant pulling every time

At that time I said: why not make them project local and fix the thing at the same time

Meanwhile it got fixed, and with the additional changes you are making I feel like this is going out of scope

craciunoiuc:
Closing this as it is no longer required

PR #2396, opened 2025-08-01
feat: Auxiliary ROM packaging

nderjung:
### Prerequisite checklist

### Description of changes

This PR introduces initial support for packaging auxiliary read-only memory blob files (auxiliary ROMs).  These are totally arbitrary files which can be used by an external provider and mounted to the kernel.  It's up to the application to decide what to do with this and provides flexibility in terms of being able to support additional types of usecases beyond a single initramfs.

Right now this can be used in two different ways:

1. Via a `Kraftfile`:
   ```yaml
   spec: v0.6

   runtime: base:latest

   rootfs: ./rootfs

   roms:
   - path/to/blob.bin
   ```

2. Via CLI:
   ```
   kraft pkg \
   --name test:latest \
   --rom path/to/blob.bin \
   [...]
   ```

### DIFF
--- a/internal/cli/kraft/pkg/packager_kraftfile_runtime.go
+++ b/internal/cli/kraft/pkg/packager_kraftfile_runtime.go
@@ -43,10 +43,6 @@ func (p *packagerKraftfileRuntime) Packagable(ctx context.Context, opts *PkgOpti
 		return false, fmt.Errorf("cannot package without unikraft core specification")
 	}
 
-	if opts.Project.Targets() == nil {
-		return false, fmt.Errorf("cannot package without project targets")
-	}
-
 	if opts.Project.Rootfs() != "" && opts.Rootfs == "" {
 		opts.Rootfs = opts.Project.Rootfs()
 	}
@@ -59,10 +55,6 @@ func (p *packagerKraftfileRuntime) Pack(ctx context.Context, opts *PkgOptions, a
 	var err error
 	var targ target.Target
 	var runtimeName, runtimeVersion string
-
-	if opts.Project.Targets() == nil {
-		return nil, fmt.Errorf("cannot package without project targets")
-	}
 
 	if len(opts.Runtime) > 0 {
 		var ok bool
--- a/internal/cli/kraft/pkg/packager_kraftfile_runtime.go
+++ b/internal/cli/kraft/pkg/packager_kraftfile_runtime.go
@@ -43,8 +43,8 @@ func (p *packagerKraftfileRuntime) Packagable(ctx context.Context, opts *PkgOpti
 		}
 	}
 
-	if opts.Project.Runtime() == nil && len(opts.Project.Rootfs()) == 0 {
-		return false, fmt.Errorf("cannot package without any of runtime or rootfs")
+	if opts.Project.Runtime() == nil && len(opts.Project.Rootfs()) == 0 && len(opts.Project.Roms()) == 0 {
+		return false, fmt.Errorf("cannot package without any of runtime, rootfs or roms")
 	}
 
 	if opts.Project.Rootfs() != "" && opts.Rootfs == "" {
@@ -65,6 +65,7 @@ func (p *packagerKraftfileRuntime) Pack(ctx context.Context, opts *PkgOptions, a
 		packKConfig      kconfig.KeyValueMap
 		packCmds         []string
 		packEnv          []string
+		packRoms         []string
 		packArgs         []string
 		packRootfs       initrd.Initrd
 		packArchitecture arch.Architecture
@@ -411,6 +412,14 @@ func (p *packagerKraftfileRuntime) Pack(ctx context.Context, opts *PkgOptions, a
 		}
 	}
 
+	if opts.Project != nil {
+		packRoms = opts.Project.Roms()
+	} else if len(opts.Roms) > 0 {
+		packRoms = opts.Roms
+	} else if targ != nil && len(targ.Roms()) > 0 {
+		packRoms = targ.Roms()
+	}
+
 	var labels map[string]string
 	if opts.Project != nil {
 		labels = opts.Project.Labels()
@@ -458,6 +467,12 @@ func (p *packagerKraftfileRuntime) Pack(ctx context.Context, opts *PkgOptions, a
 				if packRootfs != nil {
 					popts = append(popts,
 						packmanager.PackInitrd(packRootfs),
+					)
+				}
+
+				if len(packRoms) > 0 {
+					popts = append(popts,
+						packmanager.PackRoms(packRoms...),
 					)
 				}
 
--- a/internal/cli/kraft/pkg/packager_kraftfile_runtime.go
+++ b/internal/cli/kraft/pkg/packager_kraftfile_runtime.go
@@ -28,7 +28,21 @@ import (
 	"kraftkit.sh/unikraft/target"
 )
 
-type packagerKraftfileRuntime struct{}
+type packagerKraftfileRuntime struct {
+	name    string
+	version string
+	target  target.Target
+
+	// Packaging options
+	kernel       string
+	kconfig      kconfig.KeyValueMap
+	args         []string
+	env          []string
+	roms         []string
+	rootfs       initrd.Initrd
+	architecture arch.Architecture
+	platform     plat.Platform
+}
 
 // String implements fmt.Stringer.
 func (p *packagerKraftfileRuntime) String() string {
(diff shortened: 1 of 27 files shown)


internal/cli/kraft/pkg/packager_kraftfile_runtime.go
```diff
@@ -60,10 +56,6 @@ func (p *packagerKraftfileRuntime) Pack(ctx context.Context, opts *PkgOptions, a
 	var targ target.Target
 	var runtimeName, runtimeVersion string
 
-	if opts.Project.Targets() == nil {
-		return nil, fmt.Errorf("cannot package without project targets")
-	}
-
```
craciunoiuc: I wouldn't remove checks like these, we've had panics before from cases like this

Better safe than sorry 💫

internal/cli/kraft/utils/rootfs.go
```diff
@@ -19,9 +19,9 @@ import (
 
 // BuildRootfs generates a rootfs based on the provided working directory and
 // the rootfs entrypoint for the provided target(s).
-func BuildRootfs(ctx context.Context, workdir, rootfs string, compress bool, arch string) (string, []string, []string, error) {
+func BuildRootfs(ctx context.Context, workdir, rootfs string, compress bool, arch string) (initrd.Initrd, []string, []string, error) {
```
craciunoiuc, line 22: I think the commit containing this should have a breaking change marker `!`

internal/cli/kraft/pkg/packager_kraftfile_unikraft.go
```diff
@@ -93,18 +94,16 @@ func (p *packagerKraftfileUnikraft) Pack(ctx context.Context, opts *PkgOptions,
 	for _, targ := range selected {
 		var cmds []string
 		var envs []string
-		rootfs := opts.Rootfs
+		var rootfs initrd.Initrd
 
 		// Reset the rootfs, such that it is not packaged as an initrd if it is
 		// already embedded inside of the kernel.
-		if opts.Project.KConfig().AnyYes(
+		if !opts.Project.KConfig().AnyYes(
```
craciunoiuc: Why is this Negated? It Becomes `AllNo`

oci/manager.go
```diff
@@ -226,17 +226,19 @@ func (manager *OCIManager) update(ctx context.Context, auths map[string]config.A
...
 	ctx, handle, err := manager.handle(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	pkg, err := NewPackageFromTarget(ctx, handle, targ, opts...)
+	var pkg pack.Package
+
+	targ, ok := entity.(target.Target)
+	if ok {
```
craciunoiuc: Can't this be moved inside the if statement 😳

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
```diff
@@ -52,21 +56,37 @@ func (p *packagerKraftfileRuntime) Packagable(ctx context.Context, opts *PkgOpti
...
+		runtimeVersion   string
+		targ             target.Target
+		packKernel       string
+		packKConfig      kconfig.KeyValueMap
+		packCmds         []string
+		packEnv          []string
+		packArgs         []string
+		packRootfs       initrd.Initrd
+		packArchitecture arch.Architecture
+		packPlatform     plat.Platform
```
craciunoiuc: Would these sit better in a locally declared structure?

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
```diff
@@ -261,102 +284,126 @@ func (p *packagerKraftfileRuntime) Pack(ctx context.Context, opts *PkgOptions, a
...
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("could not create temporary directory: %w", err)
 		}
 
-		if err := paramodel.Start(); err != nil {
-			return nil, err
+		defer func() {
+			os.RemoveAll(tempDir)
+		}()
```
craciunoiuc: I know this was here before but why no `defer os.RemoveAll(tempDir)`, you usually wrap in a function if you want to check return codes

internal/cli/kraft/pkg/pkg.go
```diff
@@ -47,6 +47,7 @@ type PkgOptions struct {
 	Labels       []string                  `local:"true" long:"label" short:"l" usage:"Set labels to be packed into the package (k=v)"`
 	Name         string                    `local:"true" long:"name" short:"n" usage:"Specify the name of the package"`
 	NoKConfig    bool                      `local:"true" long:"no-kconfig" usage:"Do not include target .config as metadata"`
+	NoKernel     bool                      `local:"true" long:"no-kernel" usage:"Allow packaging with a kernel image"`
```
craciunoiuc: `Without`

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
```diff
@@ -43,8 +43,8 @@ func (p *packagerKraftfileRuntime) Packagable(ctx context.Context, opts *PkgOpti
 		}
 	}
 
-	if opts.Project.Runtime() == nil && len(opts.Project.Rootfs()) == 0 {
-		return false, fmt.Errorf("cannot package without any of runtime or rootfs")
+	if opts.Project.Runtime() == nil && len(opts.Project.Rootfs()) == 0 && len(opts.Project.Roms()) == 0 {
+		return false, fmt.Errorf("cannot package without any of runtime, rootfs or roms")
```
craciunoiuc, line 61: `rootfs, or roms`

internal/cli/kraft/pkg/pkg.go
```diff
@@ -53,6 +53,7 @@ type PkgOptions struct {
 	Platform     string                    `local:"true" long:"plat" short:"p" usage:"Filter the creation of the package by platform of known targets (fc/qemu/xen/kraftcloud)"`
 	Project      app.Application           `noattribute:"true"`
 	Push         bool                      `local:"true" long:"push" short:"P" usage:"Push the package on if successfully packaged"`
+	Roms         []string                  `local:"true" long:"rom" short:"R" usage:"Specify a path to an auxiliary ROM to include in the package."`
```
craciunoiuc: remove period at the end to align with the others

schema/schema.go
```diff
@@ -174,10 +183,8 @@ func specificity(err gojsonschema.ResultError) int {
 	return len(strings.Split(err.Field(), "."))
 }
 
-type SchemaVersion string
-
-const (
-	SchemaVersionV0_5   = SchemaVersion("0.5")
-	SchemaVersionV0_6   = SchemaVersion("0.6")
+var (
```
craciunoiuc, line 185: Why are these not const anymore?

schema/schema.go
```diff
@@ -59,13 +61,20 @@ func Validate(ctx context.Context, config map[string]interface{}) error {
...
+	if specVer.LessThan(SchemaVersionLatest) {
 		log.G(ctx).Warnf("specification in Kraftfile (v%s) version is not latest (v%s)", spec, SchemaVersionLatest)
+	} else if specVer.GreaterThan(SchemaVersionLatest) {
+		log.G(ctx).Warnf("specification version in Kraftfile (v%s) is greater than parsable (v%s)", spec, SchemaVersionLatest)
+		schemaLoader = gojsonschema.NewStringLoader(SchemaV_06)
+	}
+
+	if specVer.Equal(SchemaVersionV0_5) {
+		schemaLoader = gojsonschema.NewStringLoader(SchemaV_05)
+	} else {
```
craciunoiuc, line 73: I would actually love the Warning in the else branch and the current content turn into an `else if`

Right now the schema is loaded twice if a version is bigger than V06

Just switch to a bit switch case that does
* Warn if 0.4 or lower
* Warn + Load if 0.5
* Load if 0.6
* Different Warn + Load if 0.7+
* Error if malformed

craciunoiuc (changes requested):
Some comments from my side 😊

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
nderjung: The file in which this is removed has had all instances of `opts.Project.Targets()` checked -- so if we get panics, later commits are incorrect.

internal/cli/kraft/utils/rootfs.go
nderjung: I can do - but note that also it's within the `internal/` package. So we shouldn't propagate this to the PR title, as it's not exposed to users.

internal/cli/kraft/pkg/packager_kraftfile_unikraft.go
nderjung: Whoops. Artifact of previous work.

oci/manager.go
nderjung: what bit?

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
nderjung: maybe but fine for now

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
nderjung: just an artifact of moving code around. i can adjust

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
nderjung: oxford comma is not appropriate here ;-)

schema/schema.go
nderjung: since it's now a pointer to an object

schema/schema.go
nderjung: I think i just need to remove the duplicate load -- we always load v0.6 unless v0.5 is requested? this way we do get more or less what you want

oci/manager.go
craciunoiuc:
```
targ, ok := entity.(target.Target)
	if ok {
```
becomes
```
	if targ, ok := entity.(target.Target); ok {
```

schema/schema.go
craciunoiuc: sure fine, as long as the duplicate load is removed

craciunoiuc (approved):
All good here. Thanks!

craciunoiuc (approved):
All good here. Thanks!

PR #2448, opened 2025-09-02
fix(unikraft): Deep copy library list to avoid altering

craciunoiuc:
### Prerequisite checklist

### Description of changes

Previously, if a call to this was made the saved list of libraries would then contain all internal libraries, apart from external ones. This results then in broken make builds due to incorrectly including too many libraries that should not be included.

### DIFF
--- a/unikraft/app/application.go
+++ b/unikraft/app/application.go
@@ -236,7 +236,10 @@ func (app *application) Libraries(ctx context.Context) (map[string]*lib.LibraryC
 		return nil, err
 	}
 
-	libs := app.libraries
+	libs := map[string]*lib.LibraryConfig{}
+	for key, lib := range app.libraries {
+		libs[key] = lib
+	}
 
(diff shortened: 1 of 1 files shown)


unikraft/app/application.go
```diff
@@ -236,7 +236,10 @@ func (app *application) Libraries(ctx context.Context) (map[string]*lib.LibraryC
 		return nil, err
 	}
 
-	libs := app.libraries
+	libs := map[string]*lib.LibraryConfig{}
```
nderjung, line 239: We know the map length ahead of time, we can be more efficient with:
```go
libs := make(map[string]*lib.LibraryConfig, len(app.libraries))
```

craciunoiuc: as discussed privately, the hassle to do this is too big for minimal improvements

nderjung (approved):
Thanks!

PR #2570, opened 2025-11-26
feat(tools): Add option to generate uncompressed pages

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/tools/genman/main.go
+++ b/tools/genman/main.go
@@ -26,8 +26,20 @@ import (
 )
 
 func main() {
+	var uncompressed bool
+
+	// Parse flags
+	for i, arg := range os.Args[1:] {
+		if arg == "--uncompressed" {
+			uncompressed = true
+			// Remove the flag from args
+			os.Args = append(os.Args[:i+1], os.Args[i+2:]...)
+			break
+		}
+	}
+
 	if len(os.Args[1:]) < 2 || len(os.Args[1:]) > 3 {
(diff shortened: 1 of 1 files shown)


tools/genman/main.go
```diff
@@ -25,9 +25,21 @@ import (
 	"github.com/spf13/pflag"
 )
 
+var uncompressed bool
+
 func main() {
+	// Parse flags
+	for i, arg := range os.Args[1:] {
```
nderjung: Why global var

craciunoiuc: I might have vibe coded this PR 👀

tools/genman/main.go
```diff
@@ -97,27 +110,44 @@ func GenManTreeFromOpts(cmd *cobra.Command, opts GenManTreeOptions) error {
...
+		}
+		writer = f
+	} else {
+		// Use gzip compression
+		filename = filepath.Join(opts.Path, basename+"."+section+".gz")
+		f, err := os.Create(filename)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
```
jedevc, line 132: Probably not a huge issue for a run-once tool, but we should check the return code for the `Close`s here (https://stackoverflow.com/questions/19056309/not-checking-closes-return-value-how-serious-really).

jedevc (approved):

tools/genman/main.go
craciunoiuc: Yeah, we run this tool very rarely, and usually only in CI, we can leave as is

PR #2575, opened 2025-12-01
feat: use docker builder if available

jedevc:
### Prerequisite checklist

  - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
  - [x] Tested your changes against relevant architectures and platforms;
  - [x] Ran `make fmt` on your commit series before opening this PR;
  - [x] Updated relevant documentation.

### Description of changes

Splits out buildkit connection code to a new package, and additionally tries to connect to the default docker daemon! This avoids spinning up things with testcontainers, which we can aim to completely rip-out long term (by using `buildx`).

This is just a first step, but should be a good start!

PR #2593, opened 2025-12-08
fix(buildkit): regression in buildkit default address

jedevc:
### Prerequisite checklist

### Description of changes

See individual commits for more detail.

This PR fixes the regression introduced in https://github.com/unikraft/kraftkit/pull/2575 - we need to connect to `/run/buildkit/buildkitd.sock` if it's available. This was being *implicitly* done - this adds it explicitly.

The behavior is still slightly different (from before #2575). If `KRAFTKIT_BUILDKIT_HOST` is set, then we don't allow a silent failure. All the other options (e.g. `/run/buildkit/buildkitd.sock`, and the builtin docker buildkit) should fail gracefully, until there are no more options left.

We also fix some logging weirdness that made debugging this a bit harder than it needed to be.

### DIFF
--- a/buildkit/connect.go
+++ b/buildkit/connect.go
@@ -15,6 +15,7 @@ import (
 	"kraftkit.sh/log"
 
 	"github.com/moby/buildkit/client"
+	bkappdefaults "github.com/moby/buildkit/util/appdefaults"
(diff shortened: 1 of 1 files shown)


craciunoiuc (approved):
All good here. Thanks!

PR #2602, opened 2025-12-11
fix: Kraftkit successfully builds when receiving a relative path

### LINKED ISSUE #2554: kraft fails to build a Unikernel if relative path is specified
### Describe the bug

When specifying a directory to build with `kraft build [DIR]` that is not the current working directory, the build process fails when assessing initramfs:
``` 
E  getting initramfs size: stat ./Dockerfile: no such file or directory
```

### Steps to reproduce

1. Clone catalog repository (tested at https://github.com/unikraft/catalog/tree/5a22d73139c02021e863f637383e1dc94d965564)
2. Attempt to build nginx: `kraft build library/nginx/1.25 -p fc`

### Expected behavior

Nginx image should build successfully

### KraftKit version

Build fails when getting initramfs size:
```
 E  getting initramfs size: stat ./Dockerfile: no such file or directory
```

_No response_

_No response_

### Relevant log output

```shell

```

nacl-gb3:
This commit resolves Kraftkit issue 2554 by specifying the working directory path when attempting to get the initramfs size

### Prerequisite checklist

### Description of changes

In internal/cli/kraft/build/build.go, I added the working directory variable to the string passed into os.Stat in an 
attempt to get the initramfs size. This should resolve [issue #2554](https://github.com/unikraft/kraftkit/issues/2554).


craciunoiuc:
Hey, thanks for the contribution 🙏 

Unfortunately the full fix is not as simple :(

The path to the workdir that we want to build should take into account also the Kraftfile and the created artifacts. Alternative to this is if we decide that the current directory (or a separate flag one) should be the one that determines the Kraftfile, which can then be used in a separate directory.

In any case, I think it requires a bit of discussion and thinking

nacl-gb3:
Ok, I somewhat expected that it wouldn't be that easy, but wanted to get the discussion going 

I'm more than open to discussing this further, but I'm also new to the codebase.

Can someone help me understand how and where the build options are set? I couldn't quite figure it out through tracking the references.

craciunoiuc:
In the build case I think you care about `build.go` The structure is automatically populated.

In this case I guess we care about `Kraftfile` and `Workdir` and `Rootfs`. (we can then expand to all commands :( )

PR #2616, opened 2026-01-13
feat: Introduce -k|--kernel flag to save unikernel binary

### LINKED ISSUE #705: Allow setting output path of unikernel in `kraft build`
### Feature request summary

This issue tracks introducing a new flag `-k|--kernel` which is used to save the binary image of a unikernel following a successful build.  Not only does this empower the user to use the binary as they wish without hiding it away within the `.unikraft` build folder, but it is useful immediately in the context of #703 which requires knowledge of the the path of the binary.

Following the introduction of https://github.com/unikraft/unikraft/pull/1169 this should now be possible in conjunction with Unikraft's build system too.  The basename of the kernel should be set to the KConfig variable in the aforementioned PR.

Adjust [`unikraft/target/target.go`](https://github.com/unikraft/kraftkit/blob/staging/unikraft/target/target.go) to have a new interface method introduced `SetKernelPath(string)` which a). updates the internal `kernel` attribute and b). sets the relevant KConfig option mentioned in the aforementioned PR.

Since Unikraft's build system does not support an individual path for a kernel (because it concatenates the output with the kernel file name), this adjustment should simply perform a `mv` of the file to the desired location following a successful build.

### Describe alternatives

_No response_

### Related architectures

None

### Related platforms

None

### Additional context

This issue was adjusted to change the flag from `-o|--output` to `-k|--kernel` as the output could be instead used to place the `.unikraft/build` folder instead (a separate issue will track this).

ijaidev:
## Title
**feat: add -k|--kernel flag to save unikernel binary**

## Description

This PR introduces a new `-k` (or `--kernel`) flag to the `kraft build` command. 

Currently, after a build completes, the resulting unikernel binary is buried deep inside the hidden `.unikraft/build/` directory with an auto-generated name. If a user wants to take that binary and run it elsewhere (e.g., in a different tool or just to share with a colleague), they have to dig for it.

With this change, you can now simply run:
```bash
kraft build --kernel ./my-app.bin
```
and KraftKit will automatically save the final binary to your specified path.

## Changes Details
Logic: I updated the target interface (unikraft/target) to support setting a custom kernel path.
Safety: Added a robust moveFile utility that handles moving the binary across different filesystems (since simple renames fail across devices).
Correctness: Crucially, this also updates the CONFIG_UK_NAME KConfig variable during the build. This ensures that the Unikraft build system knows the binary has a new name, keeping standard build artifacts consistent.

### Related Issues
Closes #705

### DIFF
--- a/internal/cli/kraft/build/build.go
+++ b/internal/cli/kraft/build/build.go
@@ -7,6 +7,7 @@ package build
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"sort"
@@ -23,6 +24,7 @@ import (
 	"kraftkit.sh/internal/fancymap"
 	"kraftkit.sh/iostreams"
 	"kraftkit.sh/tui"
+	"kraftkit.sh/unikraft"
 
 	"kraftkit.sh/log"
 	"kraftkit.sh/packmanager"
@@ -39,6 +41,7 @@ type BuildOptions struct {
 	Env            []string        `long:"env" short:"e" usage:"Set environment variables to be built in the unikernel" split:"false"`
 	ForcePull      bool            `long:"force-pull" usage:"Force pulling packages before building"`
 	Jobs           int             `long:"jobs" short:"j" usage:"Allow N jobs at once"`
+	Kernel         string          `long:"kernel" short:"k" usage:"Set the output path of the built kernel image"`
 	KernelDbg      bool            `long:"dbg" usage:"Build the debuggable (symbolic) kernel image instead of the stripped image"`
 	Kraftfile      string          `long:"kraftfile" short:"K" usage:"Set an alternative path of the Kraftfile"`
 	NoCache        bool            `long:"no-cache" short:"F" usage:"Force a rebuild even if existing intermediate artifacts already exist"`
@@ -113,6 +116,10 @@ func Build(ctx context.Context, opts *BuildOptions, args ...string) error {
 
 	if err := build.Prepare(ctx, opts, args...); err != nil {
 		return fmt.Errorf("could not complete build: %w", err)
+	}
+
+	if opts.Kernel != "" {
+		(*opts.Target).SetKernelPath(opts.Kernel)
 	}
 
 	if _, _, _, err = utils.BuildRootfs(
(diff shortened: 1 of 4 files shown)


craciunoiuc (changes requested):
Can you make sure your commit body message does not exceed 74 characters per line? Thanks!

```
⧗   input: feat: Introduce -k|--kernel flag to save unikernel binary

Introduce a new flag -k|--kernel to the build command. This flag allows users to specify a custom output path for the built unikernel binary, moving it from the internal .unikraft/build directory to the desired location.

This change also:
- Updates the internal Target interface with SetKernelPath(string).
- Synchronizes the kernel basename with the CONFIG_UK_NAME KConfig
  variable within the build context.
- Implements a cross-filesystem move utility to handle moves between
  different devices.

Signed-off-by: Jaidev <himjaidev@gmail.com>
✖   body's lines must not be longer than 74 characters [body-max-line-length]
```

internal/cli/kraft/build/build.go
```diff
@@ -299,3 +311,47 @@ func (opts *BuildOptions) Run(ctx context.Context, args []string) error {
 
 	return nil
 }
+
+// moveFile moves a file from src to dst. It first attempts to use os.Rename,
+// which is efficient but only works within the same filesystem. If that fails,
+// it falls back to copying the file and removing the original.
+func moveFile(src, dst string) error {
```
craciunoiuc, line 332: Personal preference: If you comment a function above the signature like you did, there's no reason the add comments inline also

internal/cli/kraft/build/build.go
```diff
@@ -299,3 +311,47 @@ func (opts *BuildOptions) Run(ctx context.Context, args []string) error {
...
+	// Ensure the destination directory exists.
+	if err := os.MkdirAll(filepath.Dir(dst), 0o755); err != nil {
+		return fmt.Errorf("creating destination directory: %w", err)
+	}
+
+	// Try rename first (works within same filesystem).
+	if err := os.Rename(src, dst); err == nil {
+		return nil
+	}
```
craciunoiuc, line 341: Can rename fail for other reasons also?

unikraft/target/target.go
```diff
@@ -121,6 +124,15 @@ func (tc *TargetConfig) Kernel() string {
 	return tc.kernel
 }
 
+func (tc *TargetConfig) SetKernelPath(path string) {
+	tc.kernel = path
+	if tc.kconfig == nil {
+		tc.kconfig = kconfig.KeyValueMap{}
+	}
+	// Requirement (b): Update the KConfig with the basename
+	tc.kconfig.Set(unikraft.UK_NAME, filepath.Base(path))
```
craciunoiuc: This did not update for me, maybe the update is not done at the right moment? It looks like it is done after the kernel is moved

craciunoiuc (changes requested):
Hey I tried out the PR now. It sort of works, but I left some comments. Thanks!

internal/cli/kraft/build/build.go
ijaidev: Done, removed in 2f25484

internal/cli/kraft/build/build.go
ijaidev: Yes, it could also fail due to permissions, a read-only filesystem, or the file being busy. The reason I used a general fallback is that if os.Rename fails for any of those reasons, it will fall back to the copy method.

unikraft/target/target.go
ijaidev: You're right, the KConfig update was happening too late to have any effect on the build.

After re-evaluating the original issue, I've simplified the implementation. I've removed the KConfig update entirely to keep things straightforward and avoid overcomplicating the build process. The latest version of the PR now just performs a reliable move of the final binary once the build succeeds.

internal/cli/kraft/build/build.go
craciunoiuc: Yes, and does it make sense to try to copy in these cases if the rename failed? That's what I'm asking

Maybe it makes sense to fail early

ijaidev: I don’t think there’s any issue with trying to copy in this case. The original file will be deleted after the copy completes. The file is auto-generated, so permissions shouldn’t be a problem; and if another program is using it, the delete operation will fail, causing this operation to fail as well.

craciunoiuc: Ok, all good, I'll assume that any problem that would have in `rename` would be caught anyway about the other steps below

craciunoiuc (approved):
All good here. Thanks!

ijaidev:
Happy to contribute!

PR #2630, opened 2026-01-27
fix: Preserve original user's config path when running with sudo

### LINKED ISSUE #992: `kraft net` almost unusable with multiple users
### Describe the bug

Trying to use the `kraft net` subcommands together with `kraft run` is almost impossible when using multiple users.

In order to create a network interface you need to use `sudo`, and then you should be able to use it inside `kraft run`.

But, to use it there, you also need to use `sudo` for `kraft run`. This means that the user is changed and the path to the config also becomes changed (effectively using the home of `root`).

This can prove confusing as people ca run:
1. `sudo kraft run`
1. `kraft ps`

And then see no output, even though they should.

### Steps to reproduce

_No response_

### Expected behavior

_No response_

x86_64, arm, arm64

linux/debian, linux/fedora, linux/alpine, linux/arch, linux/other

### Relevant log output

_No response_

ijaidev:
# fix: Preserve original user's config path when running with sudo

## Description

Fixes #992

When `kraft` is invoked via `sudo`, `os.UserHomeDir()` returns `/root`, causing machine state to be stored in `/root/.local/share/kraftkit/runtime/` instead of the user's home directory. This makes machines created with `sudo kraft run` invisible to non-sudo commands like `kraft ps`.

### `config/config_file.go`
- Added `getSudoUserInfo()` - detects sudo via `SUDO_USER` environment variable and retrieves the original user's UID, GID, and home directory
- Added `getHomeDir()` - returns the original user's home directory when running under sudo
- Added `ChownToUser()` / `ChownToUserRecursive()` - fixes file ownership for files created by sudo
- Updated `ConfigDir()`, `StateDir()`, `DataDir()` to use `getHomeDir()` instead of `os.UserHomeDir()`

### Machine Drivers
Added `config.ChownToUser(machine.Status.StateDir)` after directory creation in:
- `machine/firecracker/v1alpha1.go`
- `machine/qemu/v1alpha1.go`
- `machine/xen/v1alpha1.go`

### `store/embedded.go`
- Added `config.ChownToUserRecursive(store.path)` after BadgerDB operations
- Restructured `Create()` and `Delete()` to close DB before chowning (ensures files written during close() get correct ownership)

## Testing

**Before this fix:**
```bash
$ sudo kraft run --plat fc -d ./build/app
machine_name

$ kraft ps -a
NAME  KERNEL  ARGS  CREATED  STATUS  MEM  PORTS  PLAT
# Empty - machine not visible!
```

**After this fix:**
```bash
$ sudo kraft run --plat fc -d ./build/app
compassionate_missbaker

$ kraft ps -a   # WITHOUT sudo ✓
NAME                     KERNEL  STATUS  PLAT
compassionate_missbaker  pr...   exited  fc
```

**File ownership:**
```bash
$ ls -la ~/.local/share/kraftkit/runtime/machinev1alpha1/
drwx------ 2 jaidev jaidev 000001.sst
drwx------ 2 jaidev jaidev 000002.sst
drwx------ 2 jaidev jaidev 000003.sst  # Previously owned by root!
```

## Checklist

- [x] Code follows project conventions
- [x] Commit message follows contribution guidelines
- [x] `make fmt` passes
- [x] Manual testing completed

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -139,6 +139,8 @@ jobs:
           KRAFTKIT_NO_WARN_SUDO: true
           KRAFTKIT_NO_CHECK_UPDATES: true
           DOCKER: ''
+          KRAFTKIT_DATA_DIR: /tmp/kraftkit-data
+          KRAFTKIT_STATE_DIR: /tmp/kraftkit-state
         run: make test-e2e DISTDIR="$(go env GOPATH)"/bin
 
   e2e-cloud-cli:
--- a/config/config_file.go
+++ b/config/config_file.go
@@ -27,12 +27,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"os"
-	"os/user"
 	"path/filepath"
 	"runtime"
-	"strconv"
 	"syscall"
 )
 
@@ -44,114 +41,6 @@ const (
 	APP_DATA            = "AppData"
 	LOCAL_APP_DATA      = "LocalAppData"
 )
-
-// sudoUserInfo contains information about the original user when running under sudo.
-type sudoUserInfo struct {
-	HomeDir string
-	UID     int
-	GID     int
-}
-
-// getSudoUserInfo returns information about the original user when running under sudo.
-// Returns nil if not running under sudo or if the original user cannot be determined.
-func getSudoUserInfo() *sudoUserInfo {
-	sudoUser := os.Getenv("SUDO_USER")
-	if sudoUser == "" {
-		return nil
-	}
-
-	// Try to get UID/GID from environment first (more reliable)
-	uidStr := os.Getenv("SUDO_UID")
-	gidStr := os.Getenv("SUDO_GID")
-
-	var uid, gid int
-	var homeDir string
-
-	// Look up the user to get their home directory
-	if u, err := user.Lookup(sudoUser); err == nil {
-		homeDir = u.HomeDir
-		// Use looked up UID/GID as fallback
-		if uidStr == "" {
-			uidStr = u.Uid
-		}
-		if gidStr == "" {
-			gidStr = u.Gid
-		}
-	}
-
-	if homeDir == "" {
-		return nil
-	}
-
-	// Parse UID
-	if uidStr != "" {
-		if parsedUID, err := strconv.Atoi(uidStr); err == nil {
-			uid = parsedUID
-		}
-	}
-
-	// Parse GID
-	if gidStr != "" {
-		if parsedGID, err := strconv.Atoi(gidStr); err == nil {
-			gid = parsedGID
-		}
-	}
-
-	return &sudoUserInfo{
-		HomeDir: homeDir,
-		UID:     uid,
-		GID:     gid,
-	}
-}
-
-// getHomeDir returns the appropriate home directory for kraftkit configuration.
-// When running under sudo, this returns the original user's home directory
-// to maintain consistent config paths between privileged and unprivileged commands.
-func getHomeDir() string {
-	if info := getSudoUserInfo(); info != nil {
-		return info.HomeDir
-	}
-	homeDir, err := os.UserHomeDir()
-	if err != nil {
-		log.Printf("warning: could not determine home directory: %v", err)
-	}
-	return homeDir
-}
-
-// ChownToUser changes ownership of a path to the original user when running under sudo.
-// This ensures that files created by sudo commands are accessible by the regular user.
-// If not running under sudo, this is a no-op.
-func ChownToUser(path string) error {
-	info := getSudoUserInfo()
-	if info == nil || info.UID == 0 {
-		// Not running under sudo or couldn't determine original user
-		return nil
-	}
-
-	return os.Chown(path, info.UID, info.GID)
-}
-
-// ChownToUserRecursive changes ownership of a path and all its contents to the original user.
-// This is useful for directories created by sudo commands.
-// Only files/dirs owned by root are chowned to avoid unnecessary syscalls.
-func ChownToUserRecursive(path string) error {
-	info := getSudoUserInfo()
-	if info == nil || info.UID == 0 {
-		return nil
-	}
-
-	return filepath.Walk(path, func(name string, fileInfo os.FileInfo, err error) error {
-		if err != nil {
-			return err
-		}
-		if stat, ok := fileInfo.Sys().(*syscall.Stat_t); ok {
-			if stat.Uid != 0 {
-				return nil
-			}
-		}
-		return os.Chown(name, info.UID, info.GID)
-	})
-}
 
 // Config path precedence
 // 1. KRAFTKIT_CONFIG_DIR
--- a/config/config_file.go
+++ b/config/config_file.go
@@ -27,9 +27,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"os"
+	"os/user"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"syscall"
 )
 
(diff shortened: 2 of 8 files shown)


config/config_file.go
```diff
@@ -42,6 +44,105 @@ const (
...
+	info := getSudoUserInfo()
+	if info == nil || info.UID == 0 {
+		return nil
+	}
+
+	return filepath.Walk(path, func(name string, fileInfo os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		return os.Chown(name, info.UID, info.GID)
```
craciunoiuc: Would it be faster to only chown files/dirs that are owned by `root`?

config/config_file.go
```diff
@@ -42,6 +44,105 @@ const (
...
+}
+
+// getHomeDir returns the appropriate home directory for kraftkit configuration.
+// When running under sudo, this returns the original user's home directory
+// to maintain consistent config paths between privileged and unprivileged commands.
+func getHomeDir() string {
+	if info := getSudoUserInfo(); info != nil {
+		return info.HomeDir
+	}
+	homeDir, _ := os.UserHomeDir()
```
craciunoiuc: Can you log the error as a warning now that we do it in a central function?

machine/qemu/v1alpha1.go
```diff
@@ -159,6 +159,11 @@ func (service *machineV1alpha1Service) Create(ctx context.Context, machine *mach
 		return machine, err
 	}
 
+	// Ensure the directory is owned by the original user when running under sudo
+	if err := config.ChownToUser(machine.Status.StateDir); err != nil {
```
craciunoiuc: The state directory is chowned, but what about the files inside? I guess these are also checked for?

store/embedded.go
```diff
@@ -138,19 +145,30 @@ func (store *embedded[_, _]) Create(ctx context.Context, key string, _, out runt
...
+		return err
+	}
+
+	// Close the database first, then chown all files
+	if err := store.close(); err != nil {
+		return err
+	}
+
+	// Ensure any new files created during the transaction are owned by the original user
+	return config.ChownToUserRecursive(store.path)
```
craciunoiuc: In this case, in case of a failure the file is not chowned, I think this si not intended?

I would suggest you do something like

```
defer func() {
	store.close()
	config.ChownToUserRecursive(store.path)
}
```

store/embedded.go
```diff
@@ -159,17 +177,27 @@ func (store *embedded[_, _]) Delete(ctx context.Context, key string, out runtime
...
```
craciunoiuc: Same as above?

craciunoiuc (changes requested):
Some comments my side. Changes look good.

We need to decide if we should handle all cases, or only the cases where a "mandatory sudo" might happen

config/config_file.go
ijaidev: you are right 
I have updated it

config/config_file.go
ijaidev: done

machine/qemu/v1alpha1.go
ijaidev: done, updated recursively

store/embedded.go
ijaidev: you are right, 
I have updated the code

store/embedded.go
ijaidev: done

config/config_file.go
craciunoiuc: I see you added:
```
			if stat.Uid != 0 {
				return nil
			}
```

I would go even further to say this should be `stat.Uid != info.Uid` or something like this. This would cover the very rare case where a different user creates files in another user's home directory

config/config_file.go
```diff
@@ -42,6 +45,114 @@ const (
...
+// Returns nil if not running under sudo or if the original user cannot be determined.
+func getSudoUserInfo() *sudoUserInfo {
+	sudoUser := os.Getenv("SUDO_USER")
+	if sudoUser == "" {
+		return nil
+	}
+
+	// Try to get UID/GID from environment first (more reliable)
+	uidStr := os.Getenv("SUDO_UID")
+	gidStr := os.Getenv("SUDO_GID")
```
craciunoiuc:
```suggestion
	sudoUser := os.Getenv("SUDO_USER") // <--
	if sudoUser == "" {
		return nil
	}

	// Try to get UID/GID from environment first (more reliable)
	uidStr := os.Getenv("SUDO_UID") // <--
	gidStr := os.Getenv("SUDO_GID") // <--
```

Can you make sure those env variables also exist in MacOS?

config/config_file.go
```diff
@@ -42,6 +45,114 @@ const (
...
+	info := getSudoUserInfo()
+	if info == nil || info.UID == 0 {
+		return nil
+	}
+
+	return filepath.Walk(path, func(name string, fileInfo os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if stat, ok := fileInfo.Sys().(*syscall.Stat_t); ok {
```
craciunoiuc: I think this should work on `Mac`, I'm not 100% sure, but seems fine.

Still, I would highly suggest you move this function to a separate `config/utils_unix.go` and create a separate `config/utils_windows.go` where you stub it.

I know we don't really support compiling windows at this time, but it's better to have at least stubbed compatibility.

Move there also the `getSudoUserInfo` function as it's also Unix specific (so all, but not windows)

.github/workflows/tests.yaml
```diff
@@ -139,6 +139,8 @@ jobs:
           KRAFTKIT_NO_WARN_SUDO: true
           KRAFTKIT_NO_CHECK_UPDATES: true
           DOCKER: ''
+          XDG_DATA_HOME: /tmp/kraftkit-data
```
craciunoiuc: `XDG_*` Variables are Mac specific

I understand why this fix worked in the context of Kraftkit (checks are in order), but it's hacky to do this 😔 

I would suggest we do this properly and add variables for the other fields also like we have for:
```
	if a := os.Getenv(KRAFTKIT_CONFIG_DIR); a != "" {
		path = a
	}
```

So that will be a similar change for `KRAFTKIT_STATE_DIR` and `KRAFTKIT_DATA_DIR`.

And here in this file you would set `KRAFTKIT_STATE_DIR` and `KRAFTKIT_DATA_DIR` to these same values

config/utils_unix.go
```diff
@@ -0,0 +1,152 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+//go:build unix
```
craciunoiuc: Did this work? can you also try `!windows` instead of `unix`?

config/config_file.go
```diff
@@ -70,41 +71,46 @@ func ConfigDir() string {
...
 	var path string
-	if a := os.Getenv(XDG_STATE_HOME); a != "" {
-		path = filepath.Join(a, "kraftkit")
-	} else if b := os.Getenv(LOCAL_APP_DATA); runtime.GOOS == "windows" && b != "" {
-		path = filepath.Join(b, "KraftKit")
+	if a := os.Getenv(KRAFTKIT_STATE_DIR); a != "" {
+		path = a
+	} else if b := os.Getenv(XDG_STATE_HOME); b != "" {
+		path = filepath.Join(b, "kraftkit")
+	} else if c := os.Getenv(LOCAL_APP_DATA); runtime.GOOS == "windows" && c != "" {
```
craciunoiuc:
```suggestion
	} else if c := os.Getenv(LOCAL_APP_DATA); c != "" {
```

config/config_file.go
```diff
@@ -70,41 +71,46 @@ func ConfigDir() string {
...
 	var path string
-	if a := os.Getenv(XDG_DATA_HOME); a != "" {
-		path = filepath.Join(a, "kraftkit")
-	} else if b := os.Getenv(LOCAL_APP_DATA); runtime.GOOS == "windows" && b != "" {
-		path = filepath.Join(b, "KraftKit")
+	if a := os.Getenv(KRAFTKIT_DATA_DIR); a != "" {
+		path = a
+	} else if b := os.Getenv(XDG_DATA_HOME); b != "" {
+		path = filepath.Join(b, "kraftkit")
+	} else if c := os.Getenv(LOCAL_APP_DATA); runtime.GOOS == "windows" && c != "" {
```
craciunoiuc:
```suggestion
	} else if c := os.Getenv(LOCAL_APP_DATA); c != "" {
```

craciunoiuc (approved):
All good here. Thanks!

ijaidev:
Glad to contribute!

PR #2652, opened 2026-02-16
fix(build): Validate Dockerfile using buildkit parser

Syedowais312:
This PR improves Dockerfile detection in `builder_dockerfile`.

Previously, detection relied only on filename matching which
could incorrectly accept non-Dockerfiles or reject valid ones.

The implementation now:
- Parses the file using the BuildKit Dockerfile parser
- Ensures a valid AST exists
- Verifies at least one FROM instruction is present
- Keeps warnings when filename suggests Dockerfile but syntax invalid

This makes Dockerfile handling reliable and removes dependence
on naming conventions.

### DIFF
--- a/internal/cli/kraft/build/builder_dockerfile.go
+++ b/internal/cli/kraft/build/builder_dockerfile.go
@@ -7,8 +7,11 @@ package build
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 
+	"github.com/moby/buildkit/frontend/dockerfile/parser"
 	"kraftkit.sh/log"
 )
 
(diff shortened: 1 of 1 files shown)


Syedowais312:
ping for review.
CC: @nderjung @craciunoiuc

craciunoiuc (changes requested):
Hey copilot found some stuff, at least one comment is valid

Can you go through them and fix the ones that make sense?

Syedowais312:
> Hey copilot found some stuff, at least one comment is valid
> 
> Can you go through them and fix the ones that make sense?

Yup,  I have made the changes

internal/cli/kraft/build/builder_dockerfile.go
```diff
@@ -43,16 +46,70 @@ func (build *builderDockerfile) Buildable(ctx context.Context, opts *BuildOption
 		opts.RootfsType = opts.Project.InitrdFsType()
 	}
 
-	// TODO(nderjung): This is a very naiive check and should be improved,
-	// potentially using an external library which parses the Dockerfile syntax.
-	// In most cases, however, the Dockerfile is usually named `Dockerfile`.
-	if !strings.Contains(strings.ToLower(opts.Rootfs), "dockerfile") {
+	//D ockerfile validation using buildkit parser
```
craciunoiuc: Try again, third time's the charm

Syedowais312:
I was wondering if it would be great to add test cases for this validation logic? 

Should I add them in this PR or create a follow-up?

CC: @craciunoiuc

craciunoiuc:
Sure, add the tests in the same PR as a separate commit

internal/cli/kraft/build/builder_dockerfile.go
```diff
@@ -43,16 +46,70 @@ func (build *builderDockerfile) Buildable(ctx context.Context, opts *BuildOption
 		opts.RootfsType = opts.Project.InitrdFsType()
 	}
 
-	// TODO(nderjung): This is a very naiive check and should be improved,
-	// potentially using an external library which parses the Dockerfile syntax.
-	// In most cases, however, the Dockerfile is usually named `Dockerfile`.
-	if !strings.Contains(strings.ToLower(opts.Rootfs), "dockerfile") {
+	// Dockerfile validation using buildkit parser
+	if !isDockerfile(ctx, opts.Rootfs) {
```
jedevc, line 50: 🤔

This makes the kraftfile significantly more difficult to parse + handle. To know how we'd do this, we need to bundle the entire buildkit dockerfile parser?

I hit this when working on https://github.com/unikraft-cloud/x/pull/143. Potentially, instead of having `isDockerfile`, we should instead keep all the existing heuristics for backwards compat, and introduce a new `type` field, which allows the user to explicitly define what to do.

Syedowais312: @ijaidev thanks for the detailed feedback.

My intention was mainly to avoid false negatives for valid Dockerfiles with non-standard names, but I agree builder selection probably shouldn’t depend on deep parsing semantics.

Introducing an explicit type while keeping the current heuristics for backwards compatibility sounds cleaner and preserves existing behavior.
I’ve also been thinking about a few alternative approaches and would appreciate your thoughts on them as well. I’m exploring those now and will come back shortly.

Syedowais312: After exploring further, I think the cleanest approach is to introduce an explicit source-type field in the kraftfile rootfs spec, along the lines of what you suggested: 
```yaml
 rootfs: 
    source: ./my-custom-named-file 
    source-type: dockerfile
```
Does this direction match what you had in mind, or would you prefer the type to live at a different level (e.g. part of a generic rootfs “kind” instead of dockerfile-specific)?

jedevc:
As discussed, we should not take this approach. We should be doing this: https://github.com/unikraft/kraftkit/pull/2652#discussion_r2828670713

PR #2660, opened 2026-02-24
Fix/yamlmerger deep merge

Syedowais312:
## Summary

This PR fixes scalar node merge behavior in RecursiveMerge
by replacing pointer reassignment with value overwrite.

It also improves sequence merging by introducing deep
node comparison, enabling proper handling of sequences
of mappings and nested structures.

## Changes

- Fix scalar overwrite behavior
- Implement recursive deep equality for sequences
- Update and extend test coverage

## Testing

All tests pass with:

go test ./internal/yamlmerger/...

### DIFF
--- a/internal/yamlmerger/yamlmerger.go
+++ b/internal/yamlmerger/yamlmerger.go
@@ -53,13 +53,12 @@ func RecursiveMerge(from, into *yaml.Node) error {
 			}
 		}
 	case yaml.ScalarNode:
-		// SA4006 these variables represent pointers and are propagated outside of `recursiveMerge`
-		into = from //nolint:staticcheck
+		*into = *from
 	case yaml.SequenceNode:
 		for _, fromItem := range from.Content {
 			foundFrom := false
 			for _, intoItem := range into.Content {
-				if fromItem.Value == intoItem.Value {
+				if nodesDeepEqual(fromItem, intoItem) {
 					foundFrom = true
 				}
 			}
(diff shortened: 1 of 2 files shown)


internal/yamlmerger/yamlmerger.go
```diff
@@ -53,13 +53,12 @@ func RecursiveMerge(from, into *yaml.Node) error {
 			}
 		}
 	case yaml.ScalarNode:
-		// SA4006 these variables represent pointers and are propagated outside of `recursiveMerge`
-		into = from //nolint:staticcheck
+		*into = *from
```
Syedowais312, line 56: Previously, this reassigned the local pointer (`into = from`), which did not modify the underlying node. Using `*into = *from` correctly overwrites the target node value.

internal/yamlmerger/yamlmerger.go
```diff
@@ -53,13 +53,12 @@ func RecursiveMerge(from, into *yaml.Node) error {
...
 	case yaml.SequenceNode:
 		for _, fromItem := range from.Content {
 			foundFrom := false
 			for _, intoItem := range into.Content {
-				if fromItem.Value == intoItem.Value {
+				if nodesDeepEqual(fromItem, intoItem) {
```
Syedowais312, line 61: also replaced shallow `.Value` comparison with structural deep equality. `.Value` works for scalar nodes but fails for mappings and nested structures.


@nderjung @craciunoiuc could you please take a look

Issue #673, opened 2023-07-25
Configurable toolchain


nderjung:
### Overview

Unikraft is a highly modular library operating system designed for the cloud.  It's high degree of modularization allows for extreme customization and specialization. As such, its tooling should not interfere with the user's desire to support such customization. Towards increasing the unikernel's developer's ability to customize the build whilst simultaneously systemitizing the process of retrieving, organizing and generally facilitating the build of a unikernel based on Unikraft and its many components, the supported tooling, `kraft`, should allow for the injection of the user's environment and or additional toolchain requirements. 

### Feature request summary

This project is designed to better facilitate the dynamic injection of user provided variables into Unikraft's build system through the addition of a dynamically configured `toolchain` towards greater customization of the unikernel build through the use of its command-line companion client tool, `kraft`. This manifests itself as an injection into KraftKit's core configuration system and must propagate across the codebase appropriately.

Distinct results of this addition would enable, but are not limited to: alternating the GNU Compiler Collection (GCC) version, providing alternative compile-time flags, and more.

This project tracks the addition of a map to the internal configuration system entitled `toolchain` which sets the values of programs that are ultimately passed to Unikraft's build system.

The addition is first made as an entry to the [`KraftKit`](https://github.com/unikraft/kraftkit/blob/staging/config/config.go#L17) config structure in 
```diff
  type KraftKit struct {
    [...]
+   Toolchain map[string]string `yaml:"toolchain" long:"toolchain" usage:"Set specific global programs in Unikraft's build system"
    [...]
  }
```

These values need to be passed any `make` invocations as in-line variables via the option [`make.WithVar`](https://github.com/unikraft/kraftkit/blob/staging/make/options.go#L116) in relevant parts of KraftKit's commands that interact with the build, namely in [`cmd/kraft/build/build.go`](https://github.com/unikraft/kraftkit/blob/staging/cmd/kraft/build/build.go) since it is in these files that access to the map is possible via `config.G[config.KraftKit](ctx).Toolchain`.

Additionally, the specific entry for `CC` should be added as default during the installation process, i.e. the user should be asked whether they want to change the default value for `CC` which should be set to `gcc`.

A sample change to the user's configuration could look something like:

```yaml
toolchain:
  CC: /path/to/alternative/gcc
  UK_CFLAGS: -fdiagnostics-color=always
```

~~The same top-level `toolchain` should also be added as a directive/element into the `Kraftfile`, with the ability to specify it per-target.~~

[...]

PR #2672, opened 2026-03-02
feat: Introduce global Toolchain config map

moghit-eou:
### Prerequisite checklist

### Description of changes
Adds a `Toolchain map[string]string` field to the `KraftKit` config
struct in `config/config.go` as the first step towards #673.

The field follows the existing pattern of `Auth` and `Aliases` which
are both maps using `noattribute:"true"`.

The toolchain variables are injected into any `make` invocation via
`make.WithVars()` in `builder_kraftfile_unikraft.go`, accessible via
`config.G[config.KraftKit](ctx).Toolchain`.

Allowing users to set global toolchain variables in their KraftKit
config file which will be passed directly to Unikraft's build system.
For example:

```
toolchain:
  CC: /path/to/alternative/gcc
  UK_CFLAGS: -fdiagnostics-color=always
```

Remaining work tracked in #673:
- Initialize CC: gcc default in config/defaults.go
- Add toolchain directive to Kraftfile schema and loader
- Add per-target override support in Kraftfile targets
- Add kraft toolchain set/list/unset CLI commands

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -54,6 +54,8 @@ type KraftKit struct {
 	Auth map[string]AuthConfig `yaml:"auth,omitempty" noattribute:"true"`
 
 	Aliases map[string]map[string]string `yaml:"aliases" noattribute:"true"`
+
+	Toolchain map[string]string `yaml:"toolchain,omitempty" noattribute:"true"`
 }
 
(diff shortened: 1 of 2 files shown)


nderjung (approved):
Thanks!

PR #2675, opened 2026-03-03
fix: Avoid overwriting fstype with empty value

jedevc:
### Prerequisite checklist

### Description of changes

This value will be set to the empty string from the CLI if it's not explicitly set with `--rootfs-type`. This essentially results in overwriting the pre-populated default cpio type with an empty fs type.
    
This then just silently skips the build in older versions of kraft, or causes a weirdly strange internal error on staging.

### DIFF
--- a/initrd/detect.go
+++ b/initrd/detect.go
@@ -7,6 +7,8 @@ package initrd
 import (
 	"context"
 	"fmt"
+
+	"kraftkit.sh/log"
 )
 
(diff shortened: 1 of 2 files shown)


craciunoiuc (approved):
All good here. Thanks!

craciunoiuc (changes requested):
removing my review cause I want to see how this works with my PR 👀

PR #2676, opened 2026-03-03
fix: Several fixes and features for rootfs file systems

craciunoiuc:
### Prerequisite checklist

### Description of changes

The fixes and their reasonings are:
* C1 -- `erofs->cpio` fail was mentioned but somehow missing, readd it
* C2 -- copying `erofs` files was possible, but `cpio` files not because of the custom writer; use a direct one
* C3 -- an unused `file` type was hiding, use it to copy any file as initrd, not just cpio/erofs (notice files on disk will have `.file` extension)
* C4,5 -- Add the unused `unknown` type to be used when specifying files but you want autodetect the type fully. A bit useless but might be turned useful later if we want to do even more detection

Closes: #2675

### DIFF
--- a/fsutils/fsutils.go
+++ b/fsutils/fsutils.go
@@ -126,6 +126,16 @@ func IsCpioFile(path string) bool {
 	return string(magic) == "070701" || string(magic) == "070702"
 }
 
+// IsRegularFile checks if the given path is a regular file.
+func IsRegularFile(path string) bool {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+
+	return fi.Mode().IsRegular()
+}
+
 type FileInfo struct {
 	Uid  int
 	Gid  int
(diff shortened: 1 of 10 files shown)


fsutils/fsutils.go
```diff
@@ -126,6 +126,16 @@ func IsCpioFile(path string) bool {
 	return string(magic) == "070701" || string(magic) == "070702"
 }
 
+// IsGenericFile checks if the given path is a regular file.
+func IsGenericFile(path string) bool {
```
nderjung: Probably a better name here (since it's public) is `IsRegularFile`

nderjung (commented):
This looks good to me, just one comment about the name of a method.

I've also tested locally with a couple of scenarios and it works for me 👍 thanks!

Will merge after the rename :)

nderjung (approved):
Thanks!

PR #2681, opened 2026-03-05
fix(pkg/pull): Route git/GitHub URLs to manifest provider

### LINKED ISSUE #1114: `kraft pkg pull` no longer works with GitHub sources
### Describe the bug

In previous versions of kraft, commands like this used to work:
```
kraft pkg pull github.com/unikraft/lib-nginx.git
W  could not get index: GET https://github.com/v2/unikraft/lib-nginx.git/manifests/latest: unexpected status code 404 Not Found: Not Found
W  could not find github.com/unikraft/lib-nginx.git
E  no processes to perform
```

### Steps to reproduce

1. run `kraft pkg update`
1. run `kraft pkg pull github.com/unikraft/lib-nginx.git`

### Expected behavior

Library is pulled in the path: `.unikraft/libs/<libname>`

x86_64

linux/debian

### Relevant log output

_No response_

robaa12:
### Prerequisite checklist

### Description of changes

Closes #1114 `kraft pkg pull` failed to handle git/GitHub URLs such as:

- `github.com/unikraft/lib-nginx.git`
- `https://github.com/unikraft/lib-nginx`
- `git@github.com:unikraft/lib-nginx.git`

Instead of routing them to the manifest provider, they were either claimed by the OCI manager or crashed with a nil pointer dereference.

| # | Root Cause | Fix |
|---|-----------|-----|
| 1 | `name.ParseReference` treats `github.com` as a valid registry host, so `OCIManager.IsCompatible` claimed git/GitHub URLs before the manifest manager could | Added `isGitSource()` to reject these URLs early in `IsCompatible` and `Catalog()` |
| 2 | git/GitHub args in `pull.go` were passed as `WithName`, but the manifest manager's `Catalog()` only resolves a provider when `WithSource` is set | Changed to use `WithSource` for all git/GitHub args |
| 3 | `ManifestManager.IsCompatible` had no recognition of git/GitHub URL patterns, so it never short-circuited to claim them | Added an early-return guard matching `.git`, `git@`, `github.com/`, and `://github.com/` |
| 4 | When both `NewGitProvider` and `NewGitHubProvider` failed, a nil provider was returned without error, causing a nil pointer dereference in `Manifests()` | Added a nil guard before the fallback return in `NewProvider` |
| 5 | `NewGitProvider` converted `git@HOST:path` to `ssh://git@HOST:path` (colon kept), which URL parsers reject | Fixed to produce `ssh://git@HOST/path` (colon replaced with slash) |
| 6 | `ghrepo.NewFromURL` used `url.Parse` directly, which cannot handle SCP-style SSH or bare `github.com/` URLs, resulting in an empty host | Normalize both forms to `https://` before parsing |

### DIFF
--- a/internal/ghrepo/repo.go
+++ b/internal/ghrepo/repo.go
@@ -49,6 +49,16 @@ func New(owner, repo string) Interface {
 
 // NewFromURL parses a given GitHub url and returns the populated Interface
 func NewFromURL(path string) (Interface, error) {
+	// Normalize SCP-style SSH URLs (git@github.com:owner/repo.git) to
+	// https:// so url.Parse handles them correctly.
+	if strings.HasPrefix(path, "git@") {
+		path = strings.Replace(path, ":", "/", 1)
+		path = strings.TrimPrefix(path, "git@")
+		path = "https://" + path
+	} else if strings.HasPrefix(path, "github.com/") {
+		path = "https://" + path
+	}
+
 	u, err := url.Parse(path)
 	if err != nil {
 		return nil, fmt.Errorf("could not parse url: %s", err)
--- a/internal/ghrepo/repo.go
+++ b/internal/ghrepo/repo.go
@@ -55,6 +55,14 @@ func NewFromURL(path string) (Interface, error) {
 		path = strings.Replace(path, ":", "/", 1)
 		path = strings.TrimPrefix(path, "git@")
 		path = "https://" + path
+	} else if strings.HasPrefix(path, "ssh+git@") {
+		path = strings.Replace(path, ":", "/", 1)
+		path = strings.TrimPrefix(path, "ssh+git@")
+		path = "https://" + path
+	} else if strings.HasPrefix(path, "ssh://git@") {
+		path = strings.TrimPrefix(path, "ssh://git@")
+		path = strings.Replace(path, ":", "/", 1)
+		path = "https://" + path
 	} else if strings.HasPrefix(path, "github.com/") {
 		path = "https://" + path
 	}
--- a/internal/gitutil/detector.go
+++ b/internal/gitutil/detector.go
@@ -26,6 +26,10 @@ func IsGitSource(source string) bool {
 		}
 	}
 
+	if strings.HasPrefix(source, "ssh+git@") {
+		return true
+	}
+
 	// Check for SCP-style SSH: git@host:path
 	if strings.HasPrefix(source, "git@") {
 		return true
--- a/internal/gitutil/detector.go
+++ b/internal/gitutil/detector.go
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The KraftKit Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+package gitutil
+
+import (
+	"net/url"
+	"strings"
+)
+
+// IsGitSource returns true if the source string appears to be a git repository
+// URL rather than an OCI image reference. It handles:
+// - SSH URLs: ssh://git@host/path, git@host:path, git+ssh://, ssh+git://
+// - HTTPS URLs: https://host/path.git, https://github.com/owner/repo
+// - Bare paths: github.com/owner/repo, gitlab.com/owner/repo
+func IsGitSource(source string) bool {
+	if source == "" {
+		return false
+	}
+
+	// Check for explicit SSH URL schemes
+	for _, prefix := range []string{"ssh://", "ssh+git://", "git+ssh://", "git://"} {
+		if strings.HasPrefix(source, prefix) {
+			return true
+		}
+	}
+
+	// Check for SCP-style SSH: git@host:path
+	if strings.HasPrefix(source, "git@") {
+		return true
+	}
+
+	// Try to parse as URL to distinguish from OCI refs
+	// Normalize common patterns first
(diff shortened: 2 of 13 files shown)


craciunoiuc:
Can you create commit for each fix in the table please? thanks 🙏

robaa12:
Hi @craciunoiuc , I implemented all the requested changes. Please take another look when you get a chance. Happy to make any further changes if needed.

robaa12:
I am working on the copilot changes now. Will push an update soon.

robaa12:
Hey @craciunoiuc!

I addressed the Copilot feedback by creating a shared `gitutil` package instead of patching the inline code. Here's what I did:

**What changed**

1. **New `internal/gitutil` package** : Created a centralized `IsGitSource()` function that properly detects git URLs vs OCI images. Now handles all SSH URL variants (`ssh://`, `git@`, `git+ssh://`, etc.) and correctly distinguishes OCI images like `ghcr.io/org/repo.git:latest`  from actual git sources.

2. **Updated the 3 files** : `oci/manager.go`, `manifest/manager.go`, and `pull.go` now all use this shared function instead of having the same detection logic copy-pasted in 3 places.

**Why did I choose this approach?**

- Copilot pointed out that the old code didn't handle SSH URLs without `.git` suffix and could have false positives with OCI images. Having one centralized function makes it easier to fix and test.
- Removed duplicate test code — the core logic is tested in one place now.
- All tests cases pass locally (`make fmt` also passes).

craciunoiuc (changes requested):
Hey, can you squash this commit? afterwards it will be good to merge

Image

robaa12:
Done, please have another look when u get a chance.

craciunoiuc:
Hey, this did not work for me with these cases, am I doing something wrong?
```
➜  test-pr ~/kraftkit/main pkg pull "git@github.com:unikraft/app-nginx.git"
 i  finding *
 E  could not find local reference for *
 i  finding *
 E  could not find *
 E  could not find *
➜  test-pr ~/kraftkit/main pkg pull "ssh+git@github.com:unikraft/app-nginx.git"
 i  finding *
 E  could not find local reference for *
 i  finding *
 E  could not find *
 E  could not find *
```

also note that it's showing a wildcard everywhere instead of the URL

robaa12:
reproduced it on my end too.

I'm working on a follow-up fix for those exact `git@...` / `ssh+git@...` cases (and a couple of related SSH URL edge cases). I've tracked down where it breaks in the URL normalization/provider path and I'm testing the fix now.

Also, the `*` you saw in `finding *` is a query rendering thing , source-only lookups were printing the `name` field, and when `name` is empty it just falls back to `*`. I am working on it.

I'll push an update soon and drop another comment with exactly what changed.

robaa12:
### What was going wrong
1. `git@...` / `ssh+git@...` still had gaps in normalization/parsing across the provider flow.
2. `finding *` happened because source-only queries were rendered from `name`; when `name` was empty it fell back to `*`.
3. GitHub probing could still depend on raw SSH-like input in some paths, which could fail even for public repos.
### What I changed
- `packmanager/query.go`
  - show `source` in `Query.String()` when `name` is empty (so no more wildcard `*` in logs/errors for URL pulls)
- SSH URL handling improvements
  - normalize additional SSH-style variants (`ssh+git@...`, `ssh://git@...:owner/repo...`) in:
    - `internal/gitutil/detector.go`
    - `internal/ghrepo/repo.go`
    - `manifest/provider_git.go`
  - added test coverage for these URL forms in:
    - `internal/gitutil/detector_test.go`
    - `internal/ghrepo/repo_test.go`
- `manifest/provider_github.go`
  - probe via normalized HTTPS for GitHub sources
  - preserve `manifest.Origin` as the original input source so source filtering still matches correctly
### Verification
I rebuilt and tested `./bin/kraft` manually with multiple formats, including the two you reported.
Also validated additional forms (`https://`, bare `github.com/...`, `ssh://git@...` slash/colon, `git+ssh://`, `ssh+git://`) and all passed in my local.

internal/ghrepo/repo.go
```diff
@@ -49,6 +49,24 @@ func New(owner, repo string) Interface {
...
+		path = strings.Replace(path, ":", "/", 1)
+		path = strings.TrimPrefix(path, "ssh+git@")
+		path = "https://" + path
+	} else if strings.HasPrefix(path, "ssh://git@") {
+		path = strings.TrimPrefix(path, "ssh://git@")
+		path = strings.Replace(path, ":", "/", 1)
+		path = "https://" + path
+	} else if strings.HasPrefix(path, "github.com/") {
+		path = "https://" + path
+	}
```
craciunoiuc, line 68: Is there no library that has all these checks, it doesn't look very sustainable to keep a list of all formats possible

internal/gitutil/detector.go
```diff
@@ -0,0 +1,107 @@
...
+		return true
+	}
+
+	// Try to parse as URL to distinguish from OCI refs
+	// Normalize common patterns first
+	normalized := source
+	if strings.HasPrefix(source, "github.com/") ||
+		strings.HasPrefix(source, "gitlab.com/") ||
+		strings.HasPrefix(source, "bitbucket.org/") {
+		normalized = "https://" + source
```
craciunoiuc, line 44: Same here, the moment another service is introduced, this breaks

craciunoiuc (changes requested):
I don't quite like this PR, it feels hackish and hardcoded. We should look for something that makes some assumptions generically so it can accept any git alternative to github

PR #2752, opened 2026-03-24
config: fix YAML-tag default resolution and Default() behavior

ZainabTravadi:
## Summary

Fix incorrect default value resolution for config fields defined using YAML tags.

Previously, defaults were not discovered for YAML-tagged fields due to incorrect tag lookup and reflection setup. This PR corrects the resolution logic and adds comprehensive tests to validate behavior and prevent regressions.

---

## Root Cause

The issue was caused by three independent problems:

1. **Incorrect tag lookup**
   - `findConfigDefault()` used `json` tags for field resolution
   - Config structs use `yaml` tags → fields were never matched

2. **Invalid reflection root**
   - `new([0]C)` created a pointer to an array instead of a struct
   - Prevented correct traversal in `reflect.Struct` logic

3. **Incorrect return logic**
   - `Default()` returned incorrect values due to inverted condition

These combined issues caused default values to be silently ignored.

---

## Fix

- Use `getYAMLTag()` for field name resolution
- Initialize reflection with `new(C)` to enable correct struct traversal
- Correct `Default()` return logic to return discovered defaults

---

## Tests

Added comprehensive test coverage:

### Default behavior
- Verify YAML-tagged defaults are resolved correctly
- Ensure empty string is returned when no default is present

### Nested resolution
- Validate recursive lookup for nested fields (e.g. `nested.inner`)

### Regression tests
Cover previously broken behavior:
- YAML vs JSON tag resolution
- Nested struct traversal
- Reflection root correctness
- Missing default handling

---

## Impact

- Fixes broken default resolution for YAML-based config fields
- Prevents silent misconfiguration due to ignored defaults
- Adds regression coverage to ensure long-term correctness

---

## Notes

- Changes are localized to default resolution logic
- No API changes introduced
- All tests pass locally

### DIFF
--- a/config/manager.go
+++ b/config/manager.go
@@ -321,12 +321,12 @@ func FetchConfigDirFromArgs(args []string) (path string) {
 }
 
 func Default[C any](key string) string {
-	found, _, def, _, err := findConfigDefault[C](key, "", "", reflect.ValueOf(new([0]C)))
+	_, found, def, _, err := findConfigDefault[C](key, "", "", reflect.ValueOf(new(C)))
 	if err != nil || found != key {
-		return def
+		return ""
 	}
 
-	return ""
+	return def
 }
 
 func findConfigDefault[C any](needle, offset, def string, v reflect.Value) (string, string, string, reflect.Value, error) {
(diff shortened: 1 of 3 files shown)


ZainabTravadi:
Added nested and regression tests to cover YAML-tag resolution, traversal, and reflection behavior.

Issue #102, opened 2022-09-19
Migrate all errors to constants


nderjung:
This issue tracks migrating all thrown errors from the format:

```go
  ...
  if err != nil {
    return fmt.Errorf("an error occurred")
  }
  ...
```

To:

```go
var (
  ErrAnErrorOccurred = errors.New("an error occurred")
)

  ...
  if err != nil {
    return ErrAnErrorOccurred
  }
  ...
```

rpunjani:
Hello! I'm a student at UT Austin taking a course in virtualization, and I wanted to see if I could contribute to this issue. Is this issue still open, and if so, can I get it assigned to me. Thanks.

nderjung:
Hi @rpunjani, sure!  Please start this iteratively; open PRs which change a single file only -- this makes it easier to review and merge. Thanks!

rpunjani:
@nderjung For clarification, I was wondering if I'm supposed to go through every file in this repository and making the change described above?

nderjung:
This is more-or-less the goal.  Start with errors which are not wrapped (those in the form `fmt.Error("this specific error occured")` are considered not wrapped and `fmt.Errorf("something happened: %w", err)` are considered wrapped) , as these represent final errors.  Here is an example of a recent change that was made which exemplifies this idea: https://github.com/unikraft/kraftkit/commit/9bb236853a83ea5605135461cf8019e3f959ded5.

Go through each file and find errors which are not wrapped and converting these to `var`-typed errors at the top of each file.  Do a commit for each file which is changed as this makes git bisect/blame easier.

Later on we can do better error-catching thanks to these well-defined errors and either remove wrapped errors or convert those into richer, not wrapped, error messages themselves.

Thanks!

PR #2774, opened 2026-04-07
refactor(cloud/scale): Migrate final errors to typed vars

robaa12:
### Prerequisite checklist
### Description of changes
This PR is a scoped, file-by-file contribution toward #102 , limited to the `cloud/scale` command package.
It converts **non-wrapped final errors** to **file-local typed errors** (`var Err... = errors.New(...)`) in the following files:
- `internal/cli/kraft/cloud/scale/add/add.go`
- `internal/cli/kraft/cloud/scale/remove/remove.go`
- `internal/cli/kraft/cloud/scale/get/get.go`
- `internal/cli/kraft/cloud/scale/reset/reset.go`
- `internal/cli/kraft/cloud/scale/initialize/initialize.go`

What was changed:
- Replaced final `fmt.Errorf("...")` returns with typed error vars where applicable.
- Kept wrapped operational/contextual errors (`...: %w`) unchanged.
- In some cases, preserved extra runtime context while wrapping typed errors (for future `errors.Is(...)` checks).

Validation:
- `go test ./internal/cli/kraft/cloud/scale/...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 fmt`
Notes:
- Commits are intentionally split one file at a time to keep review and `git bisect/blame` clean.
- This PR **partially addresses #102**.

### DIFF
--- a/internal/cli/kraft/cloud/scale/add/add.go
+++ b/internal/cli/kraft/cloud/scale/add/add.go
@@ -7,7 +7,6 @@ package add
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sort"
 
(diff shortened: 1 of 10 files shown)


craciunoiuc (changes requested):
These should be moved to a separate file at on the `cloud` directory I think right?

robaa12:
So you mean creating an `errors.go` file at the package level? I actually thought about doing that, but I saw this comment and went with a different approach:
[https://github.com/unikraft/kraftkit/issues/102#issuecomment-2059830689](https://github.com/unikraft/kraftkit/issues/102#issuecomment-2059830689)

Just to confirm I got it right the idea is to define all error constants in `errors.go`, then import and use them across the package?

craciunoiuc:
Yes

Though you won't import them if they are in the same package.

There are two possible options. I'm fine with both:
* You populate the Errors at the top of each file and use them only in there
* You populate them into an `errors.go` in every package

robaa12:
I think the second approach is cleaner. I’ll start adding it to the `cloud` package and go from there.

robaa12:
Hey @craciunoiuc 
I moved typed error declarations out of the command files and into package-level `errors.go` files for each touched `cloud/scale` subpackage:

- `internal/cli/kraft/cloud/scale/add/errors.go`
- `internal/cli/kraft/cloud/scale/remove/errors.go`
- `internal/cli/kraft/cloud/scale/get/errors.go`
- `internal/cli/kraft/cloud/scale/reset/errors.go`
- `internal/cli/kraft/cloud/scale/initialize/errors.go`

I also removed the inline error variable blocks from the corresponding command files. Behavior remains unchanged, including `%w` wrapping and existing context where applicable.

### Follow-up

For follow-up work, I plan to continue this migration in small, package-scoped PRs across the codebase.

craciunoiuc:
This sounds a bit excessive, maybe do a single PR and a commit per command type

otherwise you will go easily into 100 commits+

PR #2803, opened 2026-04-24
fix: Escalate package query error to warning if not found

jedevc:
### Prerequisite checklist

### Description of changes

I had accidentally expired my old github token, which was leading to auth issues, which were getting swallowed in the logs.

So, if we couldn't find *any* matching packages, then we should escalate to a warning. I thought about returning an error, but I don't know if this would break something, and I don't want to accidentally do that at this point.


craciunoiuc (approved):
All good here. Thanks!

PR #2806, opened 2026-04-30
fix(oci): Use cache only when digest matches for blob

craciunoiuc:
### Prerequisite checklist

### Description of changes

Somwhat relates to: TOOL-151

### DIFF
--- a/oci/cache/image.go
+++ b/oci/cache/image.go
@@ -19,33 +19,51 @@ var (
 
 // RemoteImage is a wrapper for v1.Image which caches requests in-memory for
 // previously requested image.  This aids in reducing the number of lookup calls
-// to the same images.  Since v1.WithPlatform is not respected, a valid lookup
-// will have had any additional options, such as WithTransport and WithAuth,
-// fully satisfied.
+// to the same images.  For tag-based references, a lightweight HEAD request is
+// performed to resolve the current digest, which is then used as the cache key.
+// This ensures stale content is never returned when a tag changes, while still
+// benefiting from caching when the tag points to the same digest.  Since
+// v1.WithPlatform is not respected, a valid lookup will have had any additional
+// options, such as WithTransport and WithAuth, fully satisfied.
 func RemoteImage(ref name.Reference, options ...remote.Option) (v1.Image, error) {
-	name := ref.Name()
+	key := ref.String()
+	cacheable := true
 
-	if imageCache == nil {
-		imageCache = make(map[string]v1.Image)
-		goto lookup
+	// For tag-based references (mutable), resolve to the current digest via
+	// a lightweight HEAD request so we can cache by the immutable digest.
+	// If the HEAD request fails we cannot trust any previously cached entry
+	// keyed by the mutable tag, so bypass the cache entirely for this call.
+	if _, isDigest := ref.(name.Digest); !isDigest {
+		desc, err := remote.Head(ref, options...)
+		if err != nil {
+			cacheable = false
+		} else {
+			key = ref.Context().Digest(desc.Digest.String()).String()
+		}
 	}
 
-	imageCacheMu.Lock()
-	if image, ok := imageCache[name]; ok {
+	if cacheable {
+		imageCacheMu.Lock()
+		if imageCache == nil {
+			imageCache = make(map[string]v1.Image)
+		}
+		if image, ok := imageCache[key]; ok {
+			imageCacheMu.Unlock()
+			return image, nil
+		}
 		imageCacheMu.Unlock()
-		return image, nil
 	}
-	imageCacheMu.Unlock()
 
-lookup:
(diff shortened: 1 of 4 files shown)


oci/manager.go
```diff
@@ -481,8 +481,21 @@ func (manager *OCIManager) Catalog(ctx context.Context, qopts ...packmanager.Que
...
+	if isTagRef {
+		log.G(ctx).
+			WithField("ref", ref.Name()).
+			Debug("tag reference detected, querying remote for fresh manifests")
+	}
+
 	// If a direct reference can be made, attempt to generate a package from it.
-	if query.Remote() && refErr == nil && !unsetRegistry {
+	if queryRemote && refErr == nil && !unsetRegistry {
 		authConfig := &authn.AuthConfig{}
```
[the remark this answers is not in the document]
craciunoiuc: Not for this PR

craciunoiuc:
should be good now

jedevc:
So just to check, this should resolve TOOL-151? Just wanted to check, did you manage to reproduce the issue, and make sure it didn't happen?

Similar with the issue the other day here: https://discord.com/channels/879723098881028167/1448722316782735463/1499040432351482067.

craciunoiuc:
Yes, this is related to TOOL-151

Afaik it should solve most if not all kraftkit issues

Because now:
- it checks each blob by sha and uses the local version if the sha matches
- checks that blobs have the size that the manifest says they have (in case someone ctrl^Ced at the bad time)
- If you query by tag it always fetches the index to see if the manifest/config matches. Anything that doesn't match is repulled.
- If you query by digest it should check locally first the index also

Use `--log-level trace --log-type basic` to see what is caached and what is not

oci/handler/directory.go
```diff
@@ -479,6 +479,28 @@ func (handle *DirectoryHandler) PullDigest(ctx context.Context, mediaType, fullr
...
+			info, err := handle.DigestInfo(ctx, layer.Digest)
+			switch {
+			case err == nil && info != nil && info.Size == layer.Size:
+				log.G(ctx).
+					WithField("digest", layer.Digest.String()).
+					Debug("layer already exists locally, skipping pull")
+				continue
+			case err != nil && !os.IsNotExist(err):
+				return fmt.Errorf("could not inspect local layer %s: %w", layer.Digest.String(), err)
+			case err == nil && info != nil && info.Size != layer.Size:
```
jedevc: May also be worth checking the digests? Not just the lens?

Just to confirm, this is an independent check from the changes to caching right? Still seems good to me, just checking my understanding.

oci/handler/directory.go
```diff
@@ -493,6 +515,21 @@ func (handle *DirectoryHandler) PullDigest(ctx context.Context, mediaType, fullr
 		}
 
 	case ocispec.MediaTypeImageLayer, ocispec.MediaTypeImageLayerGzip:
+		// Skip layers that already exist locally.  We do not have an
+		// expected size in this code path, so at minimum require a
+		// non-empty blob and surface unexpected stat errors instead of
+		// silently re-pulling.
+		info, err := handle.DigestInfo(ctx, dgst)
```
jedevc: :eyes: There's still a descriptor. All descriptors have a size + digest, so I think we could still do the same verification.

oci/cache/image.go
```diff
@@ -19,33 +19,51 @@ var (
...
+	// a lightweight HEAD request so we can cache by the immutable digest.
+	// If the HEAD request fails we cannot trust any previously cached entry
+	// keyed by the mutable tag, so bypass the cache entirely for this call.
+	if _, isDigest := ref.(name.Digest); !isDigest {
+		desc, err := remote.Head(ref, options...)
+		if err != nil {
+			cacheable = false
+		} else {
+			key = ref.Context().Digest(desc.Digest.String()).String()
+		}
```
jedevc, line 42: Hm. I'm not super sure about this actually. Does this really get called lots of times in a single process?

This is only an in-memory cache, I suspect this may actually be fine - we don't really expect a tag to mutate *during* a kraftkit run. It's more the case that *between* runs, we're currently not refreshing caches (i.e. the change in `oci/manager.go`).

craciunoiuc:
I lied, it's not related to TOOL-151 :sob:

oci/handler/directory.go
craciunoiuc: Yup, there was no check that the digest was complete after pulling (e.g. network hiccup) so this was just a very simple check.

I think I can compute the actual digests yeah

oci/cache/image.go
craciunoiuc: I think it gets called once on lookup and twice on pulling, which should be fine

we'll see if we get complaints from people with high latency internet :sweat:

jedevc (approved):

PR #2816, opened 2026-05-08
gomod(deps): Move packages to standalone and bump old deps

craciunoiuc:
### Prerequisite checklist

### Description of changes

`exec` and `log` also needed to be made standalone.

This is because `make` uses `exec` for running make commands, and `exec` uses `log` to log to debugging the command that is being executed.

Also did some bumps cause I was getting now some weird compile errors.

Closes: TOOL-849


jedevc:
This makes things a bit trickier - I think we should hold merging this until we decide to actually integrate kernel builds into the new CLI.

PR #2820, opened 2026-05-11
fix(cloud)!: Loop wait for start/stop/rollout start

craciunoiuc:
### Prerequisite checklist

### Description of changes

Previously, on a long start (now with on-demand pulling), there is
a high chance that a start would take more than 10 seconds, which
is the default max wait time in the API.

To circumvent this, wait indefinitely by setting `
-1` for the Wait
API call, and set the deadline through the context.

For sequential rollout, this is for every instance creation.
For on-demand pulling, the time penalty will happen only once.

### DIFF
--- a/internal/cli/kraft/cloud/instance/create/create.go
+++ b/internal/cli/kraft/cloud/instance/create/create.go
@@ -53,7 +53,7 @@ type CreateOptions struct {
 	Replicas            uint                           `local:"true" long:"replicas" short:"R" usage:"Number of replicas of the instance" default:"0"`
 	Rollout             *RolloutStrategy               `noattribute:"true"`
 	RolloutQualifier    *RolloutQualifier              `noattribute:"true"`
-	RolloutWait         time.Duration                  `local:"true" long:"rollout-wait" usage:"Time to wait before performing rolling out action (ms/s/m/h)" default:"10s"`
+	RolloutWait         time.Duration                  `local:"true" long:"rollout-wait" usage:"Time to wait before performing rolling out action (ms/s/m/h)" default:"5m"`
 	ServiceNameOrUUID   string                         `local:"true" long:"service" short:"g" usage:"Attach this instance to an existing service"`
 	Start               bool                           `local:"true" long:"start" short:"S" usage:"Immediately start the instance after creation"`
 	ScaleToZero         *kcinstances.ScaleToZeroPolicy `noattribute:"true"`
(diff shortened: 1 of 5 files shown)


internal/cli/kraft/cloud/instance/create/create.go
```diff
@@ -628,8 +628,16 @@ func Create(ctx context.Context, opts *CreateOptions, args ...string) (*kcclient
 				"waiting for new instance to start before performing rollout action",
 				"",
 				func(ctx context.Context) error {
-					_, err := opts.Client.Instances().WithMetro(opts.Metro).Wait(ctx, kcinstances.StateRunning, int(opts.RolloutWait.Milliseconds()), newInstance.UUID)
-					return err
+					deadline := time.Now().Add(opts.RolloutWait)
+					for {
+						_, err := opts.Client.Instances().WithMetro(opts.Metro).Wait(ctx, kcinstances.StateRunning, int(10*time.Second.Milliseconds()), newInstance.UUID)
```
jedevc: 🤔 what happens if we get to a stopped state? does wait return?

craciunoiuc: No, it waits for it to reach the state it's waiting for, even if it's for a brief moment

In this case it should be fine I think? Only case it fails if it goes from `starting` to `stopped` directly, meaning crashed at boot time I think?

jedevc: Hm, so if it crashed, we're just gonna end up looping and timing out right?

Maybe we should add a check at the end of the loop to confirm if it's still in "starting", and only keep looping if it's there.

craciunoiuc: Good idea. We waited for 20s, what's ~100ms more 😅

craciunoiuc: boom

internal/cli/kraft/cloud/instance/create/create.go
```diff
@@ -628,8 +628,24 @@ func Create(ctx context.Context, opts *CreateOptions, args ...string) (*kcclient
...
+						if err == nil {
+							return nil
+						}
+						if time.Now().After(deadline) {
+							return err
+						}
+						getResp, getErr := opts.Client.Instances().WithMetro(opts.Metro).Get(ctx, newInstance.UUID)
+						if getErr == nil {
+							if inst, getErr := getResp.FirstOrErr(); getErr == nil {
+								if inst.State != kcinstances.InstanceStateStarting && inst.State != kcinstances.InstanceStateRunning {
```
jedevc, line 643: If it's starting, we can just exit the loop right?

craciunoiuc: Not necessarily, starting can be an on-demand pull which failed (for example)

Or maybe a big to firecracker making VMs be blocked in booting (happened before)

TLDR: normally yes, but if it something fails, not necessarily

craciunoiuc:
I knew about all 3, but it didn't seem that bad to overshoot

Normally, overshooting isn't bad in this case, but I can fix, to make it not lie 😅

internal/cli/kraft/cloud/instance/create/create.go
jedevc: Oof, I needed coffee. I meant if it's *running* we can exit?

craciunoiuc: well if it's running it will exit beforehand in the `break`. And if somehow it gets into running between the two, it will exit afterwards in the next `wait` call so all good

nderjung:
> If the timeout is -1, the request will block indefinitely until the instances reach the desired states

https://unikraft.com/docs/api/platform/v1/instances#wait-for-instances-states

craciunoiuc:
We decided on something else here

PR #2823, opened 2026-05-13
feat: Hyperlight platform support

ijaidev:
### Prerequisite checklist

### Description of changes

This PR adds local Hyperlight platform support to KraftKit.

It introduces a `machine/hyperlight` driver that runs Unikraft unikernels with
the external `hyperlight-unikraft` host binary. The driver registers both the
canonical `hyperlight` platform name and the `hl` shorthand, prepares machine
state during create, starts the Hyperlight host process as a detached child,
tracks its PID and log file, and supports the standard local lifecycle surface
for `kraft ps`, `kraft logs`, `kraft stop`, and `kraft rm`.

The branch also adds Hyperlight-specific run configuration and CLI flags:

- `--hyperlight-stack`
- `--hyperlight-quiet`
- `--hyperlight-enable-tools`
- `--hyperlight-repeat`
- `--hyperlight-mount`
- `--hyperlight-exec`

Supported KraftKit run inputs are mapped to Hyperlight where possible:

- `--memory` maps to the Hyperlight memory setting.
- `--rootfs` and `--initrd` are passed as the guest initrd.
- Writable `9pfs` directory volumes are passed as Hyperlight mounts.
- Application arguments after `--` are forwarded to the guest.

Unsupported KraftKit surfaces are rejected by the Hyperlight driver with clear
errors, including port publishing, network attachments, emulation mode, runtime
environment injection, kernel arguments, read-only volumes, and non-`9pfs`
volumes. This keeps the generic `kraft run` flow platform-agnostic while
letting the machine driver enforce what Hyperlight can actually execute.

Additional fixes in this branch:

- Accept platform aliases in machine commands so `hl` works consistently.
- Build package rootfs artefacts as CPIO by default for Hyperlight runs.
- Avoid hanging log follow on already-exited Hyperlight machines.
- Reap short-lived child processes and detect zombie processes as exited.
- Return delete cleanup errors from the Hyperlight driver.
- Preserve started machine state for cleanup after `kraft start`.
- Propagate `kraft pause` failures for unsupported Hyperlight pause requests.
- Honor `kraft build --no-rootfs` by skipping rootfs generation.
- Document the Hyperlight requirements, supported flags, limitations, and the
  upstream `hyperlight-dev/hyperlight-unikraft` binary source.

### Testing

Local testing covered the Hyperlight driver and the related CLI paths:

Manual end-to-end testing was also performed with the installed `kraft` binary.

- Built the example for `--plat hyperlight --arch x86_64`.
- Reused an explicit `--rootfs` path during run commands to avoid repeatedly
  rebuilding the Docker rootfs.
- Exercised create/start/run/logs/ps/stop/rm behavior.
- Verified non-detached Hyperlight run exits cleanly.
- Verified unsupported pause returns an error.
- Verified supported Hyperlight flags are accepted.
- Verified unsupported flags such as networking, ports, env injection,
  emulation, kernel args, read-only volumes, and non-`9pfs` volumes return
  driver-level errors.
- Verified `--network kraft0` only proceeds after the user-created network is
  found, then Hyperlight rejects the network attachment as unsupported.
- Verified `--no-rootfs` skips rootfs generation.
- Checked for leftover `hyperlight-unikraft` processes after lifecycle tests.

> Note: This PR was assisted by an LLM under highly specific instructions and strict review.

### DIFF
--- a/internal/cli/kraft/events/events.go
+++ b/internal/cli/kraft/events/events.go
@@ -57,7 +57,7 @@ func NewCmd() *cobra.Command {
 
 	cmd.Flags().VarP(
 		cmdfactory.NewEnumFlag[mplatform.Platform](
-			mplatform.Platforms(),
+			mplatform.PlatformNames(mplatform.Platform("auto")),
 			mplatform.Platform("auto"),
 		),
 		"plat",
--- a/machine/hyperlight/README.md
+++ b/machine/hyperlight/README.md
@@ -42,7 +42,8 @@ kraft run   --plat hl ...
 
 ### Defaults
 
-- `DefaultMemory`: `16Mi`. Override with `--memory` for heavier guests.
+- `DefaultMemory`: `32Mi`, matching the `hyperlight-unikraft` default.
+  Override with `--memory` for guests that need a different allocation.
 - `DefaultStack`: `8Mi`. Not yet exposed through the CLI or Kraftfile.
 
 ### Limitations
--- a/machine/hyperlight/README.md
+++ b/machine/hyperlight/README.md
@@ -11,11 +11,11 @@ across separate kraft invocations via standard PID tracking.
 - Linux host with `/dev/kvm` read/write access, or Windows host with the
   Windows Hypervisor Platform (WHP) enabled.
 - The `hyperlight-unikraft` binary (from
-  [danbugs/hyperlight-unikraft](https://github.com/danbugs/hyperlight-unikraft))
+  [hyperlight-dev/hyperlight-unikraft](https://github.com/hyperlight-dev/hyperlight-unikraft))
   installed on `$PATH`:
 
   ```bash
-  cargo install --git https://github.com/danbugs/hyperlight-unikraft \
+  cargo install --git https://github.com/hyperlight-dev/hyperlight-unikraft \
       --branch main hyperlight-unikraft-host --bin hyperlight-unikraft
   ```
 
@@ -67,18 +67,19 @@ Hyperlight-specific host options are exposed with a `--hyperlight-*` prefix on
 - `--hyperlight-mount`
 - `--hyperlight-exec`
 
-Dockerfile and OCI rootfs metadata can contain default environment variables.
 `hyperlight-unikraft` does not currently expose a runtime environment-injection
-interface. If the guest needs environment variables, compile them into the
-unikernel configuration or application.
+interface. Avoid runtime environment metadata for Hyperlight guests, including
+Dockerfile/OCI env metadata, `--env`, and Kraftfile `env:` entries; compile
+required values into the unikernel configuration or application.
 
 ### Limitations
 
 - `kraft pause` is not supported; Hyperlight has no pause semantics.
 - The child process is terminated on `kraft stop` via SIGTERM; there is no
   in-VM quiesce step.
-- Runtime environment injection is not supported. Explicit `kraft run --env`
-  and Kraftfile `env:` entries fail with a clear error.
+- Runtime environment injection is not supported. Explicit `kraft run --env`,
+  Kraftfile `env:` entries, and rootfs metadata environment entries fail with
+  a clear error.
 - Network attachments, port publishing, emulation mode, and kernel arguments
   are rejected because `hyperlight-unikraft` does not support those KraftKit
   interfaces.
--- a/machine/hyperlight/README.md
+++ b/machine/hyperlight/README.md
@@ -56,6 +56,12 @@ The driver maps the common KraftKit run surface that Hyperlight can execute:
 - Application arguments after `--` are passed after the kernel path.
 - Writable `9pfs` directory volumes are passed as repeatable
   `hyperlight-unikraft --mount HOST:GUEST` entries.
+- Same-port guest listen permissions from `--port GUEST_PORT:GUEST_PORT` are
+  passed as `hyperlight-unikraft --net --port GUEST_PORT`. Hyperlight ports are
+  sandbox bind permissions, not Docker-style host forwarding.
+- Hyperlight sandbox networking policy is configured with
+  `--hyperlight-net-allow` and `--hyperlight-net-block`; allow/block entries
+  imply networking in `hyperlight-unikraft`.
 
 Hyperlight-specific host options are exposed with a `--hyperlight-*` prefix on
 `kraft run` and are ignored by other platform drivers:
@@ -63,6 +69,8 @@ Hyperlight-specific host options are exposed with a `--hyperlight-*` prefix on
 - `--hyperlight-stack`
 - `--hyperlight-quiet`
 - `--hyperlight-enable-tools`
+- `--hyperlight-net-allow`
+- `--hyperlight-net-block`
 - `--hyperlight-repeat`
 - `--hyperlight-mount`
 - `--hyperlight-exec`
@@ -80,8 +88,10 @@ required values into the unikernel configuration or application.
 - Runtime environment injection is not supported. Explicit `kraft run --env`,
   Kraftfile `env:` entries, and rootfs metadata environment entries fail with
   a clear error.
-- Network attachments, port publishing, emulation mode, and kernel arguments
-  are rejected because `hyperlight-unikraft` does not support those KraftKit
-  interfaces.
+- Network attachments (`--network`, `--ip`, `--mac`), host port forwarding,
+  emulation mode, and kernel arguments are rejected because
+  `hyperlight-unikraft` does not support those KraftKit interfaces.
+- `--hyperlight-net-allow` and `--hyperlight-net-block` are mutually exclusive
+  sandbox policies.
 - Read-only volumes and non-`9pfs` volume drivers are rejected. Rootfs/initrd
   volumes are supported only when they represent the main initrd at `/`.
--- a/machine/hyperlight/README.md
+++ b/machine/hyperlight/README.md
@@ -44,10 +44,43 @@ kraft run   --plat hl ...
 
 - `DefaultMemory`: `32Mi`, matching the `hyperlight-unikraft` default.
   Override with `--memory` for guests that need a different allocation.
-- `DefaultStack`: `8Mi`. Not yet exposed through the CLI or Kraftfile.
+- `DefaultStack`: `8Mi`. Override with `--hyperlight-stack`.
+
+### Supported run options
+
+The driver maps the common KraftKit run surface that Hyperlight can execute:
+
+- `--memory` is passed to `hyperlight-unikraft --memory`.
+- `--rootfs`/`--initrd` CPIO archives are passed as
+  `hyperlight-unikraft --initrd`.
+- Application arguments after `--` are passed after the kernel path.
+- Writable `9pfs` directory volumes are passed as repeatable
+  `hyperlight-unikraft --mount HOST:GUEST` entries.
+
+Hyperlight-specific host options are exposed with a `--hyperlight-*` prefix on
+`kraft run` and are ignored by other platform drivers:
+
+- `--hyperlight-stack`
+- `--hyperlight-quiet`
+- `--hyperlight-enable-tools`
+- `--hyperlight-repeat`
+- `--hyperlight-mount`
+- `--hyperlight-exec`
+
+Dockerfile and OCI rootfs metadata can contain default environment variables.
+`hyperlight-unikraft` does not currently expose a runtime environment-injection
+interface. If the guest needs environment variables, compile them into the
+unikernel configuration or application.
 
 ### Limitations
 
 - `kraft pause` is not supported; Hyperlight has no pause semantics.
 - The child process is terminated on `kraft stop` via SIGTERM; there is no
   in-VM quiesce step.
+- Runtime environment injection is not supported. Explicit `kraft run --env`
+  and Kraftfile `env:` entries fail with a clear error.
+- Network attachments, port publishing, emulation mode, and kernel arguments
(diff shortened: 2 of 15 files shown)


craciunoiuc:
@danbugs this is the branch we were talking about. If you have mentions you can leave them here also

I will also go over this, but maybe in 1-2 weeks

machine/hyperlight/v1alpha1.go
```diff
@@ -31,9 +31,9 @@ import (
 
 const (
 	// DefaultMemory is the default memory allocation for Hyperlight VMs when
-	// --memory is not supplied. Small guests fit comfortably; heavier
-	// interpreters need an explicit override upward.
-	DefaultMemory = "16Mi"
+	// --memory is not supplied. This matches hyperlight-unikraft's host-side
+	// default.
+	DefaultMemory = "512Mi"
```
craciunoiuc: Isn't this a bit big? Maybe 64MB or 128MB is better?

machine/hyperlight/v1alpha1.go
```diff
@@ -76,8 +76,28 @@ func (service *machineV1alpha1Service) Create(ctx context.Context, machine *mach
...
+	if len(machine.Spec.Networks) > 0 {
+		return machine, fmt.Errorf("hyperlight-unikraft does not support KraftKit network attachments (--network, --ip, --mac)")
+	}
+
+	if len(machine.Spec.Env) > 0 {
+		return machine, fmt.Errorf("hyperlight-unikraft does not support KraftKit environment injection (--env)")
+	}
+
+	if len(machine.Spec.KernelArgs) > 0 {
+		return machine, fmt.Errorf("hyperlight-unikraft does not support KraftKit kernel arguments (--kernel-arg); pass application arguments after --")
```
craciunoiuc: No need to specify `KraftKit` every time. They are already using KraftKit so they know :D

machine/hyperlight/v1alpha1.go
```diff
@@ -472,3 +491,77 @@ func validateExistingFile(path, description string) error {
...
+		return nil, nil
+	}
+
+	mounts := make([]string, 0, len(machine.Spec.Volumes))
+	guestPaths := map[string]struct{}{}
+
+	for _, volume := range machine.Spec.Volumes {
+		switch volume.Spec.Driver {
+		case "initrd":
+			if volume.Spec.Destination == "/" && machine.Status.InitrdPath != "" {
```
craciunoiuc, line 670: The initrd afaik is always mounted at root. Is this a hyperlight quirk?

machine/hyperlight/v1alpha1.go
```diff
@@ -472,3 +491,77 @@ func validateExistingFile(path, description string) error {
...
+		}
+
+		if volume.Spec.ReadOnly {
+			return nil, fmt.Errorf("hyperlight-unikraft does not support read-only KraftKit volumes")
+		}
+
+		if volume.Spec.Source == "" {
+			return nil, fmt.Errorf("hyperlight volume source cannot be empty")
+		}
+
```
craciunoiuc, line 690: This is quite a substantial commit of things that work/don't

@danbugs would you be able in the more distant future to keep this list of parameters that work/don't up to date?

machine/hyperlight/v1alpha1.go
```diff
@@ -472,3 +491,77 @@ func validateExistingFile(path, description string) error {
...
+		guestPaths[guestPath] = struct{}{}
+
+		mounts = append(mounts, fmt.Sprintf("%s:%s", hostPath, guestPath))
+	}
+
+	return mounts, nil
+}
+
+func normalizeHyperlightGuestMountPath(guestPath string) (string, error) {
+	if guestPath == "" {
```
craciunoiuc, line 782: Name is a bit wrong? No normalization happens here, only some checks

machine/hyperlight/v1alpha1.go
```diff
@@ -461,6 +457,46 @@ func getHyperlightConfigFromPlatformConfig(platformConfig interface{}) (*Hyperli
 	return &hlcfg, nil
 }
 
+func buildHyperlightArgs(hlcfg *HyperlightConfig, appArgs []string) []string {
+	args := []string{
+		"--memory", hlcfg.Memory,
+		"--stack", hlcfg.Stack,
+	}
```
craciunoiuc: Maybe for the `hlcfg` struct you could add a `MarshalArgs` function? So basically move this function under the struct and rename?

machine/hyperlight/v1alpha1.go
```diff
@@ -563,5 +599,19 @@ func normalizeHyperlightGuestMountPath(guestPath string) (string, error) {
...
+	cleaned := pathpkg.Clean(guestPath)
+	if _, reserved := reservedGuestMountPaths[cleaned]; reserved {
+		return "", fmt.Errorf("hyperlight volume destination %q shadows a reserved guest directory", guestPath)
+	}
+
+	for reserved := range reservedGuestMountPaths {
+		if reserved == "/" {
+			continue
+		}
+		if strings.HasPrefix(cleaned, reserved+"/") {
```
craciunoiuc: Maybe `filepath` can help with the file path operations in this PR?

internal/cli/kraft/events/events.go
```diff
@@ -57,7 +57,7 @@ func NewCmd() *cobra.Command {
 
 	cmd.Flags().VarP(
 		cmdfactory.NewEnumFlag[mplatform.Platform](
-			mplatform.Platforms(),
+			mplatform.PlatformNames(mplatform.Platform("auto")),
 			mplatform.Platform("auto"),
```
craciunoiuc, line 61: This looks like the fix should be somewhere else? Doesn't `Platforms()` list the correct platform names?

machine/hyperlight/v1alpha1.go
```diff
@@ -245,6 +270,10 @@ func (service *machineV1alpha1Service) Start(ctx context.Context, machine *machi
 		machine.Status.State = machinev1alpha1.MachineStateFailed
 		return machine, err
 	}
+	if hlcfg.Exec != "" && len(machine.Spec.ApplicationArgs) > 0 {
+		machine.Status.State = machinev1alpha1.MachineStateFailed
+		return machine, fmt.Errorf("hyperlight-unikraft --exec conflicts with application arguments passed after --")
```
craciunoiuc: This could be more explicit like "you can't set both at the same time"

internal/cli/kraft/run/utils.go
```diff
@@ -354,9 +354,12 @@ func (opts *RunOptions) prepareRootfs(ctx context.Context, machine *machineapi.M
 	// If the user has supplied an initram path, set this now, this overrides any
 	// preparation and is considered higher priority compared to what has been set
 	// prior to this point.
-	if opts.Rootfs == "" || machine.Status.InitrdPath != "" || opts.RootfsType == "" {
+	if opts.Rootfs == "" || machine.Status.InitrdPath != "" {
 		return nil
 	}
+	if opts.RootfsType == "" {
+		opts.RootfsType = initrd.FsTypeCpio
+	}
```
craciunoiuc, line 351: Does hyperlight support erofs anyway?

machine/hyperlight/README.md
```diff
@@ -44,10 +44,43 @@ kraft run   --plat hl ...
 
 - `DefaultMemory`: `512Mi`, matching the `hyperlight-unikraft` default.
   Override with `--memory` for guests that need a different allocation.
-- `DefaultStack`: `8Mi`. Not yet exposed through the CLI or Kraftfile.
+- `DefaultStack`: `8Mi`. Override with `--hyperlight-stack`.
+
```
craciunoiuc, line 48: You updated this, change here also

craciunoiuc (changes requested):
Note, I reviewed by-commit, so some things are out of date because they changed from one commit to another.

`fix(machine/hyperlight): Handle rootfs and log edge cases` - this commit has lines too long in the body.
Also there are some strange spaces between each sentence, not sure if intentional.

machine/hyperlight/v1alpha1.go
danbugs: Heads up, changed the default to 32MB over at hyperlight-unikraft: https://github.com/hyperlight-dev/hyperlight-unikraft/pull/77. I think that should be sufficient for most workloads, so we can probably match it here.

machine/hyperlight/v1alpha1.go
danbugs: Yeah, happy to keep that up to date :). Worth noting that networking (`--net`, `--net-allow`, `--net-block`) and port binding (`--port`) are already supported in hyperlight-unikraft but this PR rejects them.

machine/hyperlight/v1alpha1.go
danbugs: Right--it is always at root, same for HL. I think this just skips it because it's covered by `--initrd`?

internal/cli/kraft/run/utils.go
danbugs: HL mostly uses cpiovfs. e.g., https://github.com/unikraft/unikraft/pull/1822

machine/hyperlight/v1alpha1.go
craciunoiuc: @ijaidev can you investigate them also :D

internal/cli/kraft/run/utils.go
craciunoiuc: Okay, so maybe a good idea to reject everything else for now

machine/hyperlight/v1alpha1.go
ijaidev: You're right, this was to match the default

machine/hyperlight/v1alpha1.go
ijaidev: will remove :))

machine/hyperlight/v1alpha1.go
ijaidev: image
This is adding an initrd to volumes, so it was added to check for unnecessary errors. what do you think?

machine/hyperlight/v1alpha1.go
ijaidev: image
my binary is not up to date ig, I'll handle them.

machine/hyperlight/v1alpha1.go
ijaidev: Image
small cleaning ;) And it is returning the path back, so what else could we name?

machine/hyperlight/README.md
ijaidev: okay

machine/hyperlight/v1alpha1.go
ijaidev: okay

internal/cli/kraft/events/events.go
ijaidev: It doesn't list aliasis, so user can't use both `--plat hl or hyperlight`

internal/cli/kraft/run/utils.go
ijaidev: Am I supposed to do anything here? `machine` has only the rootfs string path, not type.

machine/hyperlight/v1alpha1.go
craciunoiuc: This actually happened because the other change was in a subsequent commit, feel free to ignore

internal/cli/kraft/run/utils.go
craciunoiuc: No need to do anything I believe, all good

machine/hyperlight/v1alpha1.go
ijaidev: done

machine/hyperlight/v1alpha1.go
```diff
@@ -408,3 +490,255 @@ func getHyperlightConfigFromPlatformConfig(platformConfig interface{}) (*Hyperli
...
+
+	hostPath, guestPath, hasGuestPath := strings.Cut(spec, ":")
+	if hostPath == "" {
+		return "", "", fmt.Errorf("hyperlight mount source cannot be empty")
+	}
+	if !hasGuestPath {
+		guestPath = "/host"
+	} else if guestPath == "" {
+		return "", "", fmt.Errorf("hyperlight mount destination cannot be empty")
+	}
```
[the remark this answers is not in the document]
ijaidev: @craciunoiuc what do you think about this? As of now, Krafkit don't support Windows.

machine/hyperlight/v1alpha1.go
```diff
@@ -298,13 +372,14 @@ func (service *machineV1alpha1Service) Stop(ctx context.Context, machine *machin
...
+	machine, stopErr := service.Stop(ctx, machine)
 
+	var removeErr error
 	if machine.Status.StateDir != "" {
-		os.RemoveAll(machine.Status.StateDir)
+		removeErr = os.RemoveAll(machine.Status.StateDir)
 	}
 
-	return machine, nil
+	return nil, errors.Join(stopErr, removeErr)
```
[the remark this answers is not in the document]
ijaidev: it follows the pattern of qemu and fc

machine/hyperlight/v1alpha1.go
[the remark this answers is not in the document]
craciunoiuc: Ignore Windows for now (sorry @danbugs :( )

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #2834, opened 2026-05-25
fix(run): Default rootfs type to cpio when empty

### LINKED ISSUE #2833: When RootfsType is empty, the rootfs preparation would return early and fail in `run` command.
### Describe the bug

When RootfsType is empty (not passed by cmd or by Kraftfile) in the `run` command, the rootfs preparation would return early and fail to prepare the initramfs.

### Steps to reproduce

_No response_

### Expected behavior

It should add a default type and proceed rather than exiting early.

### KraftKit version

_No response_

_No response_

_No response_

### Relevant log output

```shell

```

ijaidev:
When Rootfs is specified, but RootfsType is empty, the rootfs preparation would return early and fail to prepare the initramfs. Now, if RootfsType is empty, it defaults to CPIO format instead of returning early.

Closes: #2833


craciunoiuc (approved):
All good here. Thanks!

```
```

PR #2873, opened 2026-07-07
fix(run): Ensure absolute path for workdir and validate existing rootfs

ijaidev:
# Run: Ensure Absolute Path for Workdir and Validate Existing Rootfs

## Overview
This PR introduces path resolution improvements for the `kraft run` command, ensuring that the working directory (`workdir`) is resolved to an absolute path, and optimizes the rootfs preparation phase by skipping unnecessary builds when an already-built rootfs is present and valid.

### DIFF
--- a/internal/cli/kraft/run/runner_kernel.go
+++ b/internal/cli/kraft/run/runner_kernel.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"debug/elf"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	machineapi "kraftkit.sh/api/machine/v1alpha1"
@@ -53,6 +54,12 @@ func (runner *runnerKernel) Runnable(ctx context.Context, opts *RunOptions, args
 
 // Prepare implements Runner.
 func (runner *runnerKernel) Prepare(ctx context.Context, opts *RunOptions, machine *machineapi.Machine, args ...string) error {
+	var err error
+	opts.workdir, err = os.Getwd()
+	if err != nil {
+		return fmt.Errorf("could not get working directory: %w", err)
+	}
+
 	filename := filepath.Base(runner.kernelPath)
 	machine.Spec.Kernel = "kernel://" + filename
 	machine.Status.KernelPath = runner.kernelPath
--- a/internal/cli/kraft/run/runner_kraftfile_unikraft.go
+++ b/internal/cli/kraft/run/runner_kraftfile_unikraft.go
@@ -64,10 +64,14 @@ func (runner *runnerKraftfileUnikraft) Runnable(ctx context.Context, opts *RunOp
 		runner.workdir = cwd
 		runner.args = args
 		if f, err := os.Stat(args[0]); err == nil && f.IsDir() {
-			runner.workdir = args[0]
+			runner.workdir, err = filepath.Abs(args[0])
+			if err != nil {
+				return false, fmt.Errorf("getting absolute path of workdir: %w", err)
+			}
 			runner.args = args[1:]
 		}
 	}
+	opts.workdir = runner.workdir
 
 	popts := []app.ProjectOption{
 		app.WithProjectWorkdir(runner.workdir),
--- a/internal/cli/kraft/run/utils.go
+++ b/internal/cli/kraft/run/utils.go
@@ -356,6 +356,18 @@ func (opts *RunOptions) prepareRootfs(ctx context.Context, machine *machineapi.M
 		fmt.Sprintf(initrd.DefaultInitramfsArchFileName, machine.Spec.Architecture, opts.RootfsType),
 	)
 
+	if machine.Status.InitrdPath == opts.Rootfs {
+		stat, err := os.Stat(opts.Rootfs)
+		if err != nil {
+			return fmt.Errorf("using existing rootfs: %w", err)
+		}
+
+		if !stat.Mode().IsRegular() {
+			return fmt.Errorf("using existing rootfs: %s is not a regular file", opts.Rootfs)
+		}
+		return nil
+	}
+
 	ramfs, err := initrd.New(ctx,
 		opts.Rootfs,
 		initrd.WithOutput(machine.Status.InitrdPath),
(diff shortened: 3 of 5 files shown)


internal/cli/kraft/run/run.go
```diff
@@ -161,6 +162,11 @@ func NewCmd() *cobra.Command {
 func (opts *RunOptions) Pre(cmd *cobra.Command, _ []string) error {
 	ctx := cmd.Context()
 
+	var err error
+	opts.workdir, err = os.Getwd()
+	if err != nil {
```
craciunoiuc: is there no variable in `run` for setting this? just checking

craciunoiuc: ah I see now you changed it below, good

internal/cli/kraft/run/runner_kraftfile_unikraft.go
```diff
@@ -64,10 +64,14 @@ func (runner *runnerKraftfileUnikraft) Runnable(ctx context.Context, opts *RunOp
...
 		if f, err := os.Stat(args[0]); err == nil && f.IsDir() {
-			runner.workdir = args[0]
+			runner.workdir, err = filepath.Abs(args[0])
+			if err != nil {
+				return false, fmt.Errorf("getting absolute path of workdir: %w", err)
+			}
 			runner.args = args[1:]
 		}
 	}
+	opts.workdir = runner.workdir
```
craciunoiuc, line 74: maybe set this only if `runner.workdir` is not empty?

internal/cli/kraft/run/utils.go
```diff
@@ -356,6 +356,18 @@ func (opts *RunOptions) prepareRootfs(ctx context.Context, machine *machineapi.M
...
 
+	if machine.Status.InitrdPath == opts.Rootfs {
+		stat, err := os.Stat(opts.Rootfs)
+		if err != nil {
+			return fmt.Errorf("using existing rootfs: %w", err)
+		}
+
+		if !stat.Mode().IsRegular() {
+			return fmt.Errorf("using existing rootfs: %s is not a regular file", opts.Rootfs)
+		}
```
craciunoiuc, line 367: Is there any case where this could happen? Aka where `InitrdPath` would not point to a file?

craciunoiuc (commented):
some comments my side

internal/cli/kraft/run/runner_kraftfile_unikraft.go
ijaidev: in `run` command, every runner has its separate dirs, or it's just isolated, so I set this as the default in `run.go` because some runners don't set this. So we let it override?

---- 
As every runner is different or isolated, we can move the rootfs build and some other things to the runners to prevent rebuilds and some other stuff.

internal/cli/kraft/run/utils.go
ijaidev: just a quick check if something is wrong with rootfs, to return quick ?

craciunoiuc:
have a free AI review on the house:

| # | Severity | File | Issue | Type |
| --- | --- | --- | --- | --- |
| 1 | 🔴 High | runtime/unikraft runners, run.go | Runnable mutates shared opts.workdir for every probed candidate; last/selected mismatch | Bug (state) |
| 2 | 🔴 High | runner_kraftfile_unikraft.go:74 | New opts.workdir = runner.workdir promotes private state into probe-time shared mutation, asymmetric vs runtime runner | Design/consistency |
| 3 | 🟠 Medium | run.go:165 | Pre sets workdir only on CLI path; run.Run() programmatic entrypoint leaves it empty for kernel/linuxu/package runners | Bug (edge) |
| 4 | 🟠 Medium | utils.go:359 | Rootfs skip uses raw ==; breaks under symlinks/uncleaned paths or non-CLI callers | Robustness |
| 5 | 🟠 Medium | utils.go:368 | Early return nil skips ramfs.Env()/Args() population — behavior change for pre-built rootfs | Behavior change |
| 6 | 🟡 Low | multiple | Inconsistent error-message style across new code | Style |
| 7 | 🟡 Low | utils.go:362 | os.Stat errors all bucketed as "using existing rootfs" | Clarity |
| 8 | 🟡 Low | — | No tests for new logic | Coverage |
| 9 | 🟢 Nit | runtime/unikraft runners | Duplicate os.Getwd() calls (Pre + Runnable) | Minor |

internal/cli/kraft/run/runner_kraftfile_unikraft.go
craciunoiuc: Idk, feels weird, see also the review below

ijaidev: this was to ensure that `opts.workdir` exists, cause some runners don't set it like the `kernel` runner. How about I move this logic to those runners? so it will be isolated like the others.

ijaidev: This will solve most of the doubts

ijaidev:
update the PR, moved the workdir logic to individual runners

internal/cli/kraft/run/runner_kernel.go
```diff
@@ -53,6 +54,12 @@ func (runner *runnerKernel) Runnable(ctx context.Context, opts *RunOptions, args
 
 // Prepare implements Runner.
 func (runner *runnerKernel) Prepare(ctx context.Context, opts *RunOptions, machine *machineapi.Machine, args ...string) error {
+	var err error
+	opts.workdir, err = os.Getwd()
```
craciunoiuc, line 58: just one final confirmation

This will get replaced with the cli argument if it's set right?

ijaidev: in this case (or in updated runners), user don't pass workdir from cli. In this case the first arg will be kernel path.

ijaidev: Workdir didn't exist in these runners before. This is default for artifact related works.

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #2879, opened 2026-07-15
fix: Don't override runtime flag version with kraftfile

craciunoiuc:
### Prerequisite checklist

### DIFF
--- a/internal/cli/kraft/pkg/packager_kraftfile_runtime.go
+++ b/internal/cli/kraft/pkg/packager_kraftfile_runtime.go
@@ -63,6 +63,34 @@ func formatRuntimeReference(name, version string) string {
 	return fmt.Sprintf("%s:%s", name, version)
 }
 
+func normalizeRuntimeNameVersion(name, version string) (string, string) {
+	if version == "" && !ukruntime.HasExplicitTag(name) {
+		version = "latest"
+	}
+
+	return name, version
+}
+
+func resolveRuntimeNameVersion(runtimeFlag string, projectRuntime *ukruntime.Runtime, nameFlag string) (string, string, error) {
+	switch {
+	case len(runtimeFlag) > 0:
+		runtime := &ukruntime.Runtime{}
+		runtime.SetName(runtimeFlag)
+		name, version := normalizeRuntimeNameVersion(runtime.Name(), runtime.Version())
+		return name, version, nil
+	case projectRuntime != nil:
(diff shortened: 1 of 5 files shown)


internal/cli/kraft/pkg/packager_kraftfile_runtime.go
```diff
@@ -86,26 +86,45 @@ func (p *packagerKraftfileRuntime) Packagable(ctx context.Context, opts *PkgOpti
...
+		runtime := &ukruntime.Runtime{}
+		runtime.SetName(nameFlag)
+		name, version = runtime.Name(), runtime.Version()
+	} else {
+		return "", "", fmt.Errorf("no runtime name specified")
+	}
+
+	if runtimeFlag == "" && version == "" && projectRuntime != nil {
+		version = projectRuntime.Version()
+	}
```
ajpetersons: why isn't this done above (after L97)? it seems that both these conditions will happen at the same time

internal/cli/kraft/pkg/packager_kraftfile_runtime_test.go
```diff
@@ -27,3 +37,83 @@ func TestFormatRuntimeReference(t *testing.T) {
...
+			runtimeFlag:    "base",
+			projectRuntime: runtimeFromReference("other:v1"),
+			wantName:       "base",
+			wantVersion:    "latest",
+		},
+		{
+			name:           "runtime with registry port and tag",
+			runtimeFlag:    "localhost:5000/acme/base:v2",
+			projectRuntime: runtimeFromReference("other:v1"),
+			wantName:       "localhost:5000/acme/base:v2",
```
ajpetersons, line 70: is this correct? version can be empty? 

what would happen if we used `localhost:5000/acme/base` as the runtime flag, would version be correctly computed?

craciunoiuc: Yes, or basically it depends on the registry

But for all registries I know, no tag defaults to `:latest`

internal/cli/kraft/pkg/packager_kraftfile_runtime.go
craciunoiuc: Sure, makes sense

ajpetersons (approved):
Looks good to me, thanks!

Reviewed-by: Arturs Janis Petersons <arturs@unikraft.io>
Approved-by: Arturs Janis Petersons <arturs@unikraft.io>

PR #2881, opened 2026-07-15
feat(tools/github-action): Update action with the latest changes

craciunoiuc:
### Prerequisite checklist

### Description of changes

Periodic bump to bring in internal changes.

Image already rebuilt and confimed it works.

Closes: QA-227


jedevc (commented):
Yeah this isn't a huge terrifying change at all 😓

Where do all the changes come from? Like `CONFIG_LIBVFSCORE_AUTOMOUNT_EINITRD_PATH`, and everything in that commit.

craciunoiuc:
Don't be scared young one, we have already been using it for the past ~3 weeks

Basically this brings the action up to speed with what changed in Kraft in the past year or so since I stopped updating it

if you didn't notice it means it's good to merge ;)

jedevc:
Sure, but does this mean all the code is duplicated for the action? Why?

craciunoiuc:
Yeah, that is correct :pensive: 

Mostly because this is either:
1. Code from `internal` packages to which this does not have access
2. Command code that is ordered. Aka instead of doing `kraft build` + `kraft pkg` + `kraft push` this does it all by calling functions which need to be non-interactive (tailored for CI)

The clean solution would be to move 99% of the code out of `internal` into public packages, and keep commands as simple as possible.

Like that code could be (finally) shared between the two

craciunoiuc:
Cleaned and bumped and ready to merge

craciunoiuc:
I expect not having to do this again for the next months so would be nice if we can get this out of the way of course hehe

jedevc (approved):
Go for it then 😢 

```
```

Please do not release this until Monday, so that there is plenty of time to recover through the week if something goes wrong, I don't want an exciting weekend 🙏

craciunoiuc:
Sure thing, but I mean, as I said, we have been using this for the past 3-4 weeks. I've built and pushed it manually :D

So there's nothing else to do here

Too bad CI decided to go to sleep exactly on the last run (previous was all green)

PR #2907, opened 2026-08-17
fix: Fix config YAML write merge and `system unset` persistence

### LINKED ISSUE #2901: bug(cli/system): `kraft system unset` fails to persist key deletions to disk
### Describe the bug

When running `kraft system unset <KEY>` (e.g. `kraft system unset toolchain.CC` or `kraft system unset toolchain_profiles.gcc10.CC`), the CLI reports that the key is being unset, and `ConfigManager.Unset()` correctly removes the key from the in-memory configuration structure.

However, the change is **never persisted to disk** in `~/.config/kraftkit/config.yaml`. 

This occurs because `internal/cli/kraft/system/unset/unset.go` invokes:
```go
return config.M[config.KraftKit](ctx).Write(true)
```
When `Write(true)` is called on `YamlFeeder`, it performs `yamlmerger.RecursiveMerge(&from, &into)`. This step unmarshals the existing configuration file on disk (`from`) and merges any key present on disk back into the in-memory struct representation (`into`). As a consequence, keys that were deleted from memory during `Unset()` are resurrected by `RecursiveMerge` and re-saved to the disk file.

### Steps to reproduce

1. Set a configuration variable:
   ```bash
   kraft system set toolchain.CC=gcc
   ```
2. Verify that the key is stored in the configuration file:
   ```bash
   cat ~/.config/kraftkit/config.yaml
   ```
3. Attempt to unset the key:
   ```bash
   kraft system unset toolchain.CC
   ```
4. Inspect the configuration file on disk again:
   ```bash
   cat ~/.config/kraftkit/config.yaml
   ```
5. **Observed Result**: `toolchain.CC: gcc` remains in `~/.config/kraftkit/config.yaml`.

### Expected behavior

unning `kraft system unset <KEY>` should remove the specified key from `~/.config/kraftkit/config.yaml` on disk. 

### KraftKit version

staging

x86_64

linux/other

### Relevant log output

```shell
$ kraft system set toolchain.CC=gcc
 i  setting toolchain.CC=gcc

$ kraft system unset toolchain.CC
 i  unsetting key=toolchain.CC

$ cat ~/.config/kraftkit/config.yaml | grep -A 2 toolchain
toolchain:
    CC: gcc
```

ijaidev:
# Fix config YAML write merge and `system unset` persistence

`YamlFeeder.Write` merged the on-disk document into the in-memory node, which is the reverse of `RecursiveMerge`'s contract and of the Kraftfile writer. Disk is now `into`, memory is `from`. Scalar merge copies the node (`*into = *from`) so in-memory values actually overwrite disk. Without that copy, `kraft system set` would stop updating existing keys.

`RecursiveMerge` is additive and cannot delete keys, so `kraft system unset` with `Write(true)` restored removed entries from disk. Unset now uses `Write(false)`, matching `kraft pkg unsource`.

## Commits

- `fix(config,internal/yamlmerger): Correct YAML write merge direction`
- `test(config): Cover YAML feeder write merge direction`
- `fix(cli/system): Persist unset without merging`

Closes #2901

### DIFF
--- a/config/feeder_yaml.go
+++ b/config/feeder_yaml.go
@@ -96,8 +96,8 @@ func (yf YamlFeeder) Write(structure interface{}, merge bool) error {
 		return fmt.Errorf("could not read file: %v", err)
 	}
 
-	from := yaml.Node{}
-	if err := yaml.Unmarshal(data, &from); err != nil {
+	into := yaml.Node{}
+	if err := yaml.Unmarshal(data, &into); err != nil {
 		return fmt.Errorf("could not unmarshal YAML: %s", err)
 	}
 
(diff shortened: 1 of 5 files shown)


craciunoiuc (approved):
All good here. Thanks!

```
```

PR #2924, opened 2026-09-15
fix(bridge): Stops re-creating taps held by running machines

vukedd:
### Prerequisite checklist

### Description of changes

Starting a second machine on a bridge network while another machine on it is still running fails:
```
$ kraft network create -n 172.222.0.1/24 repro0
$ kraft run -d --network repro0 -M 64Mi --name first unikraft.org/nginx:1.25
$ kraft run -d --network repro0 -M 64Mi --name second unikraft.org/nginx:1.25
E  all iterated drivers failed: could not update repro0@if0 link: Tuntap IOCTL TUNSETIFF failed [0], errno device or resource busy
```

Stopping `first` and repeating last command succeeds. 

**Cause:** `v1alpha1Network.Update` [machine/network/bridge/v1alpha1.go](https://github.com/unikraft/kraftkit/blob/staging/machine/network/bridge/v1alpha1.go) walks every interface in `network.Spec.Interfaces` and, for each tap that already exists and has `FlagRunning` set, calls `LinkSetDown` followed by `LinkModify`. For a tap, `FlagRunning` means a process holds the device open, in this case the first machine's QEMU. `LinkModify` re-issues `TUNSETIFF` for that name, which the kernel refuses with `EBUSY` on a tap that already has an attached file. So the branch selects exactly the taps that cannot be modified, and the first running machine blocks every later one from joining the network.

The branch has no consumer: all callers of `Update` (`run`, `rm`, the failure path of `start`, `net rm --force`) only append or remove interfaces, none edits an existing interface's spec, and `tap.HardwareAddr` is nil for existing entries, so the only setting the recreate could ever re-apply was the bridge attachment.

**Fix:** Leave an existing tap alone and only re-attach it to the bridge with `LinkSetMaster` when its master differs. That call is permitted on an in-use tap. `LinkAdd` for new taps, the `alias`/`LinkSetUp` steps and the cleanup pass are unchanged.

**Verification:** With this change, step 3 above creates `second` while `first` is still running.

**Reproduced with stock kraft both inside `kraftkit.sh/base:latest` (built from `staging`; container with `CAP_NET_ADMIN`, /`dev/net/tun`, `/dev/kvm`) and on an Ubuntu host (kernel 7.0) via sudo.**


craciunoiuc (approved):
All good here. Thanks!

```
```

